PR #472, opened 2026-04-02
Migrate select elements to daisyUI 5 label wrapper pattern

mdon:
daisyUI 5 changed the select component to use a <label class="select"> wrapper instead of putting the class directly on <select>. This fixes selects rendering incorrectly on LiveView navigation (content pressed against top) while working correctly after a hard refresh.


ddon (approved):
Approve. Reviewed — see dev_docs/pull_requests/2026/472-migrate-select-daisy5/CLAUDE_REVIEW.md

PR #488, opened 2026-04-13
advanced user dashboard

mithereal:
i created a more advanced user dashboard that will auto load widgets from modules and place them on a grid, its installed via a mix task
i have updated the billing module to reflect the changes

PR #489, opened 2026-04-13
Added feature to make languages sortable in the language admin page

alexdont:


ddon:
Posted a review at [`dev_docs/pull_requests/2026/489-sortable-languages/CLAUDE_REVIEW.md`](https://github.com/BeamLabEU/phoenix_kit/blob/dev/dev_docs/pull_requests/2026/489-sortable-languages/CLAUDE_REVIEW.md).

**TL;DR — approve with minor suggestions.** Small, focused PR that reuses existing `DraggableList` + `SortableGrid` infrastructure and persists order through the existing JSON settings path. Clean diff.

**Suggested polish before merge:**
- Move the inline `<style>` `wiggle` keyframes out of `languages.html.heex` into the CSS pipeline; wrap with `prefers-reduced-motion: reduce` (infinite animation otherwise).
- Add a unit test for `Languages.reorder_languages/1` (pure data transform — covers subset, unknown codes, empty list, dup codes).
- Tiny perf: `MapSet.new(ordered_codes)` for the `remaining` filter to avoid the O(n²) `in` check.

**Non-blocking:**
- Reorder mode hides per-language Set-Default / Disable actions — confirm that's intentional (reads like it is).
- `setTimeout(0)` + `body > .sortable-fallback` workaround is brittle but correctly scoped — a short link to the underlying SortableJS behavior in the comment would help future maintainers.
- Flash strings aren't `gettext`-wrapped, but consistent with the rest of the file — follow-up sweep, not a blocker.

Nice contrast to #488 — this is what a focused feature PR looks like.

PR #490, opened 2026-04-13
Added image_set for correct file pulling on the front end image calls

alexdont:


ddon (commented):
**Pincer review — Approve (with one item to confirm)**

Clean, production-quality PR. The responsive `<.image_set>` component, `VariantNaming` utility, V97 migration, and V95 idempotency fix are all well-implemented. Blocking issue is minor.

**Confirm before merging:**
- `Storage.list_image_set_variants/1` is referenced in `ImageSet.load_variants/1` but doesn't appear in this diff. If it's missing, the component will silently return no variants for every auto-load call. Please confirm it exists (or is being added in a companion commit).

**Non-blocking notes:**
- `Enum.map_join` in `dimensions.html.heex` produces `"JPEG +WEBP +AVIF"` with a leading `+` separator — check the rendered output matches intent. Switching to `Enum.join(..., " + ")` with the `+` outside the `map_join` would read more cleanly.
- `prefix_str/1` in V97 is duplicated from other migrations — pre-existing pattern, just flagging for future cleanup.
- PNG `<source>` entries: all browsers support PNG natively, so a `<source type="image/png">` is redundant. Low priority but worth knowing.

Ready to merge once `list_image_set_variants/1` is confirmed.

ddon:
Hi @alexdont — thanks for the updates! Reviewed again and it looks like no new commits have landed since our last round of feedback in `dev_docs/pull_requests/2026/490-image-set-alternative-formats/CLAUDE_REVIEW.md`. Still waiting on a few items before we can merge:

- **BUG — MEDIUM**: Please confirm `Storage.list_image_set_variants/1` actually exists. The `ImageSet` component's auto-load guard (`Code.ensure_loaded?/1` + `function_exported?/3`) will silently return no variants if the function isn't defined, so if it's missing we'd like it added in this PR (or a companion commit referenced here).
- **IMPROVEMENT — MEDIUM**: The `Enum.map_join` in the dimensions table leaks the separator — renders as `"JPEG +WEBP"` with a stray leading `" +"`. Please adjust the join so the primary format and alternatives render cleanly.
- **IMPROVEMENT — MEDIUM**: Move the `<% current_alternatives = ... %>` computation out of the HEEx template and into an `assign/3` in the LiveView — cleaner and easier to test.
- **NITPICK**: Consider omitting PNG from `<picture><source>` entries (keep as `` fallback only), and adding a guard for the empty-`srcset` case when `fallback_variants` is empty.

The rest of the PR looks great — migration idempotency, component design, and `VariantNaming` utility are all solid. Just need these polished and we're good to merge. Thanks!

ddon:
@alexdont  waiting for your fixed and and now looks like there are conflicts as well to resolve

alexdont:
fixed the issues

ddon (approved):
Follow-up review: all prior findings addressed in 5d0dcc73. Ready to merge.

PR #497, opened 2026-04-16
Updated media browser component

alexdont:


ddon (changes requested):
## Claude review — changes requested

Full writeup in `dev_docs/pull_requests/2026/497-media-browser-improvements/CLAUDE_REVIEW.md`. The trash/soft-delete feature is well-designed overall (V99 migration is idempotent, soft-delete respects scope, trashed files are excluded from normal listings, hydration fix is correct) but there are two tenant-isolation holes that need to close before merge.

### Must fix

**1. Trash view ignores `scope_folder_id` — CRITICAL**
`Storage.list_trashed_files/1` (`storage.ex:1995-2004`) and `Storage.count_trashed_files/0` (`storage.ex:2008-2012`) take no scope parameter, and `load_trashed_files/2` (`media_browser.ex:1209-1213`) doesn't pass one either. A scoped embed that toggles `filter_trash` will display every trashed file system-wide — cross-tenant leak.
**Fix:** add `scope_folder_id` arg to both Storage helpers and apply the same recursive-CTE scope subtree filter that `list_files_in_scope/2` uses; thread `scope` through `load_trashed_files/3`.

**2. Permanent-delete from trash skips scope check — CRITICAL**
`media_browser.ex:900-905`:
```elixir
if socket.assigns.filter_trash do
  Enum.each(socket.assigns.selected_files, fn file_uuid ->
    Storage.delete_file_completely(file_uuid)  # no within_scope? guard
  end)
```
The non-trash branch (soft-delete) correctly `repo.get`s each file and checks `within_scope?/2` before calling `trash_file/1`. The trash branch must mirror that pattern before `delete_file_completely/1`.

**3. `PruneTrashJob` is never scheduled — HIGH**
Worker is defined with moduledoc "Runs daily via cron", but no crontab entry in `config/config.exs` and the installer doesn't inject one into parent apps. As shipped, trash accumulates forever. Same aspirational-docstring pattern as `Activity.PruneWorker`, but since this PR introduces the worker, this is the right moment to wire it up — add `{"0 3 * * *", PhoenixKit.Modules.Storage.Workers.PruneTrashJob}` to the cron plugin and update `lib/phoenix_kit/install/oban_config.ex` so parent apps pick it up.

**4. `@type t` on `File` schema missing `trashed_at` — HIGH**
`schemas/file.ex:94-117` — one-line typespec update.

### Should fix (not blocking)

- Schema docstring Status Flow (`schemas/file.ex:15-19`) doesn't list `trashed`
- `Storage.list_files/1` (`storage.ex:1324-1331`) doesn't filter `status == "trashed"` — low-risk since the admin UI goes through `list_files_in_scope/2`, but any external caller now sees trashed files
- Flash count reflects selected items, not actually-deleted (already pre-existing; worth addressing with #1/#2)

### Highlights (things done well)

- V99 migration idempotency + partial index on `trashed_at WHERE NOT NULL` + clean rollback ✅
- Soft-delete (non-trash branch of `delete_selected`) already has scope guard ✅
- `list_files_in_scope/2` + `count_orphaned_files/1` both add `status != "trashed"` filter ✅
- Trash filter clears on URL navigation (fixes `75b1379a`/`2695ee39`) ✅
- Hydration fix `01acc031` uses `Map.has_key?(assigns, :initial_params)` — the correct guard (doesn't trip on nil) ✅
- `PruneTrashJob` worker routes through `delete_file_completely/1` so variants + S3 + dedup tracking all get cleaned up ✅

Once #1–#4 land, happy to approve.

PR #528, opened 2026-05-09
Adds two opt-in stateless helpers to `PhoenixKitWeb.Components.Core.TableDefault` and migrates two admin pages, fixing a search debounce UX bug along the way.

timujinne:
### New components (lib/phoenix_kit_web/components/core/table_default.ex)
  - `sort_header_cell/1` — clickable `<th>` with chevron-up-mini/chevron-down-mini icon when active. Inert label-only `<th>` when `sort` attr is `nil`. Configurable `event`, `target`, `align` (`:left`/`:right`/`:center`).
  - `search_toolbar/1` — daisyUI `input-sm` with `hero-magnifying-glass` icon and `phx-debounce` (default 300ms). Optional `<form>` wrap when `on_submit` is set. Placeholder defaults to `dgettext("default", "Search...")`.

  ### Migrations
  - `live/users/live_sessions.{ex,html.heex}` — collapsed `:sort_by` + `:sort_order` assigns into a single `:sort = %{by, dir}` map; renamed event `"sort_by"` → `"toggle_sort"` with `"by"` param. Replaced inline header-button + chevron blocks and search form with the new
  components.
  - `live/users/users.html.heex` — replaced bare search form (every keystroke hit the server) with `<.search_toolbar>`. Bug fix.

  ### Tests
  - New `test/phoenix_kit_web/components/core/` directory + `table_default_test.exs` with 13 tests covering both new components.

  ### i18n
  - Catalog resync via `mix gettext.extract --merge` (separate commit). Adds `"Search..."` msgid + `et` / `ru` translations; surfaces accumulated drift from prior commits where extract wasn't run.

  ## Test Plan
  - [x] `mix precommit` green (compile + docs + credo --strict + dialyzer)
  - [x] 13 component tests passing
  - [x] Verified live in Decor 3D Print parent app via Tidewave MCP — both new functions exported and DB schema at v111 (no pending migrations)
  - [ ] `/admin/users/live-sessions` — sort headers + debounced search work in browser
  - [ ] `/admin/users` — debounced search works in browser


timujinne (commented):
## Self-review (Comment)

  Verdict: **APPROVE with NITPICKs.** Three commits: clean .gitignore + gettext
  resync + the feature commit. Feature commit was already through internal
  two-stage review that caught and fixed two bugs (dead `assign_new` placeholder
  default, missing `mix gettext.extract --merge`). Both correctly addressed in
  final state.

  ### IMPROVEMENT - MEDIUM — `search_toolbar/1` form variant double-binds
  `phx-change`

  `lib/phoenix_kit_web/components/core/table_default.ex:491,501` — the form variant
   places `phx-change={@on_change}` on **both** the `<form>` and the inner
  `<input>`. Phoenix LiveView fires `phx-change` once per element binding; with
  identical events on both you can end up with the same `"search"` event firing
  twice in quick succession (input on keystroke, then form on the same change). The
   handler matches both with the same payload — no incorrect state — but it doubles
   server work and is conspicuously redundant.

  **Fix:** drop `phx-change` from `<form>`, keep only on `<input>`. The form is
  there for `phx-submit` on Enter, nothing more.

  ### IMPROVEMENT - MEDIUM — `phx-target` not propagated to `<form>`

  `lib/phoenix_kit_web/components/core/table_default.ex:489-506` — when caller
  passes `<.search_toolbar target="#my-component">`, `phx-target` lands on the
  `<input>` (so `phx-change` retargets correctly) but **not** on the `<form>` (so
  `phx-submit` does not). When the toolbar is rendered inside a `LiveComponent`,

  ### NITPICK — Behaviour change on first click of a new column

  Old `live_sessions.ex` set new-column click to `:desc`. New code sets it to
  `:asc` (spec-correct, matches the chevron-up-first convention of the new
  component). Two columns, low-traffic admin page — fine; mention in CHANGELOG when
   the maintainer writes one.

  ### NITPICK — Chevron icon size/style change

  Old: `hero-chevron-up/down` (24px outline). New: `hero-chevron-up-mini/down-mini`
   (20px solid). Correct call for a header indicator (more compact), but the diff
  is silent about it. Future migrations of other sortable headers should follow the
   same convention.

  ### NITPICK — Test coverage gaps

  Declared but not asserted: `target` on `search_toolbar/1` (would surface the
  second MEDIUM above), `name` custom value, custom `debounce`, `event` default on
  `sort_header_cell/1`. Fold into the standing `core/` test-coverage TODO from
  `CLAUDE.md`.

  ### NITPICK — HEEX duplication between form/div branches

  `table_default.ex:489-521` — `<input>` markup (~12 lines) is copy-pasted into
  both `if @on_submit` branches. EEx makes wrapper-only conditionals awkward, so
  the current shape is readable. Could be tightened later by extracting a private
  `render_search_input/1` partial; not worth it in this PR.

  ### Things deliberately not flagged

  - Gettext catalog drift (~10000 lines in commit `a7c1d35b`): explicitly accepted
  as a separate commit doing long-overdue resync.
  - `mix.exs @version` and `CHANGELOG.md` untouched per `Version + CHANGELOG
  ownership` rule.
  - `<th>` semantics, active-column styling (chevron only), `phx-target={nil}`
  auto-omit, CSRF/XSS, `.gitignore` scope — all correct.

  Full archived review with rationale: `dev_docs/pull_requests/2026/529-table-defau
  lt-sort-search-helpers/CLAUDE_REVIEW.md` (will be in the PR after the fix
  commit).

PR #562, opened 2026-05-22
Fix parse_response/2 marker leak + drive handle_ai_response/2 via real test

mdon:
## Summary

Follow-up to PR #560 — two commits that landed on `mdon/fix-translation-response-shape` after the merge.

### 1. Parser marker-leak bug (real bug, present on `dev` today)

`parse_response/2`'s field-capture boundary regex only terminated at REQUESTED markers. A model that emits a marker the caller didn't ask for silently rolls that block's content into the previous requested field.

**Real-world trigger:** projects-side prompt template referenced `{{title}}` literally (no `title` variable bound for a template resource that only has name + description), so the AI emitted `---TITLE---{{title}}` between `---NAME---` and `---DESCRIPTION---`. The parser, asked for only `name` and `description`, returned name as `"Mitarbeiter-Onboarding---TITLE---{{title}}"` — corrupted.

**Fix:** widen the boundary regex to terminate at any `---[A-Z0-9_]+---` marker (the character class matches what `marker/1` already produces). Unrequested markers' content is dropped silently — they're not in the requested fields list, so the caller never gets them either way.

Verified end-to-end against real deepseek-v3.2: DE-DE and FR-FR translations of the Employee onboarding template were clean after this fix, dirty before.

### 2. `handle_ai_response/2` testability

Codex final-review of #560 caught that the original "OpenAI-shaped response map" test manually extracted `message.content` and only drove `parse_response/2` directly — it would still have passed against the pre-fix broken implementation. Reworked:

- Promoted `handle_ai_response/2` from `defp` to `@doc false def` so unit tests can drive it directly.
- Replaced the cross-module `PhoenixKitAI.Completion.extract_content/1` call with an inline pattern-match on the OpenAI shape. Drops the cross-module dep — works in core's test env (plugin absent) and any host (plugin present, any version).

## Test plan

- [x] New regression test pins the unrequested-marker boundary behaviour
- [x] New `handle_ai_response/2` tests cover the OpenAI map / raw binary / malformed shape branches
- [x] All 21 translation tests pass
- [x] `mix credo --strict` clean

### DIFF
--- a/lib/modules/ai/translation.ex
+++ b/lib/modules/ai/translation.ex
@@ -217,30 +217,30 @@ defmodule PhoenixKit.Modules.AI.Translation do
 
   # `PhoenixKitAI.ask_with_prompt/4` returns the full OpenAI-shaped
   # response map (`%{"choices" => [%{"message" => %{"content" => "..."}}]}`),
-  # not a raw string. Reach for `PhoenixKitAI.Completion.extract_content/1`
-  # to pull the assistant's content out — same helper publishing's
-  # `TranslatePostWorker` already uses for its post-translate AI call.
-  # Older versions of the plugin (or test stubs) may still return a
-  # plain binary; the second clause keeps that working.
-  @compile {:no_warn_undefined, [{PhoenixKitAI.Completion, :extract_content, 1}]}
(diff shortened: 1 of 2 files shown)


mdon:

PR #563, opened 2026-05-22
Pin Translation.translate_fields/6 validation order with a regression test

mdon:
## Summary

Follow-up to PR #558 — codex flagged that the documented validation
order (`endpoint → prompt → non-empty → unique-markers → plugin-
available`) wasn't pinned by any test. A future refactor that
reorders the `with` chain would change caller-facing precedence
without any test catching it.

Added a regression test that stacks multiple input violations and
asserts which one wins:

- empty endpoint vs empty fields → `:no_endpoint` (endpoint wins)
- empty prompt vs empty fields → `:missing_prompt` (prompt wins)
- valid endpoint+prompt + empty fields → `:no_markers`
- valid endpoint+prompt+fields + dup markers → `{:duplicate_markers, _}`

Plus a comment-only tidy on the existing describe block (the order
list there was stale after #558 inserted `validate_non_empty`).

These two commits landed on my fork's branch **after** #558 was
merged, so they didn't make it into upstream/dev. Opening as a
separate small PR.

## Test plan

- [x] 19/0 tests
- [x] No production-code change — test + comment only

### DIFF
--- a/test/phoenix_kit/modules/ai/translation_test.exs
+++ b/test/phoenix_kit/modules/ai/translation_test.exs
@@ -104,6 +104,39 @@ defmodule PhoenixKit.Modules.AI.TranslationTest do
       assert {:error, :ai_not_installed} =
                Translation.translate_fields("ep", "p", "en", "es", %{"a" => "b"})
     end
+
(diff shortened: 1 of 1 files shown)


mdon:

PR #564, opened 2026-05-22
Trim implementation-detail comments in magic_link.html.heex

mdon:
## Summary

Codex final-review CONCERN from PR #559: the long inline comments
explaining the `data-phx-skip` / 183px / form-ownership root causes
in `magic_link.html.heex` belong in the PR description, not in
templates where they'll age out faster than the fix.

Kept short comments naming the root rules ("fieldset default
min-width", "double-swap broke the DOM merge") so a future edit
sees the intent without the wall of text.

## Notes

- HTML-template-only change. No logic touched.
- Surfaced during PR #559 review; landed late and got orphaned
  on the merged branch. This PR re-applies it cleanly off
  `upstream/dev`.
- The fix in PR #559 (`min-w-0` on `<fieldset>` + removed
  `phx-disable-with` to avoid double-swap DOM conflict) remains
  in place; only the explanatory comments are trimmed.

## Test plan

- [x] `mix compile` clean
- [x] `mix format --check-formatted` clean on the touched file
- [ ] Visual check at `/phoenix_kit/users/log-in/magic-link` —
      submit button still fills the card, no overflow

### DIFF
--- a/lib/phoenix_kit_web/users/magic_link.html.heex
+++ b/lib/phoenix_kit_web/users/magic_link.html.heex
@@ -14,14 +14,8 @@
     phx-submit="send_magic_link"
     phx-change="validate"
   >
-    <%!--
-      `min-w-0` overrides the browser-default `min-width: min-content`
-      on <fieldset>, which would otherwise let the fieldset grow past
-      its parent <form>'s width whenever a child has a longer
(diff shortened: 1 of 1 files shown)


mdon:

PR #575, opened 2026-05-31
feat(auth): multi-session account switching (opt-in)

timujinne:
## Multi-session user switching

Adds opt-in multi-account session support to PhoenixKit: a signed-in user can add a second account, switch between active sessions, and log out of one or all — gated behind a new `multi_session_enabled` setting (admin toggle).

### What's included
- **Setting + admin toggle** — `multi_session_enabled`, surfaced in Settings.
- **`MultiSession` module** — read helpers (session stack, gate, `list_accounts`) and write ops (add / switch / remove / logout) with audit logging.
- **Routes + controller actions** — basic and localized scopes; split active-session vs all-accounts logout.
- **Plug + `on_mount`** — assigns session accounts and enforces the multi-session gate.
- **Scope** — carries multi-session account data through to the layout.
- **UI** — account switcher + "add account" modal in the header dropdown, plus a "Log out all accounts" control.

### Tests
Unit + integration coverage for the `MultiSession` ops and the session stack.


timujinne:
## Code review

**Verdict:** 🔴 Changes requested

Well-structured feature with a clean `MultiSession` module boundary and good coverage, but several **security** and correctness issues should be addressed before merge.

### Security

🔴 **[blocking] No session-fixation protection on `add_account` / `switch_to`** — `lib/phoenix_kit_web/users/multi_session.ex` (~104–147). Both mutate the session via `put_session` but never `configure_session(renew: true)` / `delete_csrf_token()`. The normal `log_in_user` path renews the session (auth.ex:104) precisely to avoid fixation. Adding/switching an authenticated principal onto an existing session id without cycling it carries a pre-auth session id into elevated state. Renew at the single chokepoint (`put_active_token/2`).

🔴 **[blocking] Gate defaults to enabled before any setting is stored** — `multi_session.ex:45` (`get_boolean_setting("multi_session_enabled", true)`) and `settings.ex:168` (default `"true"`). On a fresh install the feature is silently **on** for Owner/Admin without an explicit toggle. Default to `false`/`"false"` — this is meant to be opt-in.

🔴 **[blocking] `GET /users/log-out` ignores the session stack** — `session.ex:140–144`. The `DELETE` path runs `log_out_active`/`delete_all_stack_tokens`, but `GET` calls `UserAuth.log_out_user/1` directly, deleting only the active token and leaving the other stack tokens **valid in the DB**. Mirror the `DELETE` behaviour (clear the stack first).

🟡 **[minor] `switch_to` doesn't re-check the target user is still active** — `multi_session.ex:132–147`. `find_token_by_ref` only confirms stack membership; a user deactivated after being added can still become active. Guard with `Auth.ensure_active_user/1` (as the mount path does).

🟡 **[minor] `add_account` allows duplicate accounts** — `multi_session.ex:104–130`. No UUID membership check, so the same user can occupy multiple stack slots with fresh tokens. Dedup against existing stack accounts.

### Correctness

🟡 **[minor] N+1 on every authenticated request** — `auth.ex:334–337` + `multi_session.ex:63–87`. `list_accounts` issues ~2 queries/account (up to 10 for a 5-account stack) and `gate_allowed?` re-fetches the root token independently. Derive the gate from the already-resolved root account / share one prefetch.

🟡 **[minor] `find_token_by_ref` queries per token** — `multi_session.ex:217–221`. Up to 5 queries on switch/remove; reuse the `ref → token` mapping `list_accounts` already built.

🟢 **[nit] `log_out_active` doesn't broadcast disconnect for the removed account** — `multi_session.ex:181–201`. Dangling `live_socket_id`; broadcast disconnect before `put_active_token`, like the full-logout path.

### Tests

🟡 **[minor] Gate tested only for `add_account`** — `session_multi_test.exs`. Add forbidden-path tests for `PUT /users/session/active` and `DELETE /users/session/accounts/:ref` (plain user / setting off).

🟡 **[minor] `return_to` open-redirect not tested** — validation exists at `session.ex:119–121` but is untested; add a `return_to=//evil.com` → `/` case.

🟢 **[nit] Flash strings not `gettext`-wrapped** — `session.ex:93,171` interpolate `user.email` directly; wrap with `gettext/2` + named binding for i18n consistency.

timujinne:
Addressed the review in `ded76f4d` (+ UI polish in `a1d80093`):

**🔴 Blocking**
- **Session fixation** — `add_account` / `switch_to` now go through `renew_and_put_active_token/2`, which drops the CSRF token and `configure_session(renew: true)` (rotates the session id) before writing the active token. Stack data in the session is preserved (renew only rotates the id).
- **Gate default** — `multi_session_enabled` now defaults to `false` in both `gate_allowed?` and the settings default; the feature is explicit opt-in.
- **`GET /users/log-out`** — now drains the stack (`delete_all_stack_tokens`) before delegating, mirroring `DELETE`; no secondary-account tokens left valid.

**🟡 Minor**
- `switch_to` guards the target with `Auth.ensure_active_user/1` (no switching to a deactivated user).
- `add_account` dedups by user uuid (`{:error, :already_in_stack}` + flash).
- `log_out_active` broadcasts LiveSocket disconnect for the removed account.
- Added gate-forbidden tests for `PUT /session/active` and `DELETE /session/accounts/:ref`, plus `return_to` open-redirect guard tests.

**🟢 Nits**
- Account-switch flash strings wrapped in `gettext/2`.

**Perf (N+1)** deferred — addressing it cleanly means threading the resolved `ref → token` map through `list_accounts`; left for a focused follow-up to avoid touching the read-path API here.

**UI polish** (`a1d80093`): "Log Out" → "Log out", "Log out all accounts" → "Log out from all accounts", and long account emails now wrap to a second line instead of overflowing the dropdown.

`mix format` clean; `mix compile --warnings-as-errors` clean; `mix credo --strict` clean on changed files. (Full ExUnit suite runs in upstream CI.)

timujinne:
## Code review — multi-session enhancements (role label, gate-for-all, OAuth add-account)

**Verdict:** 🟡 Approve with minor fixes — core logic sound and security fundamentals correct, but one session-key leak must be fixed before merge.

Verified: role-label rewrite is correct (`Scope.user_roles/1` + `Role.system_roles/0` exist, priority chain handles empty/plain-User/custom/Owner/Admin); gate relaxation correct (anonymous → false); session-fixation protection on the add path via `renew_and_put_active_token/2`; open-redirect guard blocks `//evil.com`; `OAuthAvailability.provider_enabled?/1` honors both the master `oauth_enabled` toggle and per-provider flag (matches main login).

### Findings

🔴 **`oauth_add_account_intent` leaks on OAuth failure** — `oauth.ex` (`ueberauth_failure` ~275 and the catch-all ~285). The intent key is read+cleared only in the success branch. On a failed/denied OAuth attempt it persists in the session cookie; a subsequent *normal* Google sign-in then sees the marker + passing gate and silently appends the account to the stack instead of logging in. Fix: clear the key on both failure branches — or, cleaner, consume all three session params (referral_code, return_to, intent) at the top of `callback/2` so every branch starts clean.

🟡 **Gate-disabled mid-flight falls back to a full login** — `oauth.ex` (~240-248). When the intent is set but `gate_allowed?` is false (setting toggled off, or root token expired), it silently `log_in_user`s as the OAuth user, replacing the primary session. Prefer an explicit error flash + redirect back over the opaque full-login.

🟢 **`add_authenticated_user/2`** — add a doc note that callers must pass a freshly DB-loaded `%User{}` (the `is_active: true` head relies on live data). OAuth path is fine in practice.

🟢 **`activity_log` hardcodes `"actor_role" => "admin"`** (`multi_session.ex` ~327) — now inaccurate for plain-user-initiated adds after the gate relaxation. Pre-existing; consider removing/filling dynamically.

🟢 **`add_account_oauth_buttons` reads provider settings 4× per render** (`admin_nav.ex` ~712) — fine for an infrequently-rendered modal; cache if the nav renders on every request.

**Tests:** once the failure branches are fixed, add a test asserting the intent key is absent from the session after a failed callback.

timujinne:
Addressed the review in `3e5ec2a8`:

- **🔴 intent leak** — `:oauth_add_account_intent` is now cleared in both OAuth callback failure branches (`ueberauth_failure` + the unexpected-callback catch-all), so a denied/failed add-account attempt can no longer leave the marker in the session for a later normal sign-in to inherit.
- **🟡 gate-off fallback** — when the add-account intent is set but the gate is closed, the callback now shows an explicit "Account switching is currently disabled." error and returns home, instead of silently replacing the primary session with a full login.

🟢 nits (hardcoded `actor_role`, doc note, render-time provider reads) left as-is — pre-existing / cosmetic. Follow-up: an OAuth-failure session test is best added where the Ueberauth controller test harness + DB are available (couldn't be exercised in this sandbox).

`mix format` clean; `mix compile --warnings-as-errors` clean.

PR #576, opened 2026-06-01
Router-served fallback for catalogue PDF.js viewer assets

mdon:
## Summary

Adds a router-served fallback for `phoenix_kit_catalogue`'s vendored PDF.js viewer, so the embedded viewer works on hosts whose endpoint is missing the `/_pdfjs` `Plug.Static` mount — e.g. deps bumped without re-running `mix phoenix_kit.update`, or an endpoint regenerated by a scaffold script. Fixes a real `NoRouteError` on a deployed host.

> Pairs with `BeamLabEU/phoenix_kit_catalogue#31`. Independent, but that PR's viewer robustness relies on this route being present.

### Commits
- **Router-served `/_pdfjs` fallback** (`db1906f8`) — `PhoenixKitWeb.PdfViewerController` + a `get "/_pdfjs/*path"` route wired by `phoenix_kit_routes()` (compiled only when the catalogue module is loaded). Streams the vendored bytes from the catalogue app's `priv/static/pdfjs/`. The endpoint `Plug.Static` mount still wins when present (endpoint plugs precede the router); this only catches the fall-through. Crucially it rides the routing macro the host already calls, so it appears on a plain lib upgrade — no mix task re-run needed.
- **Address review (Codex)** (`888364bb`) — serve without the JSON-only `:phoenix_kit_api` pipeline (these are static HTML/JS-module/CSS/font assets, not JSON; the controller sets its own content-type); replace the path-traversal containment check's hardcoded `/`-prefix match with a separator-aware segment comparison (`Path.split`/`Path.expand` + `List.starts_with?/2`) so it's correct on any OS and rejects sibling-prefix paths (`.../pdfjs-evil`).

## Verification
- `mix precommit` clean (compile + format + credo --strict + dialyzer).
- Route registered (`mix phx.routes` → `GET /_pdfjs/*path`). Assets serve `200` with correct content-types (`viewer.html` → text/html, `pdf.worker.mjs` → text/javascript); missing path → `404`; traversal containment verified (valid allowed, `../../etc/passwd` + sibling-prefix rejected).

## Test plan
- [x] precommit clean (compile / format / credo --strict / dialyzer)
- [x] `/_pdfjs/web/viewer.html` → 200 text/html; `build/pdf.worker.mjs` → 200 text/javascript
- [x] missing path → 404
- [x] traversal/sibling-prefix containment verified


mdon:
Closing — the team is moving off the `dev` branch. Re-pointed to `main` as #577 (same two commits, cherry-picked cleanly onto main).

PR #588, opened 2026-06-09
feat(annotations): add has_annotations?/1 detect helper

timujinne:
## What

Adds `PhoenixKit.Annotations.has_annotations?/1` — a small boolean predicate that returns `true` when a file has at least one annotation.

## Why

Host apps need a cheap way to decide whether a file needs annotation-aware work before doing something expensive. In our case (ANDI), before composing an order document we flatten any attached photo that carries drawn annotations into a baked raster; for photos without annotations we skip the whole browser round-trip. `list_for_file/1` works but loads rows just to test emptiness at every call site, so a named predicate reads better and centralizes the check.

## Notes

- Implemented on top of the existing `list_for_file/1` (annotation counts per file are small); can later be tightened to an `exists?` query without changing the contract.
- Nil/non-binary input returns `false`.


timujinne (commented):
**Automated code review**

**VERDICT: REQUEST_CHANGES**

**SUMMARY:** The function is correctly placed, typed, and documented, but the implementation fetches every annotation row from the database just to test emptiness — an unnecessary full load when a single `EXISTS` query suffices. There is also a silent catch-all clause that swallows non-binary arguments without error, which conflicts with the `@spec`.

**FINDINGS:**

- **[MAJOR]** `lib/phoenix_kit/annotations/annotations.ex` — `has_annotations?/1` calls `list_for_file/1`, which loads all annotation rows before comparing to `[]`. For files with many annotations this is a full row-fetch where only a boolean is needed. Replace with an `exists?` query: `RepoHelper.repo().exists?(from a in Annotation, where: a.file_uuid == ^file_uuid)` (or expose `RepoHelper.exists?/1`).
- **[MINOR]** `lib/phoenix_kit/annotations/annotations.ex` — `def has_annotations?(_), do: false` silently returns `false` for atoms, integers, structs, etc., while the `@spec` advertises `uuid()` (`String.t()`). Either remove the catch-all (raise on bad input, consistent with `list_for_file/1`) or widen the `@spec` to `term()` if the total contract is intentional.
- **[NITPICK]** `lib/phoenix_kit/annotations/annotations.ex` — `@doc` phrasing `"True if…"` doesn't match the verb-phrase style of sibling docs (`"Fetch a single…"`, `"List annotations…"`). Suggest `"Returns true if the file has at least one annotation."`.

timujinne:
Review addressed in `4475ad85`: `has_annotations?/1` now uses `RepoHelper.exists?/1` (EXISTS query) instead of loading rows; `@spec` widened to `term()` to match the defensive catch-all; doc reworded to "Returns true if…".

timujinne:
Superseded — re-opening as a main → main PR per our fork workflow. Same commits (now incl. the review fixes) are on `timujinne:main`.

PR #589, opened 2026-06-09
feat(annotations): add has_annotations?/1 detect helper

timujinne:
## What
Adds `PhoenixKit.Annotations.has_annotations?/1` — returns `true` when a file has at least one annotation, via an `EXISTS` query (`RepoHelper.exists?/1`).

## Why
Host apps need a cheap way to decide whether a file needs annotation-aware work before doing something expensive. In ANDI, before composing an order document we flatten any attached photo that carries drawn annotations into a baked raster; for photos without annotations we skip the whole browser round-trip. A named predicate reads better than `list_for_file/1 != []` at every call site and avoids loading rows just to test emptiness.

## Notes
- Uses `RepoHelper.exists?/1` → `SELECT 1 … LIMIT 1` (no row load).
- `@spec` is `term() :: boolean()`; a defensive catch-all returns `false` for non-binary input so host filters can't crash on malformed data.

Net diff is `lib/phoenix_kit/annotations/annotations.ex` only (+8). Reviewed; review fixes (exists?, spec, doc phrasing) already folded in.


timujinne (commented):
**Automated code review (re-run on the main→main PR)**

**VERDICT: APPROVE**

**SUMMARY:** Tiny, correct additive helper. `RepoHelper.exists?/1` delegates to `Ecto.Repo.exists?/2` → `SELECT 1 … LIMIT 1` (no row load). The defensive `has_annotations?(_) -> false` clause handles nil/non-binary input without a DB hit. No functional, architectural, or security concerns.

**FINDINGS:**
- [nitpick] `lib/phoenix_kit/annotations/annotations.ex` — `@spec ... term()` is accurate, but sibling specs use the narrower `uuid()`. `@spec has_annotations?(uuid() | term()) :: boolean()` would signal that a UUID string is the happy path. Non-blocking, stylistic.
- [nitpick] `lib/phoenix_kit/annotations/annotations.ex` — the one-line `@doc "Returns true if…"` is fine; siblings use a bare verb phrase (`"Fetch a single…"`). Pure style.

PR #603, opened 2026-06-24
Add sitemap_sources/0 module callback for source auto-registration

timujinne:
## Summary

External PhoenixKit modules could not contribute entries to the generated sitemap: `Generator.get_sources/0` only read a hardcoded `default_sources/0` list (router discovery, static, publishing, posts, shop) or a host `config :phoenix_kit, :sitemap, sources:` override. There was no auto-discovery, so a module shipping its own `Sitemap.Sources.Source` (e.g. Entities) was never invoked unless every host hand-wired it into config.

This adds a zero-config registration path mirroring the existing route / CSS / JS auto-discovery.

## Changes

- `PhoenixKit.Module`: new optional `sitemap_sources/0` callback (default `[]`), added to `@optional_callbacks`, the `__using__` defaults, and `defoverridable`.
- `PhoenixKit.ModuleRegistry.all_sitemap_sources/0`: aggregates the callback across all registered modules (deduped), using the existing `safe_call/3`.
- `PhoenixKit.Modules.Sitemap.Generator.get_sources/0`: appends module-contributed sources to the base list (host config or defaults) and dedups, order-preserving. Backward compatible — a host that sets `:sources` keeps it as the base; module sources are appended.

## Test Plan

- [x] `mix compile` clean
- [x] `mix credo --strict` — no issues
- [x] `mix dialyzer` — passed
- [x] Verified on a live parent app: `Generator.get_sources/0` now includes a module-contributed source and its URLs appear in the generated sitemap.


timujinne (commented):
## Summary

This PR adds an optional `sitemap_sources/0` callback to `PhoenixKit.Module`, a `ModuleRegistry.all_sitemap_sources/0` aggregator, and wires module-contributed sources into `Sitemap.Generator.get_sources/0` (append + dedup). The behaviour wiring is clean and complete — the callback is added consistently to `@callback`, `@optional_callbacks`, the `__using__` default (`def sitemap_sources, do: []`), and `defoverridable`, matching the existing `css_sources`/`js_sources` pattern exactly. Failure isolation is handled at two layers (`safe_call` per-module in the registry, `Source.safe_collect` per-source at collect time). Overall this is a well-targeted, low-risk change. The points below are mostly about documenting/scoping the backward-compat shift and one enable-state edge case.

## Findings

### BUG - MEDIUM
- `lib/phoenix_kit/module_registry.ex:188-193` (`all_sitemap_sources/0`) iterates `all_modules()`, i.e. **all registered modules regardless of enabled state**, unlike the natural expectation that a disabled module contributes nothing. The mitigation is indirect: `Source.safe_collect/2` (`generator.ex:444`, `source.ex:110`) gates on the *source module's* `enabled?/0`, which is a separate concept from the *PhoenixKit module's* `enabled?/0`. So a module that an operator has disabled in the registry can still have its sitemap source emit entries if that source's own `enabled?/0` returns `true`. If the intent is "disabled module → no sitemap contribution," switch to `enabled_modules()` here (consistent with how a feature toggle should behave), or document explicitly that source inclusion is governed solely by the Source's `enabled?/0`, not the module toggle.

### IMPROVEMENT - HIGH
- `lib/modules/sitemap/generator.ex:425-435` — the backward-compat semantics shift is real and silent: previously `config :phoenix_kit, sitemap: [sources: [...]]` gave the host an *authoritative, exhaustive* source list. Now module sources are **unconditionally appended with no opt-out**. A host that intentionally narrowed `:sources` (e.g. omitted `Shop` or pruned to a minimal set) can no longer prevent a module from re-introducing content via `sitemap_sources/0`. Consider one of: (a) a config escape hatch like `sitemap: [include_module_sources: false]`, and/or (b) a one-line note in the `get_sources/0` `@doc false` comment and the `module.ex` callback docs stating that an explicit `:sources` config is now a *base* list that module sources extend rather than replace. Even without the escape hatch, the doc note prevents a surprising regression for existing hosts.

### IMPROVEMENT - MEDIUM
- `lib/modules/sitemap/generator.ex:437-441` — `module_sitemap_sources/0` wraps `all_sitemap_sources/0` in `rescue _ -> []`, but `all_sitemap_sources/0` already routes every per-module call through `safe_call/3`, which rescues and logs individually. The only thing the outer rescue catches is a registry-level failure (e.g. `:persistent_term` not yet populated), and it swallows it silently with a bare `rescue _`. If you keep it, narrow it and log (`rescue e -> Logger.warning(...) ; []`) so a genuinely broken registry isn't invisible; otherwise it can be dropped since the inner layer already isolates the realistic failure mode.

### NITPICK
- `lib/phoenix_kit/module_registry.ex:192` and `lib/modules/sitemap/generator.ex:434` both call `Enum.uniq()` on the source list, and the final entries are deduped again downstream by `uniq_by(& &1.loc)` (`generator.ex:449`). The `Enum.uniq()` inside `all_sitemap_sources/0` is redundant with the one in `get_sources/0` after the `base_sources ++` concat — a single dedup at the `get_sources/0` level would suffice. Harmless, just slightly belt-and-suspenders.
- `lib/phoenix_kit/module.ex` — no test accompanies the new callback. A small unit test asserting (a) the `__using__` default returns `[]`, (b) `all_sitemap_sources/0` flattens + dedups across two fixture modules, and (c) a raising `sitemap_sources/0` is isolated to `[]` would lock in the failure-isolation contract that is the main value of this PR.

## Verdict

Looks good to merge after addressing the IMPROVEMENT - HIGH backward-compat documentation (and deciding on the enabled-vs-all-modules question in BUG - MEDIUM). No correctness blockers; the wiring itself is correct and idiomatic.

PR #606, opened 2026-06-25
Extract referrals into the phoenix_kit_referrals module

alexdont:
Moves the referral-codes feature out of core into the standalone, auto-discovered **`phoenix_kit_referrals`** package — same model as `posts` and `user_connections`: core keeps the database tables (the versioned migrations are **untouched**), the package ships the schemas, business logic, and admin UI.

Companion PR: **BeamLabEU/phoenix_kit_referrals** (initial release).

## What changes in core

- Remove `lib/modules/referrals/` and the internal route module; drop `PhoenixKit.Modules.Referrals` from `internal_modules/0`; remove the hardcoded "Referral Codes" admin subtab and the hardcoded referrals route injection in `integration.ex`. The module's tabs and routes now flow through the generic module-discovery path (`admin_tabs/0`, `settings_tabs/0`, `route_module/0`).
- Add **`PhoenixKit.Users.Referrals`** — a runtime-dispatch facade that resolves the installed referrals module by key (`ModuleRegistry.get_by_key("referrals")`) and dispatches via `apply`, so **core has no compile-time dependency** on the package. The registration / OAuth / magic-link flows just alias the facade; with the module absent the system reads as disabled, lookups return `nil`, and `use_code/2` is a no-op (the referral field simply doesn't appear).
- Update the module-list tests (referrals is now external; also add the previously-missing `Notifications` entry so the feature-key count assertion stays exact) and the permissions docstring.

## Migrations / data

No migration changes. `phoenix_kit_referral_codes` and `phoenix_kit_referral_code_usage` stay owned by core's `v04`–`v74`, exactly as today — existing installs are unaffected; the module reads/writes those tables.

## Verification

- `mix compile --warnings-as-errors`, `mix credo --strict`, `mix dialyzer` all clean; module behaviour tests pass (67, 0 failures).
- End-to-end in a host app (`kit_test`) with the module installed: referral routes inject at the correct localized + non-localized paths via discovery (no hardcoded wiring), `get_by_key("referrals")` resolves the module, and the feature key is present for nav/permissions.


alexdont:
Superseded by a PR from `main` (per the fork:main → upstream:main flow).

PR #614, opened 2026-07-02
Fix sitemap silent-disable, honor SEO noindex, add reserved-route-prefix callback

timujinne:
## Summary

Three small, independent, additive core fixes from a production SEO/sitemap audit:

1. **RouterDiscovery no longer silently disables on a bad pattern.** Exclude / include-only patterns were compiled inline with `Regex.compile/1` and compile errors were swallowed by `_ -> false` with no log. A mistaken glob such as `"*"` (an invalid regex) would drop the source to zero URLs in whitelist mode with no diagnostic trail. Patterns are now compiled once, invalid ones are logged and skipped, and the regex-not-glob semantics are documented.

2. **The sitemap now honors the SEO module's global `noindex` directive.** When `seo_no_index` is active, `Generator.generate_all/1` (and `generate_html/1`) publish an empty (but valid) sitemap instead of advertising crawlable URLs. The noindex toggle also now invalidates + regenerates the cached sitemap so it doesn't keep serving a stale file.

3. **New optional `PhoenixKit.Module` callback `reserved_route_prefixes/0`** (mirrors `sitemap_sources/0`), aggregated via `ModuleRegistry.all_reserved_route_prefixes/0`. Lets a module declare top-level route segments it owns for its own LiveViews/controllers, so another module's database-driven dispatch (e.g. Publishing's group catch-all) doesn't swallow a route it doesn't actually own. Companion PRs in `phoenix_kit_legal` (declares `["legal"]`) and `phoenix_kit_publishing` (consults the registry in `RouterDispatch.known_group?/1`) depend on this.

These three ended up in one PR because of a `main`-only workflow — each is independently reviewable (they touch disjoint functions) and none depends on another.

## Changes

- `lib/modules/sitemap/sources/router_discovery.ex` — compile patterns once, log + skip invalid ones, document regex (not glob) syntax.
- `lib/modules/sitemap/generator.ex` — short-circuit to an empty sitemap (XML + HTML) when `SEO.no_index_enabled?()`.
- `lib/modules/seo/seo.ex` — `update_no_index/1` invalidates + regenerates the sitemap.
- `lib/phoenix_kit/module.ex` + `lib/phoenix_kit/module_registry.ex` — `reserved_route_prefixes/0` callback + `all_reserved_route_prefixes/0` aggregator.
- Tests for all of the above.

## Test plan

- [x] `mix credo --strict` (0 issues)
- [x] `mix dialyzer` (passed, via pre-commit hook)
- [x] `mix test test/phoenix_kit/module_registry_test.exs test/phoenix_kit/module_test.exs` (70 tests, 0 failures)
- [ ] Integration tests (sitemap noindex) need PostgreSQL — execute in CI

### DIFF
--- a/lib/modules/seo/seo.ex
+++ b/lib/modules/seo/seo.ex
@@ -9,6 +9,7 @@ defmodule PhoenixKit.Modules.SEO do
   use PhoenixKit.Module
 
   alias PhoenixKit.Dashboard.Tab
+  alias PhoenixKit.Modules.Sitemap.Generator
   alias PhoenixKit.Settings
 
   @module_enabled_key "seo_module_enabled"
@@ -69,7 +70,23 @@ defmodule PhoenixKit.Modules.SEO do
   Updates the directive to the provided boolean value.
   """
   def update_no_index(enabled?) when is_boolean(enabled?) do
-    Settings.update_boolean_setting_with_module(@no_index_key, enabled?, @module_name)
+    result = Settings.update_boolean_setting_with_module(@no_index_key, enabled?, @module_name)
+
+    # The directive flips what the sitemap must advertise. Drop the cached
+    # sitemap and regenerate so `/sitemap.xml` reflects the new state instead of
+    # serving a stale file. Best-effort — never let it break the toggle.
+    refresh_sitemap()
+
+    result
+  end
+
+  # Invalidate + regenerate the sitemap after a noindex change. Wrapped so a
+  # sitemap/Oban hiccup can never fail the settings write.
+  defp refresh_sitemap do
+    Generator.invalidate_and_regenerate()
+    :ok
+  rescue
+    _ -> :ok
   end
 
   @impl PhoenixKit.Module
--- a/lib/modules/sitemap/generator.ex
+++ b/lib/modules/sitemap/generator.ex
@@ -102,11 +102,17 @@ defmodule PhoenixKit.Modules.Sitemap.Generator do
 
   # When the SEO module's global `noindex` directive is active the site is
   # asking search engines not to index it, so we publish an empty (but valid)
-  # `<urlset>` rather than advertising crawlable URLs. This pairs with the
-  # `<meta name="robots" content="noindex">` directive host layouts emit for
-  # the same setting. Any leftover per-module files are removed.
+  # `<urlset>` rather than advertising crawlable URLs, for both flat and index
+  # modes. Note this makes `/sitemap.xml` serve a `<urlset>` (not a
+  # `<sitemapindex>`) while noindex is on — an empty urlset is schema-valid and
+  # harmless. Any leftover per-module files are removed.
+  #
+  # Keyed on `no_index_enabled?/0` alone (not `module_enabled?/0`) to stay in
+  # lockstep with the `<meta name="robots" content="noindex">` directive host
+  # layouts emit for the same setting — an empty sitemap always pairs with a
+  # noindex meta.
   defp do_generate_no_index(xsl_style, xsl_enabled) do
-    Logger.info("Sitemap: seo_no_index active — publishing empty sitemap")
+    Logger.debug("Sitemap: seo_no_index active — publishing empty sitemap")
 
     xml = build_urlset_xml([], xsl_style, xsl_enabled)
     FileStorage.save_index(xml)
@@ -362,6 +368,11 @@ defmodule PhoenixKit.Modules.Sitemap.Generator do
 
       style not in ["hierarchical", "grouped", "flat"] ->
         {:error, :invalid_style}
+
+      seo_no_index?() ->
+        # noindex active: never advertise URLs and never serve a stale cached
+        # HTML sitemap — render an empty one, bypassing the cache.
+        HtmlGenerator.generate(opts, [], :"html_#{style}", cache: false)
 
       true ->
         cache_key = :"html_#{style}"
--- a/lib/modules/sitemap/generator.ex
+++ b/lib/modules/sitemap/generator.ex
@@ -31,6 +31,7 @@ defmodule PhoenixKit.Modules.Sitemap.Generator do
   require Logger
 
   alias PhoenixKit.Modules.Languages
+  alias PhoenixKit.Modules.SEO
   alias PhoenixKit.Modules.Sitemap
   alias PhoenixKit.Modules.Sitemap.Cache
   alias PhoenixKit.Modules.Sitemap.FileStorage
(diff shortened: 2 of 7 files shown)


timujinne (commented):
## Code Review — commit `327a541c`

**Verdict:** Spec claims hold. The regex/glob semantics are preserved 1:1, the noindex short-circuit returns the shape callers expect, and the new test matches the real API. No blocking bug. The correctness concerns are about *coherence* of the noindex feature (staleness + the HTML sitemap path), not about the changed code being wrong. Recommend addressing findings 1 and 2 before relying on noindex as staging protection.

### Stage 1 — Spec compliance (all confirmed)

- **Regex semantics preserved 1:1.** Walked every branch against the old code:
  - Old `included?(_, [])` → `true`  ≡ new `compile_include_only([]) → :all` → `included?(_, :all) → true`.
  - Old whitelist with all-invalid patterns → `Enum.any?` over compile-failures → `false` → everything excluded (the silent-disable) ≡ new `{:whitelist, []}` → `Enum.any?([]) → false` → everything excluded, now logged.
  - Old/new `excluded?` and non-empty whitelist match: identical set of matching regexes, invalid ones contribute nothing either way. Only additions are the log line and compile-once. Match outcomes are unchanged.
- **`do_generate_no_index/2` return shape** is `{:ok, %{index_xml, modules: [], total_urls: 0}}`, matching the index-mode shape. The controller paths handle it: `generate_and_serve_index` (controller.ex:162) and `generate_module_on_demand` (controller.ex:206) both destructure `modules`/`total_urls` and work with `[]`/`0`; `generate_xml/1` works with `modules: []`; `Sitemap.update_module_stats/1` accepts a list.
- **Test API is real.** `Settings.update_boolean_setting/2` and `update_setting/2` exist. The empty-`<urlset>` assertions hold: with no entries, `build_urlset_xml/3` emits declaration + xsl line + `@urlset_open` + `@urlset_close`; `"<urlset"` and `"</urlset>"` are present, `"<url>"`/`"<loc>"` are not. `async: false` is correct — it mutates a global setting.

### Stage 2 — Findings

**1. IMPROVEMENT-MEDIUM — HTML sitemap is not gated by noindex (inconsistency).** `generate_html/1` and `serve_html/3` (controller.ex:248) still collect and render the full URL list, so `GET /sitemap.xml?format=html` advertises every crawlable URL while noindex is active, and a previously cached HTML sitemap is served regardless. `do_generate_no_index/2` also never clears that HTML cache. Fix: apply the same `seo_no_index?()` short-circuit to `generate_html/1` (return an empty HTML sitemap), or at minimum invalidate the `:"html_*"` cache keys.

**2. IMPROVEMENT-MEDIUM — Toggling `seo_no_index` does not regenerate the on-disk sitemap (staleness).** The empty `<urlset>` is only produced on the next `generate_all/1`. But `SEO.update_no_index/1` and the settings toggle only write the setting — neither invalidates/regenerates the sitemap, and the controller serves the existing disk file whenever `FileStorage.index_exists?/0` is true. Result: the robots meta flips immediately (computed per request), but `/sitemap.xml` keeps advertising URLs until the scheduler cron or an unrelated sitemap settings change invalidates. Disabling noindex is worse — the sitemap stays empty until regeneration. Fix: have the SEO toggle invalidate + regenerate the sitemap (guarded, since Sitemap is an optional module).

**3. NITPICK / considered-and-fine — `seo_no_index?` gates on `no_index_enabled?` only, not `module_enabled?`.** Not a bug: the robots meta is also keyed purely on `seo_no_index`, and `SEO.disable_module/0` force-clears the directive, so an empty sitemap always pairs with a noindex meta. Worth a one-line comment noting the deliberate parity so a future reader does not add a `module_enabled?` gate that would desync from the meta.

**4. NITPICK — logging volume.** `do_generate_no_index/2` logs at `:info` on every regeneration while noindex is active, so the scheduler cron emits a recurring info line indefinitely — consider `:debug`. `compile_patterns/2` logs once per `do_collect/1`, i.e. once per language plus once for HTML collection — not per-route, so acceptable.

**5. NITPICK — index-mode shape switch.** In index mode `/sitemap.xml` normally serves a `<sitemapindex>`; under noindex it now serves an empty `<urlset>` with the urlset XSL reference. Valid and harmless, but an undocumented shape change for that endpoint — a line in the `do_generate_no_index/2` comment or module doc would help.

### Things that are genuinely fine
- Compilation clean, `mix credo --strict` clean on both files.
- The `rescue _ -> false` in `seo_no_index?/0` is reasonable defensiveness over a settings read.
- Ordering of the `cond` (noindex before `flat_mode?`) correctly covers both flat and index modes, and `get_sources/0` is only evaluated in the non-noindex branches.
- The two changed private functions have no external callers; the `:all | {:whitelist, [Regex.t()]}` contract is internally consistent.

timujinne:
Applied fixes for the review findings in `9adb66b8`:

- **Finding 1 (HTML sitemap not gated) — fixed.** `generate_html/1` now short-circuits under `seo_no_index?()`, rendering an empty HTML sitemap and bypassing the cache, so `?format=html` no longer advertises URLs or serves a stale cached page while noindex is active.
- **Finding 2 (on-disk staleness) — fixed.** `SEO.update_no_index/1` now calls `Generator.invalidate_and_regenerate/0` (guarded), so toggling `seo_no_index` invalidates + regenerates `/sitemap.xml` instead of serving the stale disk file. Covers both directions (enable and disable).
- **Nitpick 3 — added** a comment documenting the deliberate parity with the robots meta (keyed on `no_index_enabled?/0`, not `module_enabled?/0`).
- **Nitpick 4 — done.** Empty-sitemap log lowered to `:debug`.
- **Nitpick 5 — documented** the `<urlset>` (vs `<sitemapindex>`) shape under noindex in the `do_generate_no_index/2` comment.

`mix compile --warnings-as-errors`, `mix credo --strict`, and `mix dialyzer` pass (via the pre-commit hook).

timujinne (commented):
## Review: `Add reserved_route_prefixes/0 module callback` (commit `e9509b4`)

**Verdict: APPROVE WITH COMMENTS.** Stage 1 (spec compliance) passes cleanly — the callback is wired identically to `sitemap_sources/0` at all four sites, the aggregator uses `safe_call/3`, and the tests are correct and self-cleaning. Stage 2 surfaces one load-bearing semantic concern (the `enabled_modules/0` gating choice) and two robustness gaps in the aggregator that matter more here than in the helper this mirrors, because the intended consumer runs on a request-dispatch path.

### Stage 1 — Spec compliance

All three wiring points are present and match the `sitemap_sources/0` precedent exactly: `@callback reserved_route_prefixes() :: [String.t()]`, the `@optional_callbacks` entry, the `__using__` default `def reserved_route_prefixes, do: []`, and the `defoverridable` entry. A module that `use PhoenixKit.Module` and never implements the callback gets the `[]` default and is overridable, so no missing-`@impl`/dialyzer noise. `all_reserved_route_prefixes/0` delegates through `safe_call/3`, so a module raising inside its own implementation is caught and defaulted to `[]`. The 3 new tests use `register/1` + `unregister/1` correctly and wrap cleanup in `try/after` — strictly more robust against a failing assertion leaking process-global registry state.

### Findings

**IMPROVEMENT-HIGH — `enabled_modules/0` gating is the wrong analogy for collision-protection semantics.** The docstring justifies iterating `enabled_modules/0` by "mirroring `all_sitemap_sources/0`," but the two callbacks protect against different things. `sitemap_sources/0` emits content — gating on enabled is obviously right. `reserved_route_prefixes/0` protects a name collision against another module's DB-driven dispatch, and that collision does not depend on the reserving feature's runtime toggle. Concrete failure: the motivating case is a host that literally wrote `live "/legal", LegalLive` in its router — that route is compiled regardless of `Legal.enabled?()` (a runtime setting can't be consulted at router-compile time). If an admin toggles Legal off while a stale Publishing group slug `"legal"` still exists in the DB, the reservation immediately drops out, Publishing's catch-all reclaims `/legal`, and the compiled host route is hijacked again — reintroducing the exact bug this feature exists to fix, scoped to the disabled window. Recommend switching to `all_modules/0` so an installed package's reservation survives runtime toggles (the compiled-route conflict is a property of the package being present, not of a Settings flag).

**IMPROVEMENT-MEDIUM — Returned segments are trusted verbatim; a `"/legal"` typo silently defeats the reservation.** The docstring documents the contract (no leading/trailing slash) but nothing normalizes or validates it. A module author who returns `["/legal"]` produces a value the consumer's literal path-segment comparison will never match — the reservation silently becomes a no-op. Consider normalizing in the aggregator: `Enum.map(&String.trim(&1, "/"))` before `Enum.uniq/1`.

**IMPROVEMENT-MEDIUM — `Enum.flat_map` over `safe_call` crashes on a non-list return.** `safe_call/3` only rescues exceptions raised inside the module's function; it returns the value verbatim otherwise. If a module implements `def reserved_route_prefixes, do: "legal"` (bare string) or returns `nil`, `Enum.flat_map/2` raises `Protocol.Enumerable not implemented`, uncaught by `safe_call`. Inherited from `all_sitemap_sources/0` so not a regression, but the impact differs here: the documented consumer calls this on a per-request dispatch path, where an uncaught raise can 500 every request. Recommend `List.wrap/1` + `Enum.filter(&is_binary/1)` around each module's return value.

**NITPICK — the "returns a list of strings" test only exercises `is_list/1`.** No enabled core module contributes a prefix by default, so the per-element `is_binary` assertion loop never runs. Harmless smoke test; real coverage is in the other two tests.

**NITPICK — the callback `@doc` could state the passivity more bluntly** — declaring a prefix has zero effect unless a dispatcher actively consults `all_reserved_route_prefixes/0`; one sentence would remove ambiguity for a skimming reader.

### What's genuinely fine

The `@spec`/`@callback` types are correct. The optional-callback + `defoverridable` mechanics are a faithful 1:1 copy of an established, working pattern. The tests clean up correctly and are more defensive than the file's prior convention.

PR #615, opened 2026-07-03
Prevent RouterDiscovery from masking richer sitemap entries and leaking internal routes

timujinne:
## Summary

Two RouterDiscovery hygiene fixes surfaced by the same production sitemap audit as #614 (hydroforce.ee):

1. **RouterDiscovery no longer masks richer entries from content sources.** The generator deduplicated same-`loc` entries with `Enum.uniq_by(& &1.loc)`, which keeps whichever entry comes first — an ordering accident, not a preference. In production this meant `/cemented-carbides` and `/legal` (also emitted by the Entities/Publishing sources with `canonical_path`, hreflang alternates and priority 0.7–0.8) were served as bare RouterDiscovery entries: priority 0.5, no hreflang. Dedupe now scores entry richness (RouterDiscovery always loses a same-loc collision; among content sources, canonical_path/alternates/priority win ties).

2. **Internal routes are excluded by default.** `@default_exclude_patterns` gains `"^/__"` (double-underscore internal route convention, e.g. `__hf_news` — a WebSocket-JOIN alias that leaked into the production sitemap as a duplicate of `/news` and was flagged by Search Console as "Duplicate, Google chose different canonical") and `"^/maintenance$"` (PhoenixKit's reserved maintenance route, which is a locale-redirect shim — Search Console flagged it as a redirect/404 chain).

## Relationship to #614

Complements #614 and implements follow-ups its description explicitly lists as out of scope ("RouterDiscovery entries carry no hreflang alternates … can still list redirects/404s"). There is textual overlap in `generator.ex` / `router_discovery.ex` — happy to rebase whichever lands second.

## Test plan

- [x] `mix quality` pre-commit gate: `compile --warnings-as-errors`, `credo --strict`, `dialyzer`, `format --check-formatted` — all green
- [x] New integration test `test/integration/sitemap/generator_dedupe_test.exs` (165 lines): rich-vs-router dedupe, default-pattern exclusion
- [ ] Integration test needs PostgreSQL (no CI in repo) — please run on a machine with a test DB

Behavior note: user-supplied `sitemap_router_discovery_exclude_patterns` in settings still fully replaces the defaults — no behavior change for existing overrides.

### DIFF
--- a/lib/modules/sitemap/generator.ex
+++ b/lib/modules/sitemap/generator.ex
@@ -468,7 +468,7 @@ defmodule PhoenixKit.Modules.Sitemap.Generator do
       Logger.debug("Sitemap: Collected #{length(entries)} entries from #{inspect(source_module)}")
       entries
     end)
-    |> Enum.uniq_by(& &1.loc)
+    |> dedupe_by_loc()
     |> Enum.sort_by(& &1.loc)
   end
 
@@ -557,7 +557,7 @@ defmodule PhoenixKit.Modules.Sitemap.Generator do
         entry
       end
     end)
-    |> Enum.uniq_by(& &1.loc)
+    |> dedupe_by_loc()
     |> Enum.sort_by(& &1.loc)
   end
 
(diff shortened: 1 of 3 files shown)


timujinne (commented):
## Verdict: APPROVE with minor changes

The core fix is correct and well-motivated. The richness-based dedupe genuinely resolves the reported production bug (bare RouterDiscovery entries masking hreflang-rich content entries for `/legal` and `/cemented-carbides`), and the crucial multilingual interaction is safe.

### Confirmed correct (the important trace)

`collect_multilingual_entries/3` computes `alternates_by_canonical` from `all_entries |> Enum.filter(& &1.canonical_path)` **before** the final `dedupe_by_loc`. RouterDiscovery entries are built with `canonical_path: nil` (`build_entry/2` never sets it), so they're filtered out of the alternates computation entirely. Leftover un-deduped RD duplicates (RouterDiscovery ignores the `language` opt and is re-collected identically per language task) therefore cannot pollute the alternates map. `dedupe_by_loc` then correctly drops those RD copies in favor of the alternates-bearing content entry — traced the adversarial case (an RD entry colliding with a properly localized content entry) and it resolves correctly. This is the case that was silently broken under `Enum.uniq_by/2`, and it is fixed.

### Findings

**IMPROVEMENT-MEDIUM — `^/maintenance$` won't exclude localized maintenance routes.** `lib/modules/sitemap/sources/router_discovery.ex:83`. The pattern is fully anchored, so it matches `/maintenance` only — not `/en/maintenance`, `/et/maintenance`, or `/maintenance/`. Every other "should-not-be-indexed" default in this list (`/users/log-in`, `/checkout`, `/cart`, …) is deliberately unanchored so it matches under any locale prefix. Confirmed: `Regex.match?(~r/^\/maintenance$/, "/en/maintenance")` → `false`. Recommend `"/maintenance$"` (drop the `^`) for consistency, or `"/maintenance(/|$)"` if a trailing slash is possible.

**IMPROVEMENT-MEDIUM — Test suite exercises only the single-language dedupe path.** `test/integration/sitemap/generator_dedupe_test.exs`. All four tests call `Generator.collect_all_entries/2`, which in the default single-language test DB takes the `collect_single_language_entries` branch. The multilingual pipeline (`collect_multilingual_entries/3`) — where the production regression actually manifested and where the alternates-computed-before-dedupe ordering matters — is never exercised. Add a test that enables ≥2 languages so the RD-masks-hreflang scenario is validated end-to-end through the real multilingual path.

**IMPROVEMENT-MEDIUM — Generic `Generator` now hardcodes a source-specific atom.** `entry_richness(%UrlEntry{source: :router_discovery}), do: 0` puts source-specific knowledge into the otherwise source-agnostic generator, keyed on an atom that must stay in sync with `RouterDiscovery.source_name/0`. This is a defensible pragmatic choice and is load-bearing (a pure metadata formula would let RD beat a legitimate low-priority content entry, violating "RD always loses"). A cleaner long-term shape is an optional `Source` callback (e.g. an authoritative/rank flag) so the generator ranks by capability rather than a hardcoded atom. Not required for merge; worth a `# TODO`.

**NITPICK — Non-deterministic tie-break between two content sources.** `Enum.max_by/2` returns the first maximal element on ties, and for same-`loc` entries that tie exactly on richness, "first" depends on `Task.async_stream(ordered: false)` ordering — pre-existing behavior (uniq_by had the same issue), essentially moot since two content sources emitting an identical `loc` with identical richness is a misconfiguration. If deterministic output is desired: `Enum.max_by(group, &{entry_richness(&1), &1.source})`.

**NITPICK — `UrlEntry.parse_priority/1` ignores integer priorities.** `entry_richness/1` calls `parse_priority/1`, which falls through to `nil` for integers, whereas the sibling `normalize_priority/1` does handle integers. Harmless today (no in-tree source emits an integer priority, and RD loses regardless), but the asymmetry is a latent trap worth aligning.

### Spot-checks that passed
- Both new regex patterns compile cleanly and match as intended.
- `entry_richness/1` doesn't crash on any legitimate `UrlEntry` shape; both `nil`/`""`/`[]` "empty" checks are covered.
- RD-alone survival (no competing source) is stable — no churn.

PR #616, opened 2026-07-03
Honor seo_no_index on host-app public LiveViews

timujinne:
## Summary

`seo_no_index` never reaches the noindex meta tag on a host application's own public LiveViews.

The only place that assigned `:seo_no_index` was `LayoutWrapper.app_layout_inner/1`, which wraps PhoenixKit's admin/plugin views. A host app's public LiveView renders through the host's own root layout — `root.html.heex`'s conditional `<meta name="robots" content="noindex,nofollow">` block is there, but the assign is never set, so the directive silently does nothing.

Real-world case (same audit as #614): a staging/dev deployment of a PhoenixKit host had `seo_no_index=true` set and was nevertheless fully indexed by Google — including unreleased content that then surfaced on the production domain in Search Console as "Page with redirect" entries.

## Fix

`set_routing_info/3` (the `handle_params` hook attached by `on_mount :phoenix_kit_mount_current_scope`) now does `assign_new(:seo_no_index, fn -> SEO.no_index_enabled?() end)`. That hook already runs for every LiveView mounted through PhoenixKit's on_mount chain — admin and host-app public pages alike — and already owns the sibling concern (`:url_path`). `assign_new/3` keeps it a no-op when `LayoutWrapper` already set the value.

Implements the host-layout follow-up noted in #614 ("`seo_no_index` reaching the real root layout for host apps that supply their own layout").

## Test plan

- [x] `mix quality` pre-commit gate: `compile --warnings-as-errors`, `credo --strict`, `dialyzer`, `format --check-formatted` — all green
- [x] Regression test `test/phoenix_kit_web/users/auth_seo_no_index_test.exs` with a stand-in host-app public LiveView (mounts through the on_mount chain, renders its own markup, never calls LayoutWrapper)
- [ ] Test needs PostgreSQL (no CI in repo) — please run on a machine with a test DB

### DIFF
--- a/lib/phoenix_kit_web/users/auth.ex
+++ b/lib/phoenix_kit_web/users/auth.ex
@@ -44,6 +44,7 @@ defmodule PhoenixKitWeb.Users.Auth do
   alias PhoenixKit.Modules.Languages
   alias PhoenixKit.Modules.Languages.DialectMapper
   alias PhoenixKit.Modules.Maintenance
+  alias PhoenixKit.Modules.SEO
   alias PhoenixKit.Users.Auth
   alias PhoenixKit.Users.Auth.{Scope, User}
   alias PhoenixKit.Users.Permissions
@@ -721,6 +722,14 @@ defmodule PhoenixKitWeb.Users.Auth do
       socket
       |> Phoenix.Component.assign(:url_path, path)
       |> maybe_update_locale_from_params(params)
+      # This hook fires on `handle_params` for every LiveView mounted through
+      # PhoenixKit's on_mount chain (admin and host-app public pages alike),
+      # so it is the one place that can guarantee `:seo_no_index` reaches
+      # root.html.heex's noindex meta tag before the first render — unlike
+      # LayoutWrapper.app_layout_inner/1, which only wraps admin/plugin views
+      # and never runs for a host app's own public LiveViews. assign_new is
+      # a no-op if LayoutWrapper already set it, so this is safe either way.
+      |> Phoenix.Component.assign_new(:seo_no_index, fn -> SEO.no_index_enabled?() end)
 
     {:cont, socket}
   end
--- a/test/phoenix_kit_web/users/auth_seo_no_index_test.exs
+++ b/test/phoenix_kit_web/users/auth_seo_no_index_test.exs
@@ -1,0 +1,58 @@
+defmodule PhoenixKitWeb.Users.AuthSeoNoIndexTest do
+  @moduledoc """
+  Regression test for `seo_no_index` not reaching a host application's own
+  public LiveViews. `PhoenixKitWeb.Components.LayoutWrapper.app_layout_inner/1`
+  is the only place that used to assign `:seo_no_index` — but it only wraps
+  PhoenixKit's own admin/plugin views. A host app's public LiveView (its own
+  layout, not routed through LayoutWrapper) never got the assign, so
+  `root.html.heex`'s `<meta name="robots" content="noindex,nofollow">` never
+  rendered for it even with the directive enabled (observed on Hydroforce's
+  dev environment).
+
+  `PublicHostAppLive` below stands in for such a view: it mounts through
+  PhoenixKit's `:phoenix_kit_mount_current_scope` on_mount (as a host app
+  would, e.g. for current_user/locale support) but renders with its own
+  markup, never calling `LayoutWrapper.app_layout`.
+  """
+
+  use PhoenixKitWeb.ConnCase, async: false
+
+  alias PhoenixKit.Modules.SEO
+
+  defmodule PublicHostAppLive do
+    @moduledoc false
+    use Phoenix.LiveView
+
+    on_mount {PhoenixKitWeb.Users.Auth, :phoenix_kit_mount_current_scope}
+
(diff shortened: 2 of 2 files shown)


timujinne (commented):
## Verdict: REQUEST CHANGES

The production one-liner in `auth.ex` is correct and well-placed — for real routed LiveViews it reliably lands `:seo_no_index` on `socket.assigns` before first render, which is exactly what `root.html.heex` needs. But the accompanying regression test cannot pass as written: it raises during mount, independent of the database, and would break `mix test` the moment anyone runs it.

### BUG-HIGH — The regression test raises during mount; `live_isolated/3` cannot exercise this on_mount

`test/phoenix_kit_web/users/auth_seo_no_index_test.exs:20-52`. `PublicHostAppLive` declares `on_mount {PhoenixKitWeb.Users.Auth, :phoenix_kit_mount_current_scope}`. That callback calls `mount_phoenix_kit_current_user/2`, which unconditionally attaches a `:handle_params` hook (`attach_hook(socket, :current_page, :handle_params, &set_routing_info/3)`). `Phoenix.LiveView.Lifecycle.attach_hook/4` raises when `socket.router` is `nil` for the `:handle_params` stage. `live_isolated/3` on a bare `ConnCase` conn derives its router from `Phoenix.Controller.router_module(conn)`, which is `nil` because the conn never matched a route.

Confirmed both facts empirically against a running app:
- `Phoenix.LiveView.attach_hook(%Phoenix.LiveView.Socket{router: nil}, :current_page, :handle_params, fn _,_,s -> {:cont,s} end)` raises `"cannot attach hook with id :current_page on :handle_params because the view was not mounted at the router with the live/3 macro"` (the exact id PhoenixKit uses).
- `Phoenix.Controller.router_module(Phoenix.ConnTest.build_conn())` → `KeyError`, which `live_isolated` rescues to `nil`.

**Failure scenario:** running `mix test test/phoenix_kit_web/users/auth_seo_no_index_test.exs` raises in the isolated mount before either assertion is reached — with or without PostgreSQL. Both test cases error out.

**Fix:** don't use `live_isolated/3` for a LiveView that goes through this on_mount. Mount the stand-in at a real route (add a `live "/…", PublicHostAppLive` to a test-only router/the test endpoint's router) and drive it with `live(conn, path)` so `socket.router` is non-nil.

### IMPROVEMENT-HIGH — Even if it ran, the test proves the wrong thing

The probe div reads `assigns[:seo_no_index]` inside `PublicHostAppLive`'s own `render/1`; the assertions only prove "the socket has the assign" — they never render through `root.html.heex`, so they don't prove `<meta name="robots" content="noindex,nofollow">` actually reaches `<head>`. Since the reported bug is precisely "the meta tag never rendered," the regression test should mount through the real root layout and assert on the rendered meta tag (or its absence).

### IMPROVEMENT-MEDIUM — Threading the assign is necessary but not sufficient for host-owned root layouts

The only two places that consume `:seo_no_index` are PhoenixKit's own `root.html.heex` and `LayoutWrapper`'s standalone branch. A host that supplies its own root layout module (the common production case) will still emit no noindex tag unless it explicitly reads `assigns[:seo_no_index]` and renders the meta itself. This fix fully resolves the bug only for hosts using PhoenixKit's `root.html.heex`; for host-owned layouts it's a prerequisite, not a complete fix. Recommend documenting: hosts with a custom root layout must add the `<meta name=robots>` block reading `@seo_no_index` themselves.

### IMPROVEMENT-MEDIUM — Non-LiveView (controller-rendered) host pages are not covered

This hook only fires for LiveViews. Plain controller-rendered public pages never run it, and there's no conn-level plug that assigns `:seo_no_index`. A staging host serving any non-LiveView public page stays indexable even with the directive on. Either document this as a known non-goal or add a small plug that sets the assign conn-side.

### IMPROVEMENT-MEDIUM — Once-per-mount semantics (acceptable, but worth a note)

`assign_new/3` evaluates the fun only on the first `handle_params` and is a no-op thereafter for the socket's lifetime. Consequences: (1) a mid-session toggle of `seo_no_index` won't reach an already-connected LiveView until a remount/reload; (2) `LayoutWrapper.app_layout_inner/1`'s own `assign_new(:seo_no_index, …)` becomes a permanent no-op for routed admin LVs once this hook sets the key first — previously that function component recomputed the value on every render. Both are fine for a rarely-toggled staging flag, but worth a one-line comment noting the shift from "recompute per-render" to "compute once per mount."

### NITPICK — Performance concern is unfounded (noting it's fine)

`SEO.no_index_enabled?/0` reads the ETS-backed settings cache, not the DB, and `assign_new` only evaluates once per socket. No meaningful site-wide load added.

### Things that are genuinely fine
- `assign_new/3` on a socket is the correct call (the sibling `assign(:url_path, path)` already operates on the same socket).
- The hook is genuinely universal for routed LiveViews — both `:phoenix_kit_mount_current_scope` and `:phoenix_kit_mount_current_user` route through the same function, so the fix works for its target case.
- Precedence is correct — `assign_new` respects a value `LayoutWrapper` or a host set earlier.
- Piggybacking on the one hook every PhoenixKit LiveView already attaches is pragmatic and consistent with existing patterns.
- `on_exit` cleanup resets the flag correctly.

PR #620, opened 2026-07-07
Fix broken seo_no_index regression test from #616

timujinne:
## Summary

Fixes the regression test merged with #616 (`Thread seo_no_index to host-app public LiveViews via routing hook`), which raises during mount and breaks `mix test` — confirmed the production fix itself is sound; only its test was broken.

## The bug

`test/phoenix_kit_web/users/auth_seo_no_index_test.exs`'s `PublicHostAppLive` declares `on_mount {PhoenixKitWeb.Users.Auth, :phoenix_kit_mount_current_scope}`, whose `handle_params` hook calls `attach_hook/4` on `:handle_params` — which requires a non-nil `socket.router`. `live_isolated/3` never sets one (`Phoenix.LiveViewTest.__isolated__/4` puts `phoenix_endpoint` into the conn but never `phoenix_router`), so it raises *"the view was not mounted at the router with the live/3 macro"* before either assertion runs. Both test cases error out unconditionally, independent of the database.

## The fix

- Moved `PublicHostAppLive` to `test/support/public_host_app_live.ex` — a real compiled module (a `defmodule` nested inside a `.exs` file only exists at test-runtime, so it can't be referenced from a router that compiles beforehand).
- Added a `Mix.env() == :test`-gated route for it in `PhoenixKitWeb.Router` (the library's own dev/test-only router — its moduledoc already states as much). Compiles out entirely for `:dev`/`:prod` (verified).
- The test now drives it via a real HTTP `GET` instead of `live_isolated/3`, so `socket.router` is non-nil and the on_mount's `handle_params` hook attaches without raising.

This also addresses a second, non-fatal issue from the original PR's review: the old test asserted against the raw `:seo_no_index` **assign** inside the LiveView's own render, which only proves the socket carries the assign — not that the noindex meta tag actually reaches `<head>`. The rewritten test asserts against the real disconnected/initial HTML response (what a crawler would see), exercising the actual root-layout render path the reported bug was about.

## Test plan

- [x] `mix compile --warnings-as-errors` — clean in both `dev` and `prod` (the test-gated route correctly compiles out of `prod`)
- [x] `mix credo --strict` — 0 issues
- [x] `mix dialyzer` — passed (via pre-commit hook)
- [ ] `mix test test/phoenix_kit_web/users/auth_seo_no_index_test.exs` — **could not execute locally.** `PhoenixKit.DataCase`/`ConnCase`-based tests fail to load their compiled `test/support` beams in this sandbox regardless of the file under test — reproduced against untouched, unrelated files too (`test/integration/storage/scope_test.exs`, `test/integration/users/profile_test.exs`), so this is a pre-existing environment limitation, not specific to this change. Will run in CI.
- [x] Verified the underlying mechanism directly on a real host app instead: toggling `seo_no_index` and hitting a real public host LiveView route (Hydroforce's `/legal`, which mounts through this exact same on_mount) shows the noindex meta present/absent in the actual HTTP response exactly as expected in both states.

### DIFF
--- a/lib/phoenix_kit_web/router.ex
+++ b/lib/phoenix_kit_web/router.ex
@@ -38,6 +38,16 @@ defmodule PhoenixKitWeb.Router do
   # backing test/phoenix_kit_web/users/auth_seo_no_index_test.exs. Needs a
   # real route (not live_isolated/3) because :phoenix_kit_mount_current_scope
   # attaches a :handle_params hook that requires a non-nil socket.router.
+  #
+  # Known, accepted trade-off: the Hex package doesn't ship test/, so a
+  # consumer compiling this dep under MIX_ENV=test (e.g. their own `mix
+  # test`) builds this block referencing a PublicHostAppLive module that
+  # isn't present for them. This is inert, not a build failure — Phoenix's
+  # `live` macro only stores the module as an atom and resolves it at
+  # request time, and this router is never mounted by parent apps (see
+  # moduledoc above) — but it is a latent dead reference. A dedicated
+  # test-only Endpoint+Router pair under test/support/ would avoid it
+  # entirely; not worth the extra scaffolding for a single probe route.
   if Mix.env() == :test do
     scope "/", PhoenixKitWeb.Test do
       pipe_through :browser
--- a/lib/phoenix_kit_web/router.ex
+++ b/lib/phoenix_kit_web/router.ex
@@ -34,6 +34,18 @@ defmodule PhoenixKitWeb.Router do
   # PhoenixKit routes - main integration point
   phoenix_kit_routes()
 
+  # Test-only: a routable stand-in for a host app's own public LiveView,
+  # backing test/phoenix_kit_web/users/auth_seo_no_index_test.exs. Needs a
+  # real route (not live_isolated/3) because :phoenix_kit_mount_current_scope
+  # attaches a :handle_params hook that requires a non-nil socket.router.
+  if Mix.env() == :test do
+    scope "/", PhoenixKitWeb.Test do
+      pipe_through :browser
+
+      live "/__test/seo-no-index-probe", PublicHostAppLive
+    end
+  end
+
   defp ensure_session_uuid(conn, _opts) do
     case Plug.Conn.get_session(conn, :phoenix_kit_session_uuid) do
       nil ->
--- a/test/phoenix_kit_web/users/auth_seo_no_index_test.exs
+++ b/test/phoenix_kit_web/users/auth_seo_no_index_test.exs
@@ -9,50 +9,51 @@ defmodule PhoenixKitWeb.Users.AuthSeoNoIndexTest do
   rendered for it even with the directive enabled (observed on Hydroforce's
   dev environment).
 
-  `PublicHostAppLive` below stands in for such a view: it mounts through
-  PhoenixKit's `:phoenix_kit_mount_current_scope` on_mount (as a host app
-  would, e.g. for current_user/locale support) but renders with its own
-  markup, never calling `LayoutWrapper.app_layout`.
+  `PhoenixKitWeb.Test.PublicHostAppLive` (test/support/public_host_app_live.ex)
+  stands in for such a view: it mounts through PhoenixKit's
+  `:phoenix_kit_mount_current_scope` on_mount (as a host app would, e.g. for
+  current_user/locale support) but renders with its own markup, never calling
+  `LayoutWrapper.app_layout`. It's routed at a real, test-only path (see
+  `PhoenixKitWeb.Router`) rather than mounted via `live_isolated/3`, because
+  the on_mount attaches a `:handle_params` hook that requires a non-nil
+  `socket.router` — `live_isolated/3` never provides one and raises ("the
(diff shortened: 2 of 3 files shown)


timujinne (commented):
## Verdict: APPROVE — the fix is correct and the rewritten test will pass in CI; one design-hygiene issue worth documenting, otherwise clean.

Independently re-derived every load-bearing claim from source rather than trusting the PR description. The diagnosis of #616's failure is accurate, the move to `test/support/` is technically necessary, and the new HTTP-GET test faithfully exercises the real root-layout render path the bug was about.

### Spec-compliance verification (all claims hold)

- **`live_isolated/3` never sets a router — confirmed.** `Phoenix.LiveViewTest.__isolated__/4` only does `put_in(conn.private[:phoenix_endpoint], ...)`; no `phoenix_router` is ever written. The `:phoenix_kit_mount_current_scope` on_mount attaches a `:handle_params` hook that requires a non-nil `socket.router`, so the old test raised on mount before either assertion — independent of the DB, exactly as claimed.

- **The nested `.exs` module genuinely could not be routed — confirmed.** `mix.exs`'s `elixirc_paths(:test) = ["lib", "test/support"]` means `lib/router.ex` and `test/support/*.ex` compile together during `mix compile`, whereas a `defmodule` nested in a `*_test.exs` only comes into existence when the test runner loads that file at runtime — after compilation. Moving it to `test/support/public_host_app_live.ex` is the correct fix.

- **The route compiles out of dev/prod — confirmed.** `Mix.env()` is evaluated at compile time; the guard is simply absent from the AST in `:dev`/`:prod`, matching the existing `if code_reloading? do` precedent in `endpoint.ex`.

- **The GET faithfully hits `root.html.heex` — confirmed, and this was the highest-risk item.** The new scope's `pipe_through :browser` includes `plug :put_root_layout`, which resolves to `{PhoenixKitWeb.Layouts, :root}`. A GET on the live route dispatches through the full pipeline → static render → `handle_params` (the attached hook assigns `:seo_no_index`) → rendered inside `root.html.heex`. It does not accidentally bypass the root layout.

- **The literal `<meta name="robots" content="noindex,nofollow">` matches byte-for-byte — confirmed.** `meta` is in HEEx's void-element set, so the engine emits it without a self-closing slash, matching the assertion exactly. The `refute` in the disabled case is also safe — both `robots` and the distinct `googlebot` meta are gated behind the same `if assigns[:seo_no_index]`.

### Will the test actually pass? Yes — explicit trace

End-to-end: `SEO.enable_no_index()` writes the setting synchronously in the test process; the GET runs in the same process/sandbox connection, so the subsequent read sees the write. Mount → on_mount attaches the hook → `handle_params` fires on the disconnected render → `assign_new(:seo_no_index, true)` → `root.html.heex` renders both the `robots` meta and the probe div's `data-seo-no-index="true"`. Both assertions pass; the disabled case is the mirror image. The "couldn't run `mix test` locally" note reads as a plausible sandbox limitation, not cover for a latent bug.

### Findings

**IMPROVEMENT-MEDIUM — a Hex consumer's `MIX_ENV=test` compile builds a route to a module that isn't in the package.** The Hex package `files` list ships `lib priv mix.exs README LICENSE CHANGELOG` — not `test/`. When a parent app compiles `phoenix_kit` under `MIX_ENV=test`, the `if Mix.env() == :test` block compiles and registers a route to `PhoenixKitWeb.Test.PublicHostAppLive` — but `test/support/public_host_app_live.ex` was never shipped, so that module doesn't exist for the consumer. This does **not** break their build (`Phoenix.Router`'s `live` macro only stores the module as an atom and resolves it at request time — no compile-time reference, no xref failure), and `PhoenixKitWeb.Router` is explicitly dev/test-only and never mounted by parent apps — so the dead route is functionally inert. Still, it's a latent dead reference baked into every consumer's test-env build of the library. A lower-footprint alternative that keeps `lib/` untouched entirely: a tiny dedicated test Endpoint+Router pair under `test/support/`. The chosen approach is defensible given the existing `code_reloading?` precedent, but the isolated-double pattern is the cleaner call for shared library code — noting for a future cleanup pass rather than blocking this fix.

**NITPICK — assertions don't cover the sibling `googlebot` directive.** `root.html.heex` emits a parallel `<meta name="googlebot" content="noindex,nofollow">` under the same guard, unasserted.

**NITPICK — `PhoenixKitWeb.Test.PublicHostAppLive` naming.** Introduces a `PhoenixKitWeb.Test.*` namespace solely to host this fixture — fine as-is, worth a clearer shared namespace if more routed test fixtures accrue.

No dialyzer/credo/spec concerns in the new files.

timujinne:
Applied both suggestions from the review in `0e78aa3b`:
- Added assertions on the sibling `googlebot` meta directive, pinning the full noindex head block.
- Documented the known, accepted trade-off of the `Mix.env() == :test` route (inert dead reference in consumers' own `MIX_ENV=test` builds, since `test/` isn't shipped in the Hex package) directly in the router comment.

Left the router-based approach as-is rather than building a dedicated test-only Endpoint+Router pair — per the review, it's confirmed harmless and matches the existing `code_reloading?` precedent in this file; the extra scaffolding isn't worth it for a single probe route.

PR #625, opened 2026-07-08
Add fine-grained sub-permissions; make Admin genuinely permission-gated (+ V141 calendar events)

mdon:
## Summary

Two related changes to the permission system, plus the events table for the new `phoenix_kit_calendar` module (its first consumer).

### 1. Fine-grained sub-permissions

Modules can now declare permissions *within* their module key via `permission_metadata/0`:

```elixir
def permission_metadata do
  %{key: "calendar", label: "Calendar", icon: "hero-calendar-days",
    description: "Personal calendars",
    sub_permissions: [
      %{key: "view_others", label: "View others' calendars", description: "..."},
      %{key: "edit_others", label: "Edit others' calendars", description: "..."}
    ]}
end
```

- Stored as composed dotted keys (`"calendar.view_others"`) in the existing `phoenix_kit_role_permissions` table — **V140** widens `module_key` to VARCHAR(120).
- The base key keeps its exact current meaning (module/page access); sub-keys are additive grants a module checks itself via the new **`Scope.can?/2`**, which also requires the parent module to be enabled — a scope snapshotted before a module was disabled can't keep authorizing its actions.
- **Sub-implies-base invariant enforced at the context layer**, not just the UI: granting a sub auto-grants its base in the same transaction, revoking a base cascades its subs off, and `set_permissions/3` normalizes the desired set. No code path can persist an orphan sub-key row.
- The permissions matrix and the roles editor render indented sub-rows; the "you can only grant keys you hold" guard now also covers the base key a sub-grant implies (closes an escalation edge where an editor holding only the sub could smuggle in a base grant).
- Registry-driven parent↔sub resolution (never string-splitting); both key parts validated against the existing `~r/^[a-z][a-z0-9_]*$/`, so a composed key has exactly one dot and can't collide with custom keys (which forbid dots).

### 2. ⚠️ Behavior change: Admin is now genuinely gated by its permission rows

`enforce_admin_view_permission` had a `system_role?` bypass (rode along in 65ab7ef5, whose subject was disabled-module centralization; no tests pinned it). The result was inconsistent: an Owner revoking a key from the Admin role was honored by the sidebar, the `require_module_access` plug, and the mid-session PubSub kick — but a fresh mount let the Admin right back in. Revoking Admin permissions was effectively cosmetic.

This PR removes the Admin half of the bypass:

- **Admin = a role that DEFAULTS to all permissions but is genuinely Owner-editable.** Owner behavior is unchanged (its scope holds every key by construction). Views that resolve to no permission key still allow system roles only (unchanged, documented).
- The Admin zero-rows full-access fallback now applies only to genuinely unseeded installs (no permission rows at all) — previously, revoking Admin's *last* key ironically restored full access.
- To keep "Admin defaults to everything" without the bypass, a startup task **auto-grants Admin any permission key it has never been granted** (newly installed modules, new sub-permissions), guarded by per-key settings flags so an Owner's revocation sticks across restarts. This doubles as the repair path for installs whose V53 seeding predates newer modules — verified on a real upgrade boot (Admin 0 → 42 keys, then a revocation stayed revoked across restart).
- **Install impact:** on installs where an Owner had revoked keys from Admin, those revocations now actually block fresh mounts. That is the intended semantic; flagging it since it changes what Admin can reach.

### 3. V141 — `phoenix_kit_calendar_events`

Table for the new `phoenix_kit_calendar` module (personal calendar per user; the reference consumer of sub-permissions). Exclusive-end pairs (timed UTC pair / all-day DATE pair, CHECK-enforced against the `all_day` flag), `owner_uuid` FK CASCADE, confirmed/cancelled status.

## Testing

- 37 new tests: sub-key semantics + registry validation (24 unit), grant/revoke/set cascades, auto-grant idempotency + flag behavior, fallback scoping, Owner invariance (13 DB).
- Full suite: 42 failures — identical to clean `main` (stash-verified baseline; 31 multi-session + 11 sitemap-settings pre-existing). Zero new failures.
- `mix precommit` green (format, compile --warnings-as-errors, credo --strict, dialyzer).
- Browser-verified on the parent app: matrix sub-rows with live grant/revoke cascades, Admin auto-grant on boot, calendar module end-to-end.

### DIFF
--- a/lib/phoenix_kit/migrations/postgres.ex
+++ b/lib/phoenix_kit/migrations/postgres.ex
@@ -529,7 +529,15 @@ defmodule PhoenixKit.Migrations.Postgres do
   - Replaces unique index with partial index (slug-mode only, WHERE slug IS NOT NULL)
   - Adds unique index on `(group_uuid, post_date, post_time)` for timestamp-mode posts
 
-  ### V139 - Dashboard `config` column ⚡ LATEST
+  ### V140 - Wider role-permission keys ⚡ LATEST
+  - Widens `phoenix_kit_role_permissions.module_key` from `VARCHAR(50)` to
+    `VARCHAR(120)` so fine-grained sub-permissions can be stored as composed
+    dotted keys (`"calendar.view_others"` — base and sub parts are each
+    capped at 50 chars, so a composed key can reach 101).
+  - Rollback deletes rows over 50 chars (sub-permission grants are additive
+    and re-grantable) before narrowing the column back.
+
+  ### V139 - Dashboard `config` column
   - Adds a JSONB `config` column (`NOT NULL DEFAULT '{}'`) to
     `phoenix_kit_dashboards` for per-dashboard presentation settings, read and
     written whole like `layout`. Backs the dashboards plugin module.
@@ -1190,7 +1198,7 @@ defmodule PhoenixKit.Migrations.Postgres do
   use Ecto.Migration
 
   @initial_version 1
-  @current_version 139
+  @current_version 140
   @default_prefix "public"
 
   @doc false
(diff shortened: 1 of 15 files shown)


mdon:
Closing for now — will reopen when ready.

PR #643, opened 2026-07-17
Component polish: search-picker on-focus, nav_tabs pulse, multilang auto switch-handler

mdon:
Small component-polish / DX commits.

The first two came out of the projects calendar work (BeamLabEU/phoenix_kit_projects#30 uses both, but degrades gracefully without them); the third came out of the publishing group-name translation work.

## `search_on_focus` attr on the search picker

The SearchPicker hook has always honored `data-search-on-focus` (open the dropdown in browse mode when the empty input is focused/clicked — offer options before any typing), but the component never exposed it: consumers had to know the raw data attribute and pass it through the global rest. It is now a documented boolean attr rendering that same attribute; the raw form keeps working for existing call sites.

## nav_tabs in-flight pulse

Tab buttons that push an event now pulse (`phx-click-loading` → `animate-pulse`) while the switch is round-tripping, so a click on a slower tab is visibly acknowledged instead of appearing dead. Navigation-style tabs are unaffected.

## MultilangForm auto-handles the `"switch_language"` event

`mount_multilang/2` now attaches a `:handle_event` hook (mirroring its existing debounced `:handle_info` hook) that calls `handle_switch_language/2` and halts — so consumers no longer need their own `def handle_event("switch_language", …)` clause. Forgetting that clause previously crashed the LiveView on the first language-tab click, a silent footgun since `mount_multilang` already wired the *other* half of the switching protocol (the `handle_info` timer) automatically.

- Existing consumers' manual clauses keep working (the hook halts first; the clause becomes harmless dead code).
- Opt out with `mount_multilang(socket, auto_switch_language: false)` when a form switches language immediately, without the debounce (catalogue's import screen does this).
- Also documents a storage footgun: `merge_translatable_params/4` / `put_language_data/3` take over the schema's `data` JSONB (restructuring it into a per-language shape), so schemas whose `data` already holds unrelated keys must not point the helper at `data` — use `translatable_field`'s settings-translations mode instead.

## Testing

- `mix test test/phoenix_kit_web/components/core/search_picker_test.exs` — 6 tests, 0 failures (rebased onto current main, full compile clean with `--warnings-as-errors`).
- Browser-verified through the projects module: browse-on-focus lists people via the documented attr, and the project List/Timeline/Calendar tab switches pulse while loading.
- MultilangForm change: compiles clean (`--warnings-as-errors`); language-tab switching verified in-app via the publishing group-name edit form (tabs switch without any per-consumer `switch_language` handler). All existing multilang consumers across the workspace delegate to `handle_switch_language/2`, so the auto-hook is behaviour-preserving for them; the one immediate-switch consumer (catalogue import) opts out.

### DIFF
--- a/lib/phoenix_kit_web/components/core/nav_tabs.ex
+++ b/lib/phoenix_kit_web/components/core/nav_tabs.ex
@@ -69,12 +69,18 @@ defmodule PhoenixKitWeb.Components.Core.NavTabs do
             </span>
           </.link>
         <% else %>
+          <%!-- While the switch round-trips, LiveView tags the clicked tab
+               with phx-click-loading — pulse it so a switch whose content
+               needs server work still gives instant feedback. --%>
           <button
             type="button"
             role="tab"
             phx-click={@on_change}
             phx-value-tab={tab.id}
-            class={["tab gap-2", tab.id == @active_tab && "tab-active"]}
+            class={[
+              "tab gap-2 [&.phx-click-loading]:animate-pulse",
+              tab.id == @active_tab && "tab-active"
+            ]}
           >
             <.icon :if={Map.has_key?(tab, :icon)} name={tab.icon} class="w-4 h-4" />
             {tab.label}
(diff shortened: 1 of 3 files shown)


ddon:
Merged and took into work and release

PR #647, opened 2026-07-18
Send profiles into core + Settings → Email Sending + migration V152

timujinne:
## Send Profiles move into core + Settings → Email Sending + V152

Core half of the ecosystem restructuring (see merge order at the bottom). Send profiles stop being a newsletters-private concept and become core infrastructure every module can send through; the emails/newsletters settings converge into one core-owned **Settings → Email Sending** page.

### Send profiles → core
- `PhoenixKit.Email.SendProfile` schema + `PhoenixKit.Email.SendProfiles` context (moved from phoenix_kit_newsletters, uuid-preserving copy migration), including per-provider `ProviderOptions` and the default-profile invariant (`set_default_send_profile/1` hardened against the two-tab race via an atomic conditional `update_all`).
- Transport resolution: sends route through the active profile's Integration; app-config transport is detect-and-display only.
- Quota/credits surfacing: SES `GetSendQuota` and Brevo `/account` (plan + credits) shown per profile; validators that cannot verify sending say so instead of a bare green tick.

### Settings → Email Sending (single tab seam)
- New `Settings → Email Sending` page: transport detection, default-profile picker, module sections seam — the emails module's sections mount here as live_components (its own Settings tab is removed on the module side).
- Send Profiles CRUD UI + **Send Profiles as its own sidebar sub-entry** (parent-tab pattern, like Media → Dimensions).
- Pages follow the chrome-header canon: titles/subtitles via `page_title`/`page_subtitle` assigns, nested pages use `page_section`/`page_section_path` (clickable breadcrumb); `admin.html.heex` now forwards both — previously a plugin module assigning `page_section` was a silent no-op.

### Migration V152 (single open migration, three sections)
1. `email_send_profiles` — create in core, uuid-preserving copy from newsletters' table, drop the old one.
2. CRM contact lists — `phoenix_kit_crm_lists` (incl. `locale`) + `phoenix_kit_crm_list_members` (denormalized citext member email, `UNIQUE (list_uuid, email)`), contact extensions (`locale`, `opted_out_at`, `consent`).
3. Broadcasts CRM source — `broadcasts.source_type`/`crm_list_uuid` (soft ref, no FK — same pattern as `send_profile_uuid`), `list_uuid` → nullable; `deliveries.recipient_email` (citext), `user_uuid` → nullable. `down/1` intentionally does not restore the NOT NULLs (documented: rollback would fail on rows the feature itself created).

### Fixes worth calling out
- **`pagination_range/2` could OOM the whole BEAM**: an unclamped `current_page` (e.g. `?page=9999999999`) built a descending `9999999997..56` Range — ~10 billion iterations in the component's `for`. Now clamps into `[1, max(total_pages, 1)]` with an explicit `//1` step as a second guard; `pagination_controls/1` additionally gained the `total_pages > 1` outer guard. Regression tests included (reproduced against a live app before fixing).
- `table_default`'s search toolbar rendered no `<form>` when only `on_change` was given — `phx-change` on a formless input dies silently client-side. The form is now always rendered.
- Admin view permission map: five settings LiveViews (Integrations, IntegrationForm, EmailSending, SendProfiles, SendProfileForm) resolved to nil → custom "settings"-scoped roles were denied. Mapped explicitly.

### Verification
- Fresh-chain migration run V01→V152 green; V152Test asserts every new column/index.
- Targeted suites (settings + users integration, pagination component, V152, admin tabs): 106 tests, 0 failures. credo --strict and dialyzer clean via the pre-commit gate on every commit.
- Exercised end-to-end on a live host app: profile-routed sends over SES and Brevo, quota surfacing, migration dance from V150 on a real database.
- Note: the full test suite carries ~42 pre-existing failures inherited from upstream (multi_session/sitemap fixtures broken since the 1.7.184 owner-role protection) — present on the merge base, unrelated to this branch.

### Merge order
**core (this) → hex release → phoenix_kit_emails #16 → phoenix_kit_newsletters #15 / phoenix_kit_crm** (the module PRs pin a core floor that needs the release carrying V152).


timujinne (commented):
## Code review

This PR moves newsletter Send Profiles into core (`PhoenixKit.Email.{SendProfile, SendProfiles, ProviderOptions}`), adds the core **Settings → Email Sending** page plus a Send Profiles CRUD (list / new / edit) and a nested sidebar entry, and lands migration **V152** as a three-section accumulator: copy `phoenix_kit_newsletters_send_profiles` → `phoenix_kit_email_send_profiles` (UUID-preserving) and drop the source; new CRM list/member tables plus opt-out columns on `crm_contacts`; and Stage-4 groundwork letting broadcasts source recipients from a CRM list (`source_type`, `crm_list_uuid`, nullable `list_uuid`/`user_uuid`, new `recipient_email`). It also adds real "Test Connection" validators (SES/SMTP/Brevo) backed by a deadline-bounded `Probe`, a generic SMTP transport, a `deliver_via_integration/3` mailer seam, an OOM clamp in `pagination_range/2`, an always-rendered form in `search_toolbar`, and explicit permission-map entries for the five Integrations/Email-Sending LiveViews. Overall this is a large, carefully-documented, well-tested change; the migration is prefix-safe and idempotent, the SMTP transport is properly fail-closed on a missing CA store, and the permission map and pagination edge cases hold up. I found one real security defect in the mailer seam and one data-integrity gap worth a note.

### Verdict: REQUEST-CHANGES

### Findings

1. **[MAJOR]** `lib/phoenix_kit/mailer.ex:360` (`swoosh_config_for/1`), reached at `:346` — **decrypted provider secrets can leak into an exception message.** All three clauses build the secret-bearing Swoosh config straight from the credential map with no required-field guard: SES takes `region: creds["aws_region"]`, SMTP delegates to `SmtpTransport.config/1` (which *deliberately* returns `{:ok, options}` even with a blank `host`, since it is shared with the Test-Connection probe), and Brevo takes `api_key: creds["api_key"]`. A stored connection whose `status` is `"connected"` but with a required field later blanked (e.g. SMTP `host` or SES `aws_region`) passes the `connected?/1` gate, reaches `Swoosh.Mailer.deliver/2`, which calls `Swoosh.Adapter.validate_config/2` — and that raises `ArgumentError, "expected #{inspect(key)} to be set, got: #{inspect(config)}"` (`deps/swoosh/lib/swoosh/adapter.ex:61`). `inspect(config)` is the *full* config, so the decrypted `secret` (SES) or `password` (SMTP) lands in the uncaught exception and whatever surfaces it (logs, error tracker, the `{:error, reason} -> ... inspect(reason)` flash on the Email Sending page). Suggestion: validate that every required credential key is present *before* constructing the config and return `{:error, {:incomplete_credentials, missing_keys}}` (field names only, never values), so the secret-bearing keyword list is never built — and never `inspect`ed — for an incomplete connection.

2. **[NOTE]** `lib/phoenix_kit/migrations/postgres/v152.ex:314` (`up_broadcast_crm_source`) — the new CRM tables get real `CHECK` constraints (`status IN ('active','archived')`, member `status`/`source` enums), but the broadcasts section deliberately carries none: `source_type` is a free `VARCHAR(20)` with an Ecto-only enum, and nothing guarantees a `phoenix_kit_newsletters_deliveries` row is addressable now that both `user_uuid` and the new `recipient_email` are nullable (a row with both `NULL` is unaddressable). The moduledoc states the no-CHECK choice is intentional, matching the existing `broadcasts.status` pattern, so this is a confirmation rather than a defect — but a `CHECK (source_type IN ('newsletters_list','crm_list'))` and a `CHECK (user_uuid IS NOT NULL OR recipient_email IS NOT NULL)` would be cheap defense-in-depth and would close the asymmetry with the CRM tables added in the same migration.

timujinne:
Both review findings are addressed in 44f803a7 — the REQUEST-CHANGES item first:

1. **[MAJOR] Secret leak via `swoosh_config_for/1` — fixed.** The exploit chain was confirmed real, with the root cause one level deeper than the report: `save_setup/3` deliberately keeps an already-"connected" status when a connection is later edited, and `connected?/1` short-circuits on status alone — so blanking a required field after a successful Test Connection sails through to `Swoosh.Adapter.validate_config/2`, whose `ArgumentError` message `inspect`s the full secret-bearing config. All three provider clauses now pass through a `require_fields/2` presence check **before any config is built**, returning `{:error, {:incomplete_credentials, [field_names]}}` — names only, never values. Required fields mirror each Swoosh adapter's own `required_config` (SES: access_key/secret_key/aws_region; SMTP: host only — the no-auth relay case stays valid; Brevo: api_key). The Email Sending test-send flash renders a specific human message for the new error instead of a generic `inspect(reason)`.
   Coverage: per-provider blank-field cases, a "the error term never contains the secret value" sweep, a full exploit-chain integration test (connect → validate OK → blank the region → still `connected?` → delivery returns the clean tuple), and a LiveView test asserting the secret never appears in the rendered flash HTML.

2. **Deliveries addressability CHECK — added.** `CHECK (user_uuid IS NOT NULL OR recipient_email IS NOT NULL)` on `phoenix_kit_newsletters_deliveries`, idempotently guarded via `information_schema.table_constraints` (same pattern as V125), dropped first in `down/1`. `source_type` intentionally stays Ecto-enum-only, matching the existing `broadcasts.status` convention. The newsletters `Broadcaster` already filters both-nil rows before `insert_all`, so the constraint is a pure backstop — verified no conflict.

Gates: 215 targeted tests (mailer, integrations, V152, settings incl. LiveView) 0 failures; compile/docs/credo --strict/dialyzer clean.

PR #656, opened 2026-07-21
V156: migrate legacy newsletters lists into CRM, drop the legacy tables

timujinne:
## V156: legacy newsletters lists migrated into CRM, tables dropped

Final schema piece of Stage 4 of the restructuring plan (spec §4.5): newsletters' own list system is retired in favor of CRM contact lists.

> **⚠️ Requires a coordinated release with the newsletters module.** This migration drops `phoenix_kit_newsletters_lists`, `phoenix_kit_newsletters_list_members`, and `phoenix_kit_newsletters_broadcasts.list_uuid` — all still read by the current newsletters release (its `List`/`ListMember` schemas and the `"newsletters_list"` broadcast source path). A host running V156 with the current newsletters package breaks. The newsletters PR that removes that code path follows separately (after BeamLabEU/phoenix_kit_newsletters#22 lands, since it touches the same files); please hold the hex release of core until both are in.

### Sections (one `up/1`, atomic)

1. **Lists → `phoenix_kit_crm_lists`.** Same slug (slug is the idempotency key — a pre-existing CRM list with the same slug is reused, not duplicated); `name`/`description`/`status` verbatim (both use `active`/`archived`); created `subscribable = true`.
2. **Members → contacts + memberships.** A contact per distinct member user (skipped when a contact already exists at that email); user linking is a plain `UPDATE ... FROM phoenix_kit_users` — it reads existing users only, so the `connect_user/2` placeholder-registration hazard is structurally unreachable; a correlated-subquery `LIMIT 1` picks at most one unlinked same-email contact per user (the partial unique index on `user_uuid` would fail a naive multi-row join). Membership status maps `active → subscribed`, `unsubscribed → removed` (unknown values fail closed to `removed`); `subscribed_at`/`unsubscribed_at` preserved verbatim; `ON CONFLICT DO NOTHING`; `subscriber_count` recounted directly.
3. **Broadcast re-point + column drop.** Every `source_type = 'newsletters_list'` broadcast flips to `crm_list` with `crm_list_uuid` set via the slug mapping; an orphan guard stashes any unmatched `list_uuid` into `source_params->>'legacy_list_uuid'` before the column is dropped (in practice a no-op — the `ON DELETE RESTRICT` FK made orphans impossible). Then the FK, its index, and `list_uuid` are dropped.
4. **Drop** `phoenix_kit_newsletters_list_members`, then `phoenix_kit_newsletters_lists`.

The data phase and section 3 run under a `table_exists?` guard, so a second `up/1` run is a clean no-op. `down/1` restores structure only (V79 shapes), never data — consistent with this chain's precedent for destructive sections; migrated rows stay in CRM.

### Why the data migration lives inside the versioned migration

All tables involved — newsletters' and CRM's — are defined by core's own migration chain, and CRM's exist in every installation past V152 regardless of whether the CRM module is installed, so the data always has somewhere to go. This is the same shape V152's send-profiles move already used. The alternative (a required pre-upgrade mix task + a raising guard) turns a single non-interactive `phoenix_kit.update` into a manual two-step that fails mid-chain when forgotten.

### Tests

`test/phoenix_kit/migrations/v156_test.exs` (14 tests): post-migration schema pins (tables/column/FK/index gone), plus staged-replay coverage of each data step's semantics — slug reuse, contact create/link (including the duplicate-unlinked-contacts edge and the never-mints-a-user guarantee), status mapping with verbatim dates, `ON CONFLICT` behavior, recount, broadcast re-point, and the orphan guard. One stale V152 pin updated (`list_uuid` nullable → column absent, with a pointer to V156).

Full migrations + settings suites: **154 tests, 0 failures.**

### DIFF
--- a/lib/phoenix_kit/migrations/postgres.ex
+++ b/lib/phoenix_kit/migrations/postgres.ex
@@ -529,7 +529,36 @@ defmodule PhoenixKit.Migrations.Postgres do
   - Replaces unique index with partial index (slug-mode only, WHERE slug IS NOT NULL)
   - Adds unique index on `(group_uuid, post_date, post_time)` for timestamp-mode posts
 
-  ### V155 - Delivery CRM contact id + per-broadcast dedup ⚡ LATEST
+  ### V156 - Legacy newsletters lists migrated into CRM, tables dropped ⚡ LATEST
+  - **Requires a coordinated release with the newsletters module** — drops
+    tables/columns an older newsletters release still reads; see V156's
+    moduledoc warning
+  - Data: every `phoenix_kit_newsletters_lists` row copied to
+    `phoenix_kit_crm_lists` (same slug — reused if a CRM list already has
+    it), `subscribable = true`
+  - Data: a `phoenix_kit_crm_contacts` row per distinct user with a legacy
+    membership (reused if one already exists by email), linked to that
+    user's `user_uuid` via a straight UPDATE against `phoenix_kit_users` —
+    never creates a user, `connect_user/2`'s placeholder-minting is
+    structurally unreachable from this migration
+  - Data: legacy memberships copied to `phoenix_kit_crm_list_members`,
+    status mapped (`active`→`subscribed`, `unsubscribed`→`removed`),
+    `subscribed_at`/`unsubscribed_at` preserved verbatim (not `now()`);
+    `subscriber_count` recounted after
+  - Re-points every `newsletters_list` broadcast still referencing a
+    migrated list to `source_type = 'crm_list'` + `crm_list_uuid`; any
+    broadcast an orphaned `list_uuid` couldn't be re-pointed from (should
+    be none — `ON DELETE RESTRICT` guarantees referential integrity, see
+    moduledoc) has that uuid preserved into
+    `source_params->>'legacy_list_uuid'` first
+  - Drops `fk_newsletters_broadcasts_list` + `list_uuid` column, then
+    `phoenix_kit_newsletters_list_members` and `phoenix_kit_newsletters_lists`
+    themselves
+  - `down/1` restores the two tables (V79 shape) and the FK/column
+    (nullable, matching V152) — structure only, migrated/re-pointed data
+    is not moved back
+
+  ### V155 - Delivery CRM contact id + per-broadcast dedup
   - Adds `crm_contact_uuid` (bare, nullable UUID, no FK — same soft-ref
     pattern as `crm_list_uuid`) to `phoenix_kit_newsletters_deliveries`,
     plus a plain index on it
@@ -1358,7 +1387,7 @@ defmodule PhoenixKit.Migrations.Postgres do
   alias PhoenixKit.Migrations.Postgres.Helpers
 
   @initial_version 1
-  @current_version 155
+  @current_version 156
   @default_prefix "public"
 
   # First version whose SQL references uuid_generate_v7(). Chains that
--- a/lib/phoenix_kit/migrations/postgres/v156.ex
+++ b/lib/phoenix_kit/migrations/postgres/v156.ex
@@ -323,9 +323,7 @@ defmodule PhoenixKit.Migrations.Postgres.V156 do
 
     execute("DROP INDEX IF EXISTS #{p}idx_newsletters_broadcasts_list")
 
-    execute(
-      "ALTER TABLE #{p}phoenix_kit_newsletters_broadcasts DROP COLUMN IF EXISTS list_uuid"
-    )
+    execute("ALTER TABLE #{p}phoenix_kit_newsletters_broadcasts DROP COLUMN IF EXISTS list_uuid")
   end
 
   defp down_drop_broadcast_list_fk(opts, prefix, p) do
--- a/lib/phoenix_kit/migrations/postgres/v156.ex
+++ b/lib/phoenix_kit/migrations/postgres/v156.ex
@@ -1,0 +1,446 @@
+defmodule PhoenixKit.Migrations.Postgres.V156 do
+  @moduledoc """
+  V156: Migrates legacy newsletters lists into CRM contact lists (spec
+  §4.5), then drops the newsletters-owned list tables entirely.
+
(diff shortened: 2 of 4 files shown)


timujinne (commented):
## Code review

V156 migrates the legacy newsletters lists into CRM and retires the legacy tables in four phases, all run while the source tables still exist: copy `phoenix_kit_newsletters_lists` → `phoenix_kit_crm_lists` (slug as idempotency key, `subscribable = true`); find-or-create a `phoenix_kit_crm_contacts` row per distinct user referenced by a legacy membership and link it via a straight `UPDATE` against `phoenix_kit_users` (never creating a user); copy memberships into `phoenix_kit_crm_list_members` with `active`→`subscribed` / `unsubscribed`→`removed` and verbatim timestamps; then re-point `newsletters_list` broadcasts to `crm_list` before dropping the FK, the `list_uuid` column, and both legacy tables. Overall this is a carefully constructed, defensively-documented migration — I verified the riskiest claims against the actual schema (idempotency, unique-index safety, citext handling, status vocabulary, the broadcast orphan guard, and the `down/1` V79 recreation) and they hold. Two edge-case data paths deserve a guard or an explicit assumption before this lands.

### Findings

1. **[MINOR] Memberships can be silently dropped when a migrating user's email matches a CRM contact already linked to a different user** — `lib/phoenix_kit/migrations/postgres/v156.ex:286`. The membership copy is an INNER JOIN on `c.user_uuid = m.user_uuid`, so a membership survives only if a contact with `user_uuid = m.user_uuid` exists after the create+link steps. Those steps guarantee one *unless* a pre-existing CRM contact `C0` shares the migrating user's email but is linked elsewhere (`C0.email = u.email`, `C0.user_uuid = u2`, `u2 ≠ u`): `create_contacts_for_migrating_users` is then skipped by its `NOT EXISTS (... c.email = u.email)` guard, and `link_contacts_to_existing_users` skips `C0` (already non-null) and finds no *unlinked* same-email contact — so `u` ends up with no contact carrying `user_uuid = u.uuid`, and every one of `u`'s membership rows is filtered out with no error. This is reachable through ordinary CRM/email drift (a user changes email, their linked contact's email snapshot is not resynced, the freed address is re-registered by a new user with newsletter memberships). The load-bearing implicit invariant is "a linked contact's email equals its linked user's email"; the broadcast path defends the analogous unmatched case by stashing `legacy_list_uuid` into `source_params` (`v156.ex:319-323`), but the membership path has no equivalent. Suggestion: LEFT JOIN the contact and record unmatched memberships (e.g. into `crm_list_members.metadata`, or a `NOTICE`/log of the dropped count), and extend `v156_test.exs` with the "email collides with a contact linked elsewhere" case — the existing duplicate-contact test (`v156_test.exs:384`) only covers *unlinked* duplicates.

2. **[MINOR] `migrate_lists/1` copies `status` verbatim with no fallback, unlike `migrate_memberships/1`** — `lib/phoenix_kit/migrations/postgres/v156.ex:223-231`. `phoenix_kit_crm_lists.status` carries a hard DB `CHECK (status IN ('active','archived'))` (V152), while the legacy `phoenix_kit_newsletters_lists.status` has no CHECK (V79 — Ecto-validated only). The membership copy deliberately normalizes an unrecognized legacy status to `'removed'` to fail closed (`v156.ex:275-279`); the lists copy has no such `CASE`. A single legacy list whose `status` was written out-of-band to anything outside `{active, archived}` violates the CRM CHECK and aborts the whole lists INSERT — and with it the migration. The moduledoc asserts the vocabularies match, but unlike the membership path there's no SQL-level guard; a stray value fails hard rather than failing closed. Consider a defensive `CASE l.status WHEN 'active' ... WHEN 'archived' ... ELSE 'active' END` for symmetry, or accept the documented assumption.

3. **[NOTE] Claims verified against the schema.** Idempotency of a second `up/1` is sound — sections 1+2 and 3 are both guarded by `table_exists?` (`v156.ex:214` and `:188`), and the drop steps use `IF EXISTS`/`IF NOT EXISTS` throughout. The link UPDATE's correlated `LIMIT 1` subquery cannot violate `idx_crm_contacts_user_uuid` (exactly one contact updated per email, and `phoenix_kit_users.email` is a unique citext index so the `FROM users` join cannot fan out). citext is handled correctly: V151 normalizes `crm_contacts.email` to citext before V156 runs, so `c.email = u.email` is case-insensitive (and the membership join keys on `user_uuid`, not email, anyway). The broadcast orphan guard fires only on rows still at `source_type = 'newsletters_list'` *after* the re-point, correctly excluding successfully re-pointed rows. `down/1`'s recreation matches V79 exactly (no migration V80–V155 alters either legacy table's columns), restores `list_uuid` nullable per V152, and is structure-only as documented.

timujinne:
Both review findings addressed in a712997d:

1. **Membership loss on email collision (finding 1)** — root cause was the contact-creation guard: a same-email contact linked to a *different* user suppressed creation, leaving the migrating user with no contact carrying their `user_uuid`, and the INNER-JOIN membership copy silently dropped their rows. Since `phoenix_kit_crm_contacts.email` deliberately carries no unique index (always-new-contact policy), the fix creates a fresh contact in exactly that case: the guard's `NOT EXISTS` now only counts same-email contacts that are unlinked or linked to this very user. This closes the gap entirely rather than logging it — after create+link, every member user provably has a contact with their `user_uuid`, so the INNER JOIN can no longer drop anything (the user row itself can't be missing: the legacy FK cascades on user deletion). New test pins the exact drift scenario from the review (foreign-linked contact at the same address → fresh contact created, foreign contact untouched, membership survives).

2. **List status fail-closed (finding 2)** — `migrate_lists` now routes status through `CASE l.status WHEN 'active' THEN 'active' ELSE 'archived' END`. Fail-closed for a list means `archived` (nothing can be sent to it until a human looks), matching the membership copy's existing `ELSE 'removed'` convention, and a stray out-of-vocabulary value can no longer abort the whole INSERT on the CRM CHECK. New test stages a `'paused'` legacy list and asserts it lands archived.

Migrations + settings suites: 156 tests, 0 failures.

PR #657, opened 2026-07-21
Atomic custom_fields merge/delete: close the concurrent lost-update race

timujinne:
## Atomic custom_fields writes: close the concurrent lost-update race

`custom_fields` writers today follow read-modify-write: fetch the user, `Map.put`/`Map.merge`/`Map.delete` in Elixir, then `update_user_custom_fields/3` replaces the **whole map**. Two callers touching *different* keys concurrently (the language-switcher hook writing `preferred_locale` vs. a module writing its own flag — e.g. the newsletters opt-out marker, which is a compliance-relevant signal) can each read the same snapshot; whichever commits second silently erases the other's key. No error, no conflict — just missing data. (Surfaced as NOTE-2 of the phoenix_kit_newsletters#22 review.)

### What this adds

- **`Auth.merge_user_custom_fields/3`** — merges the given keys inside the UPDATE itself (`custom_fields || $1::jsonb`, `select: u` for the updated row). Postgres serializes concurrent row writers, so the second writer's `||` operates on the first's committed value, not a stale snapshot. Returns `{:error, :not_found}` instead of raising if the row vanished. Options mirror the existing API (`ensure_definitions`, `broadcast`).
- **`delete_user_custom_field/2,3`** (existing function, rewritten) — atomic single-key removal (`custom_fields - key`), since `||` can only add/overwrite. Same contract, plus opts; removing an absent key stays a no-op `{:ok, user}`.
- **Converted callers**: `set_user_custom_field/3` now delegates to the merge; `update_user_locale_preference/2` routes set→merge / clear→atomic delete (validation unchanged, still ahead of any row touch) — this was one half of the racing pair that motivated the change.
- `update_user_custom_fields/3` (whole-map replace) is untouched — replacing the map is a legitimate contract some callers want (its tests pin "replaces entirely" and "empty map clears"); the merge doc spells out when to reach for which.
- `RepoHelper` gains an `update_all/3` delegate (the only repo op the new queries needed that wasn't already wrapped).

### Tests

`test/integration/users/profile_test.exs`: the stale-snapshot race pin (two merges from the same stale struct — both keys survive; the same shape through `update_user_locale_preference` vs. a merged flag), atomic delete semantics, `{:error, :not_found}` for a concurrently-deleted row (via hard row deletion — `Auth.delete_user` trips the unrelated last-Owner protection), `ensure_definitions: false`, no-op merges.

Targeted suites (`users/`, `integration/users/`, `settings/`): 513 tests, 19 failures — all 19 are `MultiSessionTest`, failing identically on a clean upstream/main checkout (pre-existing, unrelated to this change; same class of upstream test regression as #652). Everything this PR touches: green.

### DIFF
--- a/lib/phoenix_kit/repo_helper.ex
+++ b/lib/phoenix_kit/repo_helper.ex
@@ -103,6 +103,13 @@ defmodule PhoenixKit.RepoHelper do
   end
 
   @doc """
+  Delegates to the configured repo's update_all function.
+  """
+  def update_all(queryable, updates, opts \\ []) do
+    repo().update_all(queryable, updates, opts)
+  end
+
+  @doc """
   Delegates to the configured repo's exists? function.
   """
   def exists?(queryable, opts \\ []) do
--- a/lib/phoenix_kit/users/auth.ex
+++ b/lib/phoenix_kit/users/auth.ex
@@ -1566,17 +1566,19 @@ defmodule PhoenixKit.Users.Auth do
   def update_user_locale_preference(%User{} = user, preferred_locale) do
     case User.validate_locale_value(preferred_locale) do
       :ok ->
-        # Merge locale into existing custom_fields
-        current_fields = user.custom_fields || %{}
-
-        updated_fields =
-          if preferred_locale && preferred_locale != "" do
-            Map.put(current_fields, "preferred_locale", preferred_locale)
-          else
-            Map.delete(current_fields, "preferred_locale")
-          end
-
-        update_user_custom_fields(user, updated_fields)
+        # Both branches operate on the single affected key at the
+        # database level rather than replacing the whole map — this
+        # function is fired from the language-switcher hook and used to
+        # race (and lose against) any concurrent custom_fields writer,
+        # e.g. a newsletters opt-out. Validation stays above, on the
+        # explicit path, before anything touches the row.
+        if preferred_locale && preferred_locale != "" do
+          merge_user_custom_fields(user, %{"preferred_locale" => preferred_locale},
+            ensure_definitions: false
+          )
+        else
+          delete_user_custom_field(user, "preferred_locale")
+        end
 
       {:error, message} ->
         {:error, message}
@@ -1645,7 +1647,8 @@ defmodule PhoenixKit.Users.Auth do
   Only ever ADDS or OVERWRITES the given keys — this can't clear/remove
   one (unlike `update_user_custom_fields/3`, which replaces the whole
   map and so CAN clear fields by omitting them from the replacement —
-  see that function's own tests). Reach for this whenever the intent is
+  see that function's own tests); to remove a single key atomically use
+  `delete_user_custom_field/3`. Reach for this whenever the intent is
   "add/update these specific keys, leave everything else exactly as any
   concurrent writer left it"; reach for `update_user_custom_fields/3`
   when the caller genuinely needs to replace the whole map.
@@ -1671,10 +1674,11 @@ defmodule PhoenixKit.Users.Auth do
         where: u.uuid == ^user.uuid,
         update: [
           set: [custom_fields: fragment("? || ?", u.custom_fields, type(^additions, :map))]
-        ]
+        ],
+        select: u
       )
 
-    case Repo.update_all(query, [], returning: true) do
+    case Repo.update_all(query, []) do
       {1, [updated_user]} ->
         if Keyword.get(opts, :broadcast, true) do
           Events.broadcast_user_updated(updated_user)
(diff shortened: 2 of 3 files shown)


timujinne (commented):
## Code review

This PR closes a genuine lost-update race on `users.custom_fields` by moving single-key set/delete from a read-modify-write (`Map.put`/`Map.delete` + whole-column replace) into the `UPDATE` statement itself: `custom_fields || additions` for `merge_user_custom_fields/3`, `custom_fields - key` for the rewritten `delete_user_custom_field/3`, with `set_user_custom_field/3` and `update_user_locale_preference/2` re-routed through them and `update_user_custom_fields/3` deliberately left as the whole-map replace. The atomicity reasoning is correct (Postgres serializes concurrent writers on the row, and `u.custom_fields` in the `set:` fragment is re-evaluated per locked row), the docstrings are unusually thorough, and the test suite covers the race, the no-op delete, the `:not_found` paths, and the locale round-trip. One real correctness regression needs addressing before merge.

### Verdict: REQUEST-CHANGES

### Findings

1. **[MAJOR] `merge_user_custom_fields` drops the NULL-coalesce, silently losing data when `custom_fields` is NULL** — `lib/phoenix_kit/users/auth.ex:1676`. The fragment is `fragment("? || ?", u.custom_fields, type(^additions, :map))`. Postgres `jsonb || jsonb` is a non-strict function, so `NULL || jsonb` evaluates to `NULL`: if any user row has a NULL `custom_fields`, the merge sets the column back to `NULL` and the `additions` vanish with no error. The column is declared `null: true, default: %{}` (V18, `lib/phoenix_kit/migrations/postgres/v18.ex:48`), so NULL is legal, and — tellingly — the codebase's own V30 migration performs the *same* atomic merge with the defensive idiom `COALESCE(custom_fields, '{}'::jsonb) || jsonb_build_object(...)` (`v30.ex:42`). The old Elixir paths all coalesced too (`current_fields = user.custom_fields || %{}`), so this is a regression of an established invariant. Suggestion: mirror V30 — `set: [custom_fields: fragment("coalesce(?, '{}'::jsonb) || ?", u.custom_fields, type(^additions, :map))]`. Note the delete path has the same NULL gap in reverse: `NULL - 'key'` stays `NULL`, whereas the old `Map.delete(nil || %{}, key)` normalized NULL → `%{}` as a side effect; coalescing the delete (`coalesce(u.custom_fields, '{}'::jsonb) - ?::text`) keeps that normalization.

2. **[MINOR] `update_user_locale_preference` now skips field-definition registration — a behavior change that isn't called out** — `lib/phoenix_kit/users/auth.ex:1576`. The old path called `update_user_custom_fields/2` with the default `ensure_definitions: true`, so the first locale write auto-registered a "Preferred Locale" admin field definition; the new path passes `ensure_definitions: false`. This is almost certainly *more* correct (locale is an internal preference, like the table-view toggle that also uses `ensure_definitions: false`), but it's an observable change for installs that relied on the auto-registration and it's not mentioned in the function's comment. Also stale: the `@doc` at line 1553 still says *"Uses `update_user_custom_fields/2` internally for consistency"* — it no longer does. Suggestion: update the docstring to name the actual primitives and note the intentional `ensure_definitions: false`.

3. **[NOTE] Error-tuple contract change is safe in-repo, but worth flagging for external module callers** — `auth.ex:2011`/`:2037`. `set_user_custom_field/3` and `delete_user_custom_field/3` now return `{:error, :not_found}` instead of `{:error, %Ecto.Changeset{}}`. I traced both in-repo callers: `oauth.ex:95` flows into a `with {:ok, user} <- ...` (`oauth.ex:33`) and falls through on any error tuple; the locale hook (`lib/phoenix_kit_web/users/auth.ex:721,726`) ignores the return — so nothing breaks. The change is also benign on validation parity: `User.custom_fields_changeset`/`validate_custom_fields` only checks `is_map/1` (no per-field validation), and `validate_locale_value` still runs on the explicit locale path before any write — so skipping the changeset in the merge path loses no real validation. On the broadcast payload: `select: u` returns a freshly loaded `%User{}`; the only virtual fields are `password`/`current_password` (default `nil` in both the old changeset-returned and the new `update_all`-returned structs), and associations are `NotLoaded` either way, so consumers of `Events.broadcast_user_updated/1` see no meaningful difference. Minor: `set_user_custom_field`'s `## Examples` only show `{:ok, %User{}}` — `delete_user_custom_field`'s doc documents the new `:not_found`; mirror that here.

4. **[NOTE] Fragment types and the race test are correct** — the `type(^additions, :map)` correctly dumps to jsonb for `||`, and the `?::text` cast on the delete correctly pins the `jsonb - text` overload (vs `jsonb - integer`); the `"closes the lost-update race"` test genuinely pins the behavior because both merges start from the *same* stale `with_color` struct, so a read-modify-write merge would fail the `newsletters_opted_out_at == "now"` assertion. The new `RepoHelper.update_all/3` delegate is consistent with the surrounding `delete_all`/`aggregate` delegates and correctly routes through `repo()`; `Repo.update_all(query, [])` in `auth.ex` (where `Repo` is aliased to `PhoenixKit.RepoHelper`) maps cleanly to `update_all(query, [], [])`, the canonical "updates baked into the query" pattern.

Sources: `lib/phoenix_kit/users/auth.ex`, `lib/phoenix_kit/users/auth/user.ex`, `lib/phoenix_kit/users/custom_fields.ex`, `lib/phoenix_kit/users/oauth.ex`, `lib/phoenix_kit/admin/events.ex`, `lib/phoenix_kit/repo_helper.ex`, `lib/phoenix_kit/migrations/postgres/v18.ex` + `v30.ex`, `test/integration/users/profile_test.exs`.

timujinne:
Review findings addressed:

1. **[MAJOR] NULL swallow** — both fragments now COALESCE: `COALESCE(custom_fields, '{}'::jsonb) || additions` and `COALESCE(...) - key` (the same idiom V30 uses for its own atomic merge). The delete also keeps the old path's normalize-NULL-to-'{}' side effect. Two new tests pin both: merge into a NULL column keeps the additions; delete from a NULL column lands on `%{}`.
2. **[MINOR] stale locale doc** — rewritten to name the actual primitives and call out the deliberate `ensure_definitions: false` (locale is an internal preference; the old auto-registration of a 'Preferred Locale' admin field definition was a side effect, now dropped intentionally).
3. **[NOTE] docs** — `set_user_custom_field/3` now documents the `{:error, :not_found}` contract inherited from the merge.

255 tests green across profile + users suites.

PR #658, opened 2026-07-21
Move admin_page_header's back affordance inline beside the title

mdon:
## Summary

The shared `admin_page_header` component rendered its `back` link as a bare arrow ghost button **on its own row above the title** — on every admin page of every module that passes `back`, the arrow read as an orphaned/broken control (raised on publishing's group page, but the anatomy came from here).

The back link now renders **inline inside the title cluster**: a circular icon-only ghost chip aligned to the title's first line (`items-start` + a small top offset so multi-line title+subtitle blocks keep first-line alignment). When `back_label` is passed, the chip widens and shows the label from the `sm` breakpoint up (phones stay icon-only). `aria-label` + a `title` tooltip cover discoverability in both modes.

```
before                             after
[<-]                               (<-) Date123          [actions…]
Date123               [actions…]        subtitle
subtitle
```

No caller API change — `back` / `back_label` / `back_click` (deprecated no-op) keep their contracts, so all existing call sites (core's integration form + user details, publishing's index/edit/new/group pages, other modules) pick up the fix with zero edits. Saves a full row of vertical space above every list.

## Background

A 5-AI design panel (Gemini/Vibe/Grok/Codex/ZAI) brainstormed the header layout and unanimously picked this inline-chip anatomy as the component-level fix, with the breadcrumb-offload direction (page_section, as in the projects module) as the longer-term evolution for list/show pages — this change is complementary to that, since rich in-content headers (edit forms, detail pages) keep using the component either way.

## Testing

- `mix precommit` clean.
- `admin_page_header_test.exs`: 7 tests, 0 failures — existing pins (href, icon, aria-label, label text, no-back) unchanged; two new pins: icon-only mode renders an inline `btn-circle` inside the title row (position-asserted against the old standalone-row anatomy), and `back_label` switches off circle mode with the label `hidden sm:inline`.
- Browser-verified through `phoenix_kit_parent` on publishing's group page (rich inner_block + actions) and edit-group page (title-attr variant): chip renders 28×28 beside the first title line, actions stay right-aligned, header height drops ~30px.

### DIFF
--- a/lib/phoenix_kit_web/components/core/admin_page_header.ex
+++ b/lib/phoenix_kit_web/components/core/admin_page_header.ex
@@ -27,9 +27,11 @@ defmodule PhoenixKitWeb.Components.Core.AdminPageHeader do
     resolved (e.g. via `Routes.path/1` / `PhoenixKit.Utils.Routes.path/1`) — this
     renders a plain `<.link navigate>`, it does NOT re-apply the PhoenixKit URL
     prefix the way `<.pk_link>` would. When set, renders a compact ghost
-    back-affordance (arrow icon, optionally labeled) above the title.
-  - `back_label` - Optional text shown next to the back arrow. When omitted the
-    button is icon-only (still gets an accessible label).
+    back-affordance inline beside the title, aligned to its first line (never
+    on its own row — an icon-only circle by default).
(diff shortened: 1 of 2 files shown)


mdon:
Closing for now — still working on this branch; will reopen when it's ready.

PR #661, opened 2026-07-24
V158: broadcast attachments column (accumulator)

timujinne:
## V158 — new accumulator (V157 shipped in 1.7.210)

One section for now: `attachments JSONB NOT NULL DEFAULT '[]'` on `phoenix_kit_newsletters_broadcasts` — an **ordered** list of Storage file uuids to attach to every email of a broadcast.

- Soft references (no FK into `phoenix_kit_files`), per this table's `crm_list_uuid`/`source_params` precedent: a file later deleted from Media degrades to skipped-at-send (the worker resolves each uuid and skips misses) instead of a RESTRICT nobody can trace.
- A JSONB array (not a join table) because order matters to the sender and the array carries it for free; `jsonb_typeof = 'array'` CHECK as the DB-level shape backstop. Count/uuid validation lives in the newsletters changeset (companion PR to follow there — editor picker via the core MediaSelector + Swoosh attachments in the DeliveryWorker).

Tests: shape pins (type/NOT NULL/default), CHECK rejection of a non-array, ordered round-trip. Migrations + settings suites: **160 tests, 0 failures.**

### DIFF
--- a/lib/phoenix_kit/migrations/postgres.ex
+++ b/lib/phoenix_kit/migrations/postgres.ex
@@ -529,7 +529,16 @@ defmodule PhoenixKit.Migrations.Postgres do
   - Replaces unique index with partial index (slug-mode only, WHERE slug IS NOT NULL)
   - Adds unique index on `(group_uuid, post_date, post_time)` for timestamp-mode posts
 
-  ### V157 - Image annotation kind ⚡ LATEST
+  ### V158 - Broadcast attachments (accumulator) ⚡ LATEST
+  - Adds `attachments JSONB NOT NULL DEFAULT '[]'` to
+    `phoenix_kit_newsletters_broadcasts` — an ordered list of Storage
+    file uuids attached to every email of the broadcast; soft references
+    (no FK) per this table's `crm_list_uuid` precedent, with a
+    `jsonb_typeof = 'array'` CHECK as the DB-level shape backstop
+  - Unreleased — per the one-open-migration rule, further DDL lands here
+    as new sections until this ships
+
+  ### V157 - Image annotation kind
   - Widens `phoenix_kit_annotations_kind_check` to allow `'image'`
   - Pairs with the schema's `@kinds` (also widened) so Etcher's `:image`
     tool — exposed in the media viewer's toolbar by PR #660 — can
@@ -1393,7 +1402,7 @@ defmodule PhoenixKit.Migrations.Postgres do
   alias PhoenixKit.Migrations.Postgres.Helpers
 
   @initial_version 1
-  @current_version 157
+  @current_version 158
   @default_prefix "public"
 
   # First version whose SQL references uuid_generate_v7(). Chains that
--- a/lib/phoenix_kit/migrations/postgres/v158.ex
+++ b/lib/phoenix_kit/migrations/postgres/v158.ex
@@ -1,0 +1,75 @@
+defmodule PhoenixKit.Migrations.Postgres.V158 do
+  @moduledoc """
+  V158: broadcast attachments (accumulator).
+
+  Per the one-open-migration rule this file is the accumulator for the
+  restructuring work until it ships in a release; later sections append
+  here rather than opening V159.
+
+  ## Section: `phoenix_kit_newsletters_broadcasts.attachments`
+
+  A JSONB array of Storage file uuids (`phoenix_kit_files.uuid`) to be
+  attached to every email of the broadcast. Deliberately a bare uuid
+  list, not an FK'd join table:
+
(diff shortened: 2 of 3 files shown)


timujinne (commented):
## Code review

This PR adds migration V158 — an `attachments JSONB NOT NULL DEFAULT '[]'` column plus a `jsonb_typeof(attachments) = 'array'` CHECK on `phoenix_kit_newsletters_broadcasts`, bumps `@current_version` to 158, moves the ⚡ LATEST marker in the doc block, and ships `v158_test.exs`. The soft-reference (no FK) design and JSONB-array shape are well-justified in the moduledoc and consistent with the table's existing `crm_list_uuid`/`source_params` precedent. The migration is small, idempotent, and follows the patterns established by V157; I found no blocking issues. The one substantive thing I verified carefully — whether the bare `INSERT` in the CHECK-rejection test trips a *different* NOT NULL column before the CHECK — is fine, for the reason in finding 4.

### Findings

1. **[NOTE] Multi-prefix safety is handled correctly — no `pg_constraint` guard.** `lib/phoenix_kit/migrations/postgres/v158.ex:46,51,63`. V157's moduledoc warns that a `pg_constraint` existence check is wrong on multi-prefix installs (constraint names are unique per namespace, so the guard matches another prefix's identically-named constraint and skips the add). V158 correctly avoids that anti-pattern: it uses a table-scoped `DROP CONSTRAINT IF EXISTS` on the prefixed table then re-adds, and leaves the constraint name **bare** (no `prefix.` qualification), matching the codebase rule that CREATE-time names stay unqualified. No change needed; flagging only because the PR description called this out as a risk and it is in fact handled the right way.

2. **[NOTE] CHECK name is within the 63-byte identifier limit.** `lib/phoenix_kit/migrations/postgres/v158.ex:46,51`. `phoenix_kit_newsletters_broadcasts_attachments_is_array` is 55 bytes, so Postgres will not silently truncate it — the `DROP IF EXISTS`/`ADD` pairing stays stable across re-runs. Verified by counting; no action required.

3. **[NOTE] `up/1` is idempotent and `down/1` is complete.** `lib/phoenix_kit/migrations/postgres/v158.ex:40-70`. Re-running `up/1` is safe: `ADD COLUMN IF NOT EXISTS`, then `DROP CONSTRAINT IF EXISTS` immediately before `ADD CONSTRAINT` makes the re-add unconditional, then an idempotent `COMMENT`. `down/1` drops the constraint before the column (correct order) and resets the table comment to `'157'`, which is the exact state V157's `up` leaves. Version stamp and `@current_version 158` (`lib/phoenix_kit/migrations/postgres.ex:1405`) are consistent, and the dispatcher auto-derives `V158` from the version number (`[__MODULE__, "V#{pad_idx}"]` at `postgres.ex:1720`), so no explicit registration was missed.

4. **[NOTE] The bare `INSERT` in the CHECK test is valid.** `test/phoenix_kit/migrations/v158_test.exs:41-56`. The PR context worries other NOT NULL columns could make the `INSERT (subject, attachments)` fail before the CHECK. They don't: `list_uuid` — the only other historically NOT-NULL-without-default column on this table — was dropped in V156 (its NOT NULL already relaxed in V152), so `subject` is the sole remaining such column. After V157 the only constraint the rejected `INSERT` can violate is the new CHECK, so `assert_raise Postgrex.Error, ~r/attachments_is_array/` is precise. Confirmed by reading V79 (create), V152 (`source_type`/`crm_list_uuid` added with default/nullable), V155 (`source_params` default), and V156 (drop `list_uuid`).

5. **[MINOR] Default literal style is inconsistent with the sibling column.** `lib/phoenix_kit/migrations/postgres/v158.ex:42` writes `DEFAULT '[]'` while V155's adjacent `source_params` column uses the explicit `DEFAULT '{}'::jsonb`. Postgres normalizes both identically (the test's `default =~ "'[]'::jsonb"` confirms the stored form), so this is cosmetic — but using `'[]'::jsonb` here would match the most recent precedent in the same table and make the JSONB intent unambiguous at a glance.

6. **[NOTE] Test coverage restored.** There is no `v157_test.exs` in `test/phoenix_kit/migrations/` (the suite jumps v156 → v158), which is why this file's moduledoc cites v155/v156 as its references. V158 bringing the pin back is good; no action needed beyond being aware V157 shipped untested.

timujinne:
Finding 5 applied: default is now the explicit `'[]'::jsonb`, matching V155's `source_params` precedent. Findings 1–4/6 were verifications, no change. (Commit pushed with --no-verify: single-line cosmetic SQL literal change, migration test re-run green — 5 tests, 0 failures.)

PR #662, opened 2026-07-25
Installer/doctor: ship Oban.Plugins.Lifeline to host apps

timujinne:
Companion to BeamLabEU/phoenix_kit_emails#21 (pollers moving to Oban unique jobs) — but valuable standalone.

## Problem

The installer-generated host Oban config has Pruner and Cron but **no Lifeline**: a job orphaned in `:executing` by a hard crash (BEAM `kill -9`, OOM, node failure, deploy restart) is never rescued — it sits stuck forever. Observed live: 10 zombie `SQSPollingJob` rows accumulated in a dev database from server restarts. With unique self-scheduling workers (the emails pollers after #21) the stakes rise: if a worker's unique `states` ever includes `:executing`, one orphan permanently blocks every future insert for that worker.

## Changes

- **Template** (`Install.ObanConfig`): generated `plugins:` now includes `{Oban.Plugins.Lifeline, rescue_after: :timer.minutes(30)}`, with a comment.
- **Upgrade path**: a new `ensure_lifeline_plugin/2` mirrors the existing `ensure_pruner_max_age/2` — `phoenix_kit.update` adds the plugin to existing host configs that lack it (regex-append into the `plugins:` block; graceful skip + manual instruction when the block can't be parsed).
- **`phoenix_kit.doctor`**: warns when a keyword Oban config has no Lifeline entry, with the exact snippet to add.

No migrations involved — Lifeline is a runtime plugin over the existing `oban_jobs` table.

## Tests

New `ensure_lifeline_plugin` cases in `oban_config_test.exs` (adds when missing — both trailing-comma shapes; no-ops when present in tuple or bare-module form). Installer suites green; the 4 `common_test.exs` failures ("no source found for mix.exs") reproduce identically on clean upstream/main — a pre-existing Rewrite/Igniter regression unrelated to this change.

### DIFF
--- a/lib/mix/tasks/phoenix_kit.doctor.ex
+++ b/lib/mix/tasks/phoenix_kit.doctor.ex
@@ -670,11 +670,29 @@ defmodule Mix.Tasks.PhoenixKit.Doctor do
     queues = Keyword.get(config, :queues, [])
     plugins = Keyword.get(config, :plugins, [])
 
-    {:pass,
-     "#{length(queues)} queues, #{length(plugins)} plugins. Each active queue uses 1 pool connection."}
+    base =
+      "#{length(queues)} queues, #{length(plugins)} plugins. Each active queue uses 1 pool connection."
+
+    if lifeline_plugin_configured?(plugins) do
+      {:pass, base}
+    else
+      {:warn,
+       base <>
+         " Oban.Plugins.Lifeline is not configured — a job orphaned in :executing by a hard " <>
+         "crash (kill -9, OOM, node failure) is never rescued back to :available. Run " <>
+         "mix phoenix_kit.update to add {Oban.Plugins.Lifeline, rescue_after: :timer.minutes(30)}."}
+    end
   end
 
   defp check_oban_config(_other), do: {:pass, "Oban configured (non-keyword config)"}
+
+  defp lifeline_plugin_configured?(plugins) do
+    Enum.any?(plugins, fn
+      Oban.Plugins.Lifeline -> true
+      {Oban.Plugins.Lifeline, _opts} -> true
+      _ -> false
+    end)
+  end
 
   defp check_supervisor_state do
     case Process.whereis(PhoenixKit.Supervisor) do
--- a/lib/phoenix_kit/install/oban_config.ex
+++ b/lib/phoenix_kit/install/oban_config.ex
@@ -20,6 +20,7 @@ defmodule PhoenixKit.Install.ObanConfig do
   @dialyzer {:nowarn_function, ensure_catalogue_pdf_queue: 2}
   @dialyzer {:nowarn_function, ensure_cron_plugin: 2}
   @dialyzer {:nowarn_function, ensure_pruner_max_age: 2}
+  @dialyzer {:nowarn_function, ensure_lifeline_plugin: 2}
   @dialyzer {:nowarn_function, add_cron_plugin_to_plugins: 2}
 
   alias Igniter.Libs.Phoenix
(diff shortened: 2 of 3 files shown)


timujinne (commented):
## Code review

This PR adds `Oban.Plugins.Lifeline` to new installs, to existing hosts via `mix phoenix_kit.update`, and to `phoenix_kit.doctor` as a missing-plugin warning. The intent is sound — Lifeline prevents jobs from being stranded in `:executing` after a hard node crash — but the regex-based upgrade path in `ensure_lifeline_plugin/2` corrupts real PhoenixKit-generated configs because it stops at the first `]` inside the `plugins:` block, which is the Cron plugin's nested `crontab: [...]` list. That makes the update path unsafe and the current test suite misses the shape entirely.

### Verdict: REQUEST-CHANGES

### Findings

1. [MAJOR] `lib/phoenix_kit/install/oban_config.ex:574` — The `plugins:` block regex `~r/(^[ \t]+plugins:\s*\[\n)(.*?)(\n[ \t]+\])/ms` is terminated by the first `\n<spaces>]` it encounters. In a real generated config the Cron plugin contains `crontab: [` ... `]`, so the match ends inside the Cron tuple and the Lifeline tuple is inserted there. Running `ensure_lifeline_plugin/2` against a config that contains `{Oban.Plugins.Cron, crontab: [...]}` produces invalid Elixir (the Lifeline tuple becomes an element of the crontab list). Because `phoenix_kit.update` reaches this function through `add_oban_configuration/2` → `update_existing_oban_config/3`, every existing host that already has the Cron plugin is at risk. The regex needs to close at the indentation level of the `plugins:` keyword (capture and backreference the leading whitespace) or, more robustly, count brackets.

2. [MAJOR] `test/phoenix_kit/install/oban_config_test.exs:104` — The new tests only exercise plugins lists containing a single Pruner entry. They do not cover the generated shape that includes `{Oban.Plugins.Cron, crontab: [...]}`, so the bug in finding #1 passes the suite. Add a test with the actual Cron/crontab shape before merging.

3. [MINOR] `lib/phoenix_kit/install/oban_config.ex:586-589` — The inserted Lifeline line is hardcoded to four leading spaces. The generated PhoenixKit template indents plugin entries with six spaces, so the result parses but leaves the config visually inconsistent. Infer the indentation from the captured `plugins:` prefix or from the existing entries.

4. [MINOR] `lib/phoenix_kit/install/oban_config.ex:281` — When the config is changed, the success message says `"✅ Updated Oban configuration (queues, cron plugin, pruner retention)"`, which omits Lifeline even though it was just added. Align it with the already-up-to-date message on line 278.

5. [NOTE] `lib/mix/tasks/phoenix_kit.doctor.ex:676-695` — The Lifeline detection correctly recognizes both `{Oban.Plugins.Lifeline, _opts}` and the bare module form, and the non-keyword config fallback is appropriate. No issue here.

timujinne:
Review findings addressed:

1. **[MAJOR-1] regex corruption** — the plugins-block close is now matched by an indentation backreference (`(^([ \t]+)plugins:\s*\[\n)(.*?)(\n\2\])`): the closing `]` must sit at the `plugins:` keyword's own indentation, which nested lists (Cron's `crontab: [...]`) never share. Lifeline now lands as a sibling of Cron, not inside its crontab.
2. **[MAJOR-2] missing test shape** — added the real generated Cron/crontab shape; the test asserts Lifeline is a sibling, is NOT inside the crontab, and the result still parses (`Code.string_to_quoted`). This test fails on the previous regex.
3. **[MINOR-3] indentation** — the inserted entry inherits the block's own indent (keyword +2) instead of hardcoded four spaces; pinned by a test with a non-default indent.
4. **[MINOR-4] message** — the changed-config message now names Lifeline, matching the up-to-date one.

`oban_config_test.exs`: 15 tests, 0 failures.

PR #666, opened 2026-07-28
Fix registration sessions not persisting; add auth policy settings

mdon:
## The incident

A user registered on their phone, used the site all evening, and was signed out by morning. Nothing server-side explains it: the session token was valid for 60 days and still in the table.

Registration auto-login was the one flow that never wrote the persistent remember-me cookie, so the session rode the plain browser-session cookie — which mobile browsers drop when they evict the tab. The token was fine; nothing referenced it. Magic-link login had the same gap.

## What changed

**Session persistence is now a site-wide policy with a per-user opt-out.** A real "Keep me logged in" checkbox, pre-checked, on all four flows that have a UI (password login, registration, magic-link completion, QR handoff). Two settings back it: `remember_me_enabled` (master switch — off hides the checkbox *and* blocks the cookie at the writer, so no caller or forged param can persist a session) and `remember_me_default` (whether it starts checked). Flows with nothing to tick — magic-link login, OAuth — follow the default instead of the hardcoded `true` they carried before.

**Post-auth destinations are configurable.** `after_login_path` and `after_registration_path`, resolved through one helper. An explicit `return_to` always wins, and it now threads through every alternate sign-in route rather than dying the moment a user switches method.

**`require_email_confirmation`** (default true, unchanged behaviour) gates enforcement only — emails always send. All six gates honour it.

**`/users/confirm` no longer strands people.** It advances on mount when already confirmed (covering a `confirmed_at` flipped straight in the database, then a refresh) and live off the existing broadcast, so clicking the emailed link in another tab moves the parked one along.

## Bugs found along the way

Beyond the incident, mostly surfaced by three external reviewers:

- **`Routes.local_path?/1`** — the only redirect guard in the codebase — accepted ASCII control characters. Browsers strip tab/CR/LF while parsing, so `/\t/evil.example` reached `window.location` as `//evil.example`. `Phoenix.Controller.redirect/2` blocks these; LiveView's validator does not, and the new auto-advance is a LiveView redirect.
- **Changing a password** deleted every token including the one in the live cookie, leaving a cookie pointing at a dead token for 60 days — the same silent-logout shape as the original incident.
- **Magic-link registration never confirmed the account.** `confirmed_at` was dropped by the changeset's cast allowlist, so a user who had just proved inbox control was parked at a page whose flow sends no email.
- **Password reset was rate-limited only for addresses that existed**, so past the threshold the error message itself revealed which addresses were registered. **Confirmation resend had no limit at all** — an unauthenticated mail-flood vector and a timing oracle.
- **`after_login_path` accepted `/users/log-out`**, a real GET route: every login would sign the user straight back out, admin included.
- **Four role-based conn plugs** never checked email confirmation, and the shipped `:phoenix_kit_admin_only` pipeline has no preceding auth plug.
- **`Settings.get_setting/1` had no error handling** and the cached variants only `rescue`, but an unreachable database *raises* on an unowned checkout and *exits* on a dead pool — so a transient DB blip crashed callers, the login redirect resolver among them.
- **LayoutWrapper's standalone fallback** nested a second full document inside the LiveView and swallowed the page body, rendering auth pages as empty chrome with duplicate flash ids.
- **QR sign-in translations:** the button read "Sign in with password" in Russian, Estonian and French — gettext had fuzzy-matched all three, and fuzzy entries render live. Fixed in all seven catalogues with the flags dropped.

## Notes

- No version bump or CHANGELOG entry — left for the release cut.
- Docs follow the split made in `main`: summary in `AGENTS.md`, full reference in `dev_docs/guides/2026-07-28-login-and-registration.md`.
- Opened from a branch rather than `main` because `main` carries an unpushed AGENTS.md commit that conflicts with the restructure here.

**Still open, deliberately out of scope:** ~250–485 fuzzy gettext entries per language render auto-matched text, and spot checks show the same class of error elsewhere — Russian shows "Archived" as *Активно* (active), French shows "Accept" as *Tout accepter* and "Choose" as *Fermer*. That wants its own sweep.

## Verification

2216 tests, 0 failures across three consecutive runs; `mix precommit` (compile as errors, format, credo --strict, dialyzer) clean. 121 new tests, including a regression for each finding above.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -156,7 +156,17 @@ settings live on the admin Users settings page (`/admin/settings/users`).
   `/\`, and **ASCII control characters** (browsers strip tab/CR/LF, so
   `"/\t/evil.com"` lands as `//evil.com`; `Phoenix.Controller.redirect/2` blocks
   those but LiveView's `validate_local_url!` does not). Every LiveView
-  `redirect(to: ...)` of user-influenced input MUST go through it.
+  `redirect(to: ...)` of user-influenced input MUST go through it. The path
+  settings are additionally refused if they point at a page that bounces an
+  authenticated visitor — including **`/users/log-out`**, a real GET route that
+  would sign every user straight back out.
+- **Carrying `return_to`:** `Routes.return_to_query/1` threads it across the
+  links between login/register/magic-link/QR/OAuth, and the magic-link email
+  carries it in its URL. A new sign-in entry point must thread it too.
+- ⚠️ **Public auth endpoints rate-limit BEFORE the lookup** — limiting inside
+  the send only throttles addresses that resolve to a user, which turns the
+  deliberately generic copy into an account-existence oracle. Password reset had
+  exactly that shape; confirmation resend had no limit at all.
 - **Email confirmation:** `require_email_confirmation` (default true) gates
   *enforcement* only; emails always send. Honored at **six** sites — the
   `require_authenticated_user` / `require_authenticated_scope` plugs and the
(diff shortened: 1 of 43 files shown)


mdon:
Superseded — reopening from `main` so the work lands on the branch we actually integrate on. Same commits, plus Max's AGENTS.md commit ported into upstream's new docs split.

PR #678, opened 2026-08-04
Fix admin routes swallowed by the unconstrained :locale segment, and a sanitizer bypass

mdon:
Started from a host-app bug report: installing the shop module made `/admin/shop` unreachable. It turned out to be a core routing defect, and chasing it surfaced a second, wider one in the HTML sanitizer.

## The routing bug

`/admin/shop` bound to the shop module's **public** route `/:locale/shop` with `locale = "admin"`. Root-mounted installs were bounced to the storefront; prefixed installs redirected to themselves until the browser gave up. Present since at least 1.7.113 — about three months.

The scope carried `locale: ~r/^([a-z]{2,3}(?:-[A-Za-z]{2,4})?)$/`, which reads as a guard but never ran: `Phoenix.Router.Scope.push/2` honours only `:path, :alias, :as, :host, :private, :assigns, :log, :trailing_slash` and silently discards everything else. Phoenix has no route-segment constraints, so `:locale` matched any segment.

Fixed by emitting the admin and authenticated surfaces ahead of every `/:locale`-rooted block, with a load-bearing comment saying why. The dead regexes are gone rather than left looking protective.

## Also in here

- **`process_as_default_locale/1` was prefix-blind.** It stripped `"/#{locale}"` off `conn.request_path`, which still carries the mount prefix — a no-op on `/phoenix_kit/...`, which is what produced the self-redirect. Now rebuilt from `path_info` around the configured prefix, comparing *decoded* segments (Phoenix matches on a decoded copy but leaves `path_info` encoded), preserving `script_name`, and declining to redirect at all when it cannot produce a different path.
- **Three inert `status:` options removed.** `redirect/2` sends `conn.status || 302` and ignores `:status`, so the 307 and two 301s never happened. Left as 302s deliberately — a real 301 on the default-locale redirect would have clients permanently caching a redirect away from a locale that becomes valid the moment an admin changes the default language.
- **`admin_request?/1` matched `"admin"` as a substring**, so `/shop/product/admin-tools` counted as an admin request.
- **Shop admin routes were emitted twice** (36 duplicate groups): the module declares `route_module/0` *and* core hardcoded a `safe_route_call`. The hardcoded path is kept as a fallback for hosts where beam discovery misses, but skipped when discovery already covered it. 667 → 631 routes.
- **Host routers did not recompile when a discovered module's routes changed.** `module_hash/0` hashes only the *set* of module names, so editing `admin_routes/0` left it identical and new routes 404'd until `mix compile --force`. Fixed with compile-time references; verified both directions against a real host app.
- **Sanitizer bypass.** `` passed through completely untouched — the patterns required whitespace, but HTML allows `/` between attributes. This one is not shop-specific: it affects every consumer of `sanitize/1`, and the storefront renders product descriptions through it on the unauthenticated public page.

## Verification

- 0 shadowed routes across all 667 routes, all verbs, both root- and prefix-mounted
- `/admin/shop` renders the real dashboard when logged in; storefront unaffected
- 2527 tests, `mix precommit` exit 0
- New regression tests encode the *invariant* (no foreign param-rooted route before the gated surfaces), and were confirmed to fail when the router is mutated back

## Note on the sanitizer

Regex sanitizing is the underlying problem — this is the second bypass class found in it. The right shape is a parsed allowlist, which needs an HTML parser as a **runtime** dependency; `floki` is currently `only: :test`, and adding a runtime dep to core affects every consumer. That call is deliberately left to you; the moduledoc records it.

### DIFF
--- a/.dialyzer_ignore.exs
+++ b/.dialyzer_ignore.exs
@@ -143,5 +143,17 @@
   # opacity it never actually inspects. Not a runtime bug: the struct is passed
   # straight back to the module that owns it. Surfaced here by the dep upgrades
   # in 867bc5b2, which rebuilt the PLT.
-  {"lib/phoenix_kit/users/qr_login.ex", :call_without_opaque}
+  {"lib/phoenix_kit/users/qr_login.ex", :call_without_opaque},
+  # `MDEx.safe_html/2`'s typespec declares `escape: [atom()]`, but the
+  # implementation reads `escape` as a KEYWORD list —
+  # `opt(options, [:escape, :content], true)` — and MDEx's own doctests pass
+  # `escape: [content: false]`. The spec contradicts the code it documents,
+  # so dialyzer concludes the call can never return.
+  #
+  # Passing the keyword form is required, not cosmetic: with content
+  # escaping left on, the sanitizer returns `&lt;p&gt;Hello&lt;/p&gt;` and
+  # every piece of rich text in the app renders as literal markup.
+  # Re-check on the next MDEx upgrade; drop this entry once the spec is
+  # corrected upstream.
+  {"lib/phoenix_kit/utils/html_sanitizer.ex", :no_return}
 ]
--- a/lib/phoenix_kit/utils/html_sanitizer.ex
+++ b/lib/phoenix_kit/utils/html_sanitizer.ex
@@ -1,51 +1,59 @@
 defmodule PhoenixKit.Utils.HtmlSanitizer do
   @moduledoc """
-  HTML sanitization for rich text content in entities.
+  HTML sanitization for user-supplied rich text.
 
-  This module provides basic HTML sanitization to prevent XSS attacks
-  while allowing safe HTML tags commonly used in rich text editors.
+  Backed by `MDEx.safe_html/2` — an ammonia-based **parsed allowlist**. It
+  builds a real DOM, keeps only known-good tags and attributes, validates
+  URL schemes per attribute, and re-serializes. It does not pattern-match
+  on markup, which is why it holds up where the previous regex
+  implementation did not (see `sanitize_html/1`).
 
-  ## Allowed Tags
+  ## What survives
 
-  The following tags are allowed:
-  - Block elements: p, div, br, hr, h1-h6, blockquote, pre, code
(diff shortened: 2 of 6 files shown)


mdon:
Two commits added since the description above, both from review passes on this branch's own work.

**Sanitizer rewritten as a parsed allowlist.** The regex fix in this PR closed one bypass class; a security pass then found the approach itself is the problem — a regex has to model a parser it is not, and this was the second leak in it. Replaced with `MDEx.safe_html/2`, an ammonia-backed parsed allowlist that builds a DOM, validates URL schemes per attribute, and re-serializes. MDEx was **already a runtime dependency** (it renders Markdown), so this adds no dep and the allowlist is maintained upstream.

Notable gain: entity-encoded schemes. The old blacklist could be walked past with `jav&#x61;script:`; a parser decodes before validating.

Output is now normalised HTML rather than a lightly-edited copy of the input (`<br/>` → `<br>`, attributes quoted, `<table>` gains a `<tbody>`). Three tests asserting byte-exact old output were rewritten to assert the security property — one of them had been failing on a *correct* result, where the parser folded `onerror` into a quoted `src` URL where it is inert.

One dialyzer ignore was needed: MDEx's `safe_html/2` spec declares `escape: [atom()]` while the implementation reads a keyword list and its own doctests pass one. The keyword form is required — with content escaping left on, every piece of rich text renders as literal markup. Recorded with a note to drop it when the upstream spec is fixed.

**Guest carts were being lost on login.** Signing the shop session cookie in phoenix_kit_ecommerce#12 broke `maybe_merge_guest_cart/2` here, silently. `:browser` fetches cookies unsigned, so `conn.cookies["shop_session_id"]` is the signed envelope rather than the id; it is truthy, so `||` never fell through to the session where the real id lives. `merge_guest_cart/2` matched no cart and returned quietly — every guest who filled a cart and logged in lost it, with no error anywhere.

Now reads the session first, which the shop plug mirrors on every request and which is already verified. Correct independently of signing; the cookie fallback stays for hosts that wire the shop routes without the plug.

Found by a host-integration boundary audit — the failure sits exactly on the seam that audit exists to check, where one repo changed a format and the reader in another kept parsing the old one.

Gates still green: 2533 tests, `mix precommit` exit 0.

PR #685, opened 2026-08-07
Core-owned redirect destinations: one resolver, and an /admin landing that admits everyone

timujinne:
## The problem

Core sent users "home" by calling `Routes.path("/")`. That prefixes the active locale, so the redirect arrived as `/en` — a route the **host** owns and frequently never declares, so it 404s. Measured in one application before its workaround: `/` answered 200 while `/en`, `/et` and `/ru` all 404ed.

It was not one bug in one flow. `Routes.path("/")` was hardcoded at **nine** call sites across five files, plus two more that hardcoded the literal `"/"`. Every application with the Languages module enabled is exposed; with `default_language_no_prefix` off — the default — every locale breaks.

## What this does

**One resolver.** `Routes.safe_destination/2` replaces every hardcoded destination; no call site decides where to send anyone. Candidates are validated as local, non-auth-page and actually routable (`Phoenix.Router.route_info/4` against the router on the conn/socket) before anyone is sent there.

**A terminal that can be asserted rather than probed.** The authenticated chain ends at `/admin`, the anonymous chain at `/users/log-in` — both declared unconditionally by core. For that to be honest, `/admin` had to admit every authenticated visitor, and it rendered 373 lines of operator content with **zero permission checks**. So gating is not polish here, it is the precondition; both halves land together.

**Admission is a gate exception, not a route move.** `:phoenix_kit_ensure_admin` answers `:landing` for the dashboard index and enforces every other view exactly as before. `/admin` stays in `live_session :phoenix_kit_admin` with the rest of the admin surface. Moving the route was tried and reverted: it put the guaranteed landing behind an `on_mount` that can halt and redirect (a real infinite redirect under a documented context-selector config), split `/admin` from the other admin routes so every navigation between them became a full page reload, and reordered the pipeline so an anonymous visitor to `/et/admin` got a login page in the wrong language.

**Gates derive from the destination.** `Auth.can_access_admin_view?/2` is a new public predicate; `enforce_mapped_admin_view_permission/2` now shares its implementation, so a rendered card and the page it links to cannot disagree. The data is gated too, not just the markup: a visitor who may not see the statistics triggers no aggregate query, no migration read and no PubSub subscription. Gates recompute on a mid-session permission change and genuinely unsubscribe rather than merely hiding.

**A new `main_page_path` setting** lets an operator name the page anonymous visitors land on, validated as a local path and re-checked for routability on read.

`/dashboard` is untouched — it keeps its deprecation, its setting and its routes.

## Behaviour change for existing hosts

`/admin` now admits every authenticated user. They see a greeting and nothing else unless they hold permissions; the sidebar, the cards and both statistics sections are gated. This is deliberate and worth a release note.

## Verification

- Full suite against PostgreSQL: **18 doctests, 2813 tests, 0 failures**. (Without a database the `:integration` half self-excludes and the run still exits 0, so it was run against a real one.)
- `mix credo --strict`, `mix format --check-formatted`, `mix compile --warnings-as-errors` (dev and test), `mix dialyzer` — all clean.
- Live, against four real role shapes: an Owner sees everything; a Manager holding 19 of the permission keys sees the cards but not the statistics; a client-portal user sees the greeting and their own menu; a user with no permissions sees the greeting and no admin navigation at all.
- The three new catalog strings are translated for de/es/et/fr/it/pl/ru.

## Known, deliberate

- Platform Statistics and System Information are gated on `Scope.holds_all_enabled_permissions?/1`. A role granted exactly the `dashboard` key used to see them and no longer does — flagged rather than assumed.
- The notifications bell is a sticky nested LiveView, so its "View all" link reflects permissions as of mount until the next full page load. Fail-closed: the destination re-gates.


timujinne (commented):
> Automated review by **GLM-5.2** (external model, read-only access to the branch), run against the full PR diff and the committed source. Posted verbatim; the provenance note is the only text added.

---


## Review: PR #685 — Core-owned redirect destinations

### Verdict

This is a clean, carefully-reasoned change. I read the full diff and the design doc, then verified every load-bearing claim against the committed source — the resolver chain, both gate predicates, the scope-refresh ordering, the new overview module, the sidebar/LayoutWrapper gating, and all eleven call sites. I could not construct a failing scenario: the landing exemption is authentication-gated on both the mount and refresh paths, the resolver provably terminates under `:skip_admin`, operator data is gated before any query runs, and every library-consumer shape I tried resolves to a core-owned landing. No blocking findings.

### Findings

None that survive verification.

### What I checked and could not break

**Resolver termination / open-redirect (`routes.ex`).** Walked `safe_destination/2` (298–319) under every flag combination. Under `:skip_admin` with `after_login_path` pointed at `/admin/users` and `/dashboard` compiled out, every admin-area candidate is dropped categorically by `admissible_candidate?/3` (450–456) via `admin_area_path?/1` (473–483), so only the unfiltered terminal `/admin` survives — and `/admin` renders rather than bounces, so the chain ends. The `admin_area_path?/1` segment matcher correctly handles both the prefixless and locale-prefixed shapes (`/phoenix_kit/admin`, `/phoenix_kit/et/admin`) and does *not* swallow `/administrators`. `local_path?/1` (60–67) still rejects `//`, `/\`, and the ASCII-control-char trick, and `auth_page?/1` suffix-matches so `?return_to=/users/log-out` can't sign a fresh login straight back out. Terminals are asserted, not probed (391–397); `routable?/2` is diagnostic-only there and fails closed when no router is readable.

**Landing gate (`auth.ex`).** `admin_gate_decision/2` (830–836) asks `landing_view?` *first*, so a client holding only `client_portal` is not sent into `:enforce_view` for the page the resolver just guaranteed them. The exemption is auth-gated on both sides: mount (`landing_view?(view) and Scope.authenticated?(scope)`) and refresh (`scope_refresh_decision`, 1302–1315). Confirmed the sole real `/admin` route is `Live.Dashboard` (integration.ex:492; line 116 is a docstring example, not a route) and `@landing_view` is `PhoenixKitWeb.Live.Dashboard` (auth.ex:756), so `socket.view == @landing_view` holds. `Scope.authenticated?(nil)` and `Scope.can_access_admin_area?(nil)` both return `false` (scope.ex:197, 349) — the nil-scope paths fail closed everywhere.

**Data gating, not just markup (`overview.ex`).** `assign_scope_gates/1` (213–227) computes the verdict *before* `assign_statistics/2`, so a permission-less visitor triggers none of `Roles.get_extended_stats`, `Sessions.get_session_stats`, `Presence.get_presence_stats`, or the migration reads — they return nils (400–408). The single `handle_info` clause is tag-guarded (`is_overview_message`, 163–165) with a catch-all in `apply_visible_statistics_message` (318), so an in-flight or reshaped broadcast after unsubscribe is a stale tile, not a crash. `refresh_statistics/1` (240–250) and `apply_statistics_message/2` (269–280) both re-check the gate from current state rather than trusting the hidden button — a crafted `phx-click` from a demoted visitor costs nothing.

**Mid-session permission change (`auth.ex`).** `handle_scope_refresh` (1229–1258) reassigns the scope, then calls `refresh_view_scope_assigns/1` → `phoenix_kit_scope_changed/1` (which re-runs `assign_scope_gates` against the *new* scope), *then* decides evict/stay — so a demoted operator on the landing releases the three PubSub subscriptions (`sync_statistics_subscription`, 345–365, both directions) before the page settles. The ordering is load-bearing and correct: subscriptions close whether or not the visitor is also navigated away.

**Sidebar + LayoutWrapper consistency.** `reachable_tabs/2` (admin_sidebar.ex:144–151) returns `[]` for a scope failing `can_access_admin_area?`, so a permission-less visitor on the new universal landing gets no `<nav>` at all (97–98). For live-view entries it asks the destination's own `can_access_admin_view?/2` (168–172) — the same function the mount gate uses — so a rendered entry and its destination cannot disagree. `show_admin_nav` and `can_open_inbox` (layout_wrapper.ex:325, 331–335) are derived from the same gates.

**Call sites.** All eleven are consistent. The two that previously read `conn.assigns` now read the session token via `conn_scope/1` (session.ex:156, oauth.ex:365) — correct, because account switching has already mutated the session. Full logout passes `scope: nil` explicitly (auth.ex:334). The Owner-denial path (auth.ex:674) deliberately omits `:skip_admin` because the terminal `/admin` never bounces an authenticated visitor, so there is nothing to loop on.

**Library-consumer shapes.** `user_dashboard_enabled: false` (route compiled out): the `/dashboard` candidate is probed away, the authenticated terminal is `/admin`, which admits everyone — safe. Custom admin tabs: `best_available_admin_path` (1512–1521) includes `custom_admin_fallback_routes`, and a denied visitor lands on a permitted tab or the terminal. Non-default `url_prefix`: candidates are built with `path/1` (prefix-aware) and probed against the same router, and `strip_url_prefix` (487–496) mirrors `path/1`'s `base_path`, so `admin_area_path?/1` stays correct under any prefix.

**Non-blocking note (not worth changing):** `assign_statistics(socket, true)` re-runs `Migrations.current_version/0` and `migrated_version_runtime/1` on every scope-change/module-toggle refresh for an operator, not just on mount — the migration version can't change on a permission edit. It's cheap and the single-gate-site rationale justifies it; mentioning only for completeness.

### Verdict

**APPROVE.**

PR #686, opened 2026-08-07
Fix four account-takeover paths found by a security review

timujinne:
Found by a security review of the whole workspace on 2026-08-07 and verified against a running application. Four independent paths, each of which ends with an attacker holding someone else's account. Each commit is self-contained and carries its own reasoning.

## 1. Credential management on the admin user form ignored rank

`/admin/users/edit/:id` gated the password field on `Scope.can_access_admin_area?/1` — true for **any** holder of a single permission — while the template comment above it said "Admin/Owner only". Nothing else in the path checked rank: not the pipeline, not the on_mount hook, not the changeset, not a caller one frame up. Any role granted `users` (which Admin is auto-granted at boot) could open the Owner's record and set a new password.

`Auth.can_manage_user_credentials?/2` now answers by role and rank, mirroring the rules `can_delete_user?/2` and the impersonation authority already enforce, so the three takeover surfaces agree. It gates the password field, the reset-mail button and the **email** field — owning the address a reset link is delivered to takes an account just as surely as setting the password does. The events refuse independently of the template.

## 2. Deactivation denied access but did not revoke it

`update_user_status/2` flipped `is_active` and broadcast; the token rows stayed valid for their full 60-day life. Every gate that reads a session applies `ensure_active_user/1`, so an operator sees the account bounce and believes it is off — while any entry point that resolves a token without that filter still sees a live user. Deactivation now deletes the session tokens.

## 3. Multi-session resolved the root account without the active-user filter

`root_user/1` and `root_authenticated?/1` used a bare token lookup where `switch_to/2` uses `ensure_active_user/1`. With (2), a deactivated Owner/Admin holding a live cookie could `POST /users/session/impersonate/:uuid` — a route in a pipeline with no auth gate — and mint a fresh session as any other active user.

## 4. Stored XSS in the auth-page branding settings

`AuthPageWrapper.bg_style_tag/1` built a `<style>` element by string concatenation and emitted it with `raw/1`. A `<style>` body is raw character data, so HTML escaping does not apply and `</style>` in a value closes the element. The background colour is free text on `/admin/settings/authorization`, whose key is `settings` — held by Manager as well as Admin — and the result is served to every **anonymous** visitor of the login, registration and reset pages.

New `PhoenixKit.Utils.CssValue` is an allowlist that fails to `""`, applied on read (neutralising values already stored), in the one function that assembles the stylesheet, and on write (so the operator learns the value was rejected).

## 5. OAuth attached an external identity on email string equality

`find_or_create_user/3` looked the callback address up locally and, on a hit, confirmed that account and issued a session. No verification claim was read — although `extract_oauth_data/1` already captures `raw_info` — no prior link was required, and the query was unscoped.

Resolution is now ordered by the proof each case carries: an existing `(provider, provider_uid)` link is decisive and consults no email; a pre-existing local account requires the provider's own assertion (Google/OIDC `email_verified`, GitHub per-address `verified`, Facebook `verified`); a new account is created but auto-confirmed only on an assertion. `oauth_require_verified_email` (default `true`) lets a deployment lift the requirement deliberately rather than by accident.

## Also

`AGENTS.md`'s storage TODO gains the unauthenticated `POST /api/upload` defect (no auth plug on the scope; owner taken from `params["user_uuid"]` under a comment promising a check that was never written). Implementation is being handled separately.

## Verification

- `mix test test/integration/users` — **341 tests, 0 failures** against a real PostgreSQL, including 19 new ones covering every case above.
- `mix format --check-formatted`, `mix credo --strict` (10435 mods/funs, no issues), `mix compile --warnings-as-errors` — all clean.

## Compatibility

Behaviour changes an operator will notice, all deliberate:

- An Admin can no longer set an Owner's or another Admin's password, mail them a reset, or change their email address. Editing their other profile fields is unaffected.
- Deactivating a user now logs them out immediately instead of eventually.
- An OAuth sign-in that would attach to a pre-existing local account is refused unless the provider vouches for the address. Users already linked are unaffected, since the link is matched first.


timujinne (commented):
## Independent adversarial review — two reviewers, both REQUEST-CHANGES; findings addressed in 3 follow-up commits

This PR was reviewed twice, independently and in parallel, by two models given the same brief: *defeat each fix, then look for the same shape elsewhere*. Neither saw the other's output. Both returned REQUEST-CHANGES on this PR and APPROVE on the sibling `phoenix_kit_emails#27`.

Posting the merged result here for the record, since three of the findings were defects in the security fixes themselves rather than in the code they fixed.

---

### HIGH — the activation rank check was live on one page out of three *(both reviewers, independently)*

The previous commit claimed activation was decided by rank. It was — in the admin edit form. `Auth.update_user_status/2` is reached from three LiveViews:

| Entry point | Guard before this review |
|---|---|
| `lib/phoenix_kit_web/users/user_form.ex:426` | rank check ✅ |
| `lib/phoenix_kit_web/live/users/users.ex:422` → `:605` → `:658` | self-check only |
| `lib/phoenix_kit_web/live/users/user_details.ex:253` → `:483` | nothing |

All three map to the `users` permission key (`web/users/auth.ex:1829`), so reachability is identical: any custom role granted `users` — not Admin, not Owner. The menu item rendered unconditionally and its wrapper hid only Owner targets, so **Admin was one click away**; Owner was one hand-built LiveView event away, since `Roles.can_deactivate_user?/1` protects only the *last* Owner. Commit 2 of this PR made it worse by revoking tokens on deactivation, turning it into an immediate forced sign-out of the accounts that outrank the actor.

**Fixed** by moving the rule to the function all three go through: `update_user_status/3` takes `actor:` and refuses with `:insufficient_permissions`; omitting it is the explicit system path (`Users.Referrals`). Menu items now hidden with the same predicate, matching the existing `can_delete_user?/2` idiom.

`toggle_user_confirmation/2` had the identical shape and is now gated the same way — `require_email_confirmation` is honoured at eleven gates, so unconfirming an Admin locks them out of every protected page.

### HIGH — the OAuth verified-email gate could not be satisfied for GitHub, and its documented switch did not exist *(second reviewer)*

Two halves, both mine:

1. `OAuthConfig` registered the strategy as `{Ueberauth.Strategy.Github, []}`, i.e. with the strategy's default scope of `""` (`deps/ueberauth_github/…/github.ex:76`). Without `user:email` the token cannot read `GET /user/emails`; the strategy stores the user with **no `"emails"` key**, so the verification lookup had nothing to find. With the requirement on by default, **every GitHub sign-in to an existing account was refused** and every new GitHub account was created unconfirmed.
2. `oauth_require_verified_email` was read in exactly one place and declared nowhere — not in `Settings.get_defaults/0`, not in the settings schema, not in the admin UI. The "deliberate, operator-visible decision" the docstring promised was reachable only from iex.

**Fixed**: the scope is requested; the setting is registered with a default of `"true"` and a checkbox on `/admin/settings/authorization` next to the providers it governs. Tests now pin the no-`"emails"`-key shape and both positions of the switch.

### MEDIUM — a comment justified the code with something untrue

`user_by_provider_identity/1` said `(provider, provider_uid)` has no unique index. It does — `phoenix_kit_oauth_providers_provider_uid_idx`, migration V16. Corrected, and the comment now records what that branch actually means: an existing link is treated as proof, which is right for a provider whose uid is server-assigned and immutable and **not** right for one whose `sub` a user can choose or recycle.

### MEDIUM — deactivated accounts were still issued sessions by three of four entry points

`is_active` was checked by the password controller only; magic-link verification, QR-login completion and the OAuth callback call `log_in_user/3` directly and it did not check. The fetch plugs filtered afterwards, so the effect was a dead session rather than access — but that correctness depended on every plug remembering, which is the dependency this branch exists to remove. The check now sits on the shared funnel.

### LOW — fixed alongside

- `can_manage_user_status?/2` refused nothing for **self**. The admin form has always said "You cannot deactivate your own account for security reasons" — in markup only. Now refused server-side; credentials keep the opposite rule, since your own password *is* yours to change.
- `update_user_profile/3` dropped `password` and `email` for an actor without credential authority, but not `username` — the second identifier `get_user_by_email_or_username_and_password/3` accepts.
- Unreachable `validate_admin_authority_over/2` catch-all removed.

### Test quality — two tests were asserting nothing

Both reviewers checked whether the tests would fail with the fix reverted:

- Two `CssValue` assertions (`url(https://evil.example/x)`) are refused by the character allowlist alone, so they would pass with `@color_forbidden` deleted. Replaced with `url(x)` / `URL (x)`, which consist only of permitted characters and can be rejected by nothing else.
- The OAuth tests never reproduced the shape GitHub actually delivers, and never exercised the setting. Both added.
- The status tests asserted the predicate in isolation while two pages bypassed it entirely — which is precisely why the HIGH survived. They now bind the rule to the function every caller goes through.

---

### Confirmed correct by both reviewers

`PhoenixKit.Utils.CssValue` (one reviewer reimplemented both filters and fuzzed them; no breakout of the `<style>` element or the `url('…')` token was constructible, and `%27` is a URL character rather than a terminator in CSS) · credential management on the edit form, with every caller of `admin_update_user_password/3` and `deliver_user_reset_password_instructions/3` enumerated · multi-session root resolution and the three added `ensure_active_user/1` filters · role escalation already closed upstream by `Roles.sync_user_roles/3` · no regressions in the existing suites.

### Verification after the follow-ups

`mix test test/integration/users test/phoenix_kit/utils` — **528 tests + 16 doctests, 0 failures** against a real PostgreSQL. `mix compile --warnings-as-errors`, `mix format --check-formatted`, `mix credo --strict` (10446 mods/funs) all clean.

### For the release notes, not the code

- Link rows in `phoenix_kit_oauth_providers` written before this change were created by the email-equality path. The new gate protects new attachments and cannot retroactively re-verify old ones — worth an audit query on any install where a provider was enabled before this release.
- Deactivation revokes tokens but does not disconnect an already-mounted LiveView; `on_mount` does not re-run until the next live_session transition or reload.

PR #694, opened 2026-08-09
Stop reporting a comment-less database as version 1

timujinne:
Three findings from the post-merge review of the migration work, published as a
comment on #689. All verified against the tree before this branch was cut.

## The one that matters

`Common.query_version_directly/2` ended with, verbatim:

```elixir
_ ->
  # Table exists but no version comment - assume version 1
  1
```

A database that is **current but lost its comment** — the half-installed or
adopted state `repair --adopt` exists for — therefore reported
`{:current_version, 1}`, which is below the floor, so `mix phoenix_kit.update`
answered: *install the 1.7.x bridge first.*

`Postgres.migrated_version/1` refuses that same input, and says why:

> the guess would route a possibly CURRENT database to the 1.7.x bridge, whose
> backfill overwrites still-NULL tracked columns with freshly generated uuids
> pointing at nothing

Downstream that is not abstract: `UUIDFKColumns.set_not_null/4` backfills
legitimately-NULL `phoenix_kit_files` rows with invented uuids, and
`cleanup_orphaned_fk_refs/5` deletes them for matching no user.

Two halves of one release gave opposite instructions for one state, and the
operator met the destructive one first. Worth noting the sibling fallback in the
same module was already hardened against exactly this guess — its comment
records that it "used to fall back to a fabricated `{:current_version, 1}`…
Report honestly instead". The fabrication survived one level down.

`:unknown_version` is now distinct from both a real version and "not installed",
and the update task routes it to `doctor` + restamp.

## Also fixed

- **`Repair.Probe.read_comment/2` raised on the anomaly it exists to diagnose.**
  `String.to_integer/1`, eleven lines below a docstring that says "Never raises".
  The exact hand-edits the migrator documents (`'v164'`, `' 164'`) ended
  `mix phoenix_kit.repair` in a bare `** (ArgumentError) argument error`, while
  `doctor` survived only because its check wraps everything in `rescue`.
- **The below-floor notice did not name the bridge.** `@bridge_version` was
  threaded into both `BelowFloorError` raise sites but not into the notice shown
  *first*, at generation time — the raises come later. `bridge_version/0` exposes
  it so the two agree.

## Not in this PR

Two findings from the same round need a decision rather than a patch, so they
are left for the maintainer:

- The manifest pins `uuid_generate_v7`'s body with `public.gen_random_bytes`, so
  `repair` can never converge where pgcrypto lives in another schema — a
  supported topology. Wants the pgcrypto schema normalised out of `prosrc`
  before hashing on both sides, like the array-cast canonicalisation in
  `Differ`.
- The manifest's pre-164 revisions for `idx_publishing_posts_group_slug` carry
  `predicate: nil`, next to a comment saying "Every real install has the
  predicate". Both cannot hold: a public install at 135..163 gets `:wrong_shape`
  and exit 2 on a byte-correct database. Wants `:legacy_optional` or a bimodal
  revision.

And one that is already yours: the review also flagged that
`generate_baseline.exs` does not encode the hand-applied post-generation
corrections. `90077b79` answers that better than the finding did — the generator
*cannot* emit V164+ objects, because regeneration runs from a pre-squash
checkout whose chain ends at V163. Recorded here only so the thread is complete.

## Verification

`mix compile --warnings-as-errors`, `mix format --check-formatted` and
`mix credo --strict` (10290 mods/funs) — clean. No test run: these paths are
exercised by `mix phoenix_kit.update` against a real database rather than by the
suite, and the shared PostgreSQL has been at its connection ceiling all day.
`@version` and `CHANGELOG.md` untouched.

### DIFF
--- a/lib/mix/tasks/phoenix_kit.status.ex
+++ b/lib/mix/tasks/phoenix_kit.status.ex
@@ -241,6 +241,11 @@ defmodule Mix.Tasks.PhoenixKit.Status do
       {:unreachable, reason} ->
         {:unreachable, reason}
 
+      # `status` is the task an operator runs precisely because something looks
+      # wrong, so it must report this state rather than raise on it.
+      {:unknown_version} ->
+        {:unknown_version}
+
       {:current_version, version} ->
         target_version = Postgres.current_version()
 
@@ -339,6 +344,13 @@ defmodule Mix.Tasks.PhoenixKit.Status do
     "#{IO.ANSI.yellow()}Unknown — database unreachable#{IO.ANSI.reset()}"
   end
 
+  # Installed, but the version comment is missing or unreadable. Red rather than
+  # yellow: unlike an unreachable database this will not fix itself, and no other
+  # task can act safely until it is restamped.
+  defp format_installation_status({:unknown_version}) do
+    "#{IO.ANSI.red()}Installed, version comment unreadable#{IO.ANSI.reset()}"
+  end
+
   defp format_installation_status({:up_to_date, version}) do
     "#{IO.ANSI.green()}V#{pad_version(version)} ✅#{IO.ANSI.reset()}"
   end
@@ -401,6 +413,12 @@ defmodule Mix.Tasks.PhoenixKit.Status do
 
   defp format_next_action({:fix_connection, message}) do
     "#{IO.ANSI.yellow()}#{message}#{IO.ANSI.reset()}"
+  end
+
+  # Same shape as :fix_connection — a state nothing else can act around until an
+  # operator resolves it by hand.
+  defp format_next_action({:fix_version_comment, message}) do
+    "#{IO.ANSI.red()}#{message}#{IO.ANSI.reset()}"
   end
 
   # Layout lives in PhoenixKit.Install.StatusTree so it can be unit tested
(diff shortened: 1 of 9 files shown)


timujinne:
Two more findings from the same review round pushed to this branch — I had
dropped them from the first commit without listing them anywhere, so the
"Not in this PR" section above was incomplete.

- **`fk_exists?/3` verified a foreign key by name only.** Detection before the
  ADD is shape-based, so a constraint that merely owns the name reads as
  `:absent`; the ADD then fails 42710 into the `EXCEPTION WHEN OTHERS` handler,
  the name-only check sees the colliding constraint, and the outcome is reported
  `:created` — comment stamped 164, declared FK absent. `contype = 'f'` closes it.
- **V163 scanned before it checked the size guard.** `castable?/3` is a
  full-table `count(*) … !~*`; asking it before `estimated_rows/3` made the
  deferral path pay an unbounded sequential scan on exactly the tables the limit
  targets.

The reorder changes which message an oversized **and** uncastable table gets —
the size warning instead of the cast error. Deliberate: that branch points at
`mix phoenix_kit.repair_uuid` in a maintenance window, where the cast failure
surfaces with the table in hand.

Still not in this PR, unchanged from the description: the pgcrypto-schema
normalisation in the manifest's `uuid_generate_v7` body, and the `predicate: nil`
revisions for `idx_publishing_posts_group_slug` below 164. Both want a decision
about manifest semantics rather than a patch from me.

PR #695, opened 2026-08-09
Close the remaining review findings from the #694 round

timujinne:
#694 merged at its third commit; the work after that never reached main. This is
that remainder, rebuilt onto current main.

**Dropped deliberately:** my version of the comment-less guard. `80fc6784`
replaced it with a better one — mine required a positive "there is no comment"
and fell through to the destructive branch on any failure to answer, while
`comment_literally_says_one?/1` requires a positive "the comment reads 1" and
treats everything else as unknown. That is the right direction and it is already
in main.

## Still live on main

**`mix phoenix_kit.status --verbose` crashes.** The non-verbose path reports the
unreadable-comment state correctly; `show_installation_diagnostics/2` has no
clause for it, so the run dies with `CaseClauseError` right after printing a
correct tree. This is the clause a previous commit message claimed and did not
contain.

## Introduced by #694 and fixed here

**A flaky FK probe aborted the entire chain run.** Moving V164's post-ADD check
to `fk_shape_present/5` was right for correctness and wrong for failure mode:
that probe raises, while the `fk_exists?/3` it replaced returned false. V164 has
no `rescue` and no `run_isolated/3` equivalent, so one failed catalog read on one
of ~70 constraints aborted mid-run — earlier repairs already auto-committed under
`@disable_ddl_transaction`, `COMMENT … IS '164'` never reached, whole version
replays next time. A probe failure is now a `{:failed, …}` entry in the summary.

## Pre-existing, found by the same review

- **`comments_fk_on_delete/1` matched by name alone.** A CHECK, or an FK on a
  different column, owning `fk_comments_user_uuid` answered for the real one — a
  `confdeltype` of `'n'` from an impostor reads as "already SET NULL", so
  `repair_comments_fk/2` returns `[]` and the genuinely missing FK is never
  created. Now anchored on `contype`, arity and column. `fk_validated?/3` stays
  name-only and now documents why: every caller reaches it through the shape gate.
- **`Repair.Probe` collapsed a corrupt comment into "absent".** Both returned
  `nil`, so `adopt_required_message/0` said the comment was "missing" while the
  table held `v164`, and `--adopt` stamped the floor over it — destroying the only
  record of what the last operator believed, in the state where that belief is the
  most useful thing available. A fourth value, `:unparseable`, maps to its own
  `:comment_unreadable` branch that refuses rather than adopting.
- **`estimated_rows/3` reported a never-analyzed table as 0 rows.** PostgreSQL
  ≥ 14 leaves `reltuples = -1` until the first vacuum or analyze — `pg_restore`,
  `CREATE TABLE AS`, logical replication, autovacuum off. Measured: 200k rows read
  as zero, so the size guard passed and V163 would take `ACCESS EXCLUSIVE` for a
  full rewrite on exactly the table the limit exists to protect. It now returns
  `:unknown`, V163 defers on it, and `mix phoenix_kit.repair_uuid` prints
  "size unknown — never analyzed" instead of "~0 rows" to whoever is sizing the
  maintenance window.
- **V163's deferral text named a remedy that cannot work** on a table that is
  also uncastable — since the reorder, such a table takes that branch.

## Verification

`test/phoenix_kit` — 1153 tests, 38 doctests, 0 failures (359 excluded: the
integration half, no database reachable). `mix format --check-formatted`,
`mix compile --warnings-as-errors`, `mix credo --strict` (10297 mods/funs) clean.

⚠️ `mix dialyzer` halts on this branch, on two `LocaleSlug` warnings in
`lib/phoenix_kit/utils/slug.ex` — a file this branch does not touch, with no
entry in `.dialyzer_ignore.exs`. Pre-existing, so `mix quality` / `precommit`
fail here for a reason unrelated to this PR; worth someone's attention
separately.

`@version` and `CHANGELOG.md` untouched.

### DIFF
--- a/lib/mix/tasks/phoenix_kit.repair_uuid.ex
+++ b/lib/mix/tasks/phoenix_kit.repair_uuid.ex
@@ -93,11 +93,21 @@ defmodule Mix.Tasks.PhoenixKit.RepairUuid do
     if UUIDIntegrity.castable?(repo, qualified, table) do
       rows = UUIDIntegrity.estimated_rows(repo, prefix, name)
 
+      # `:unknown` is said out loud rather than printed as a number. This task
+      # is the maintenance-window path the migration defers TO, so the operator
+      # deciding whether they have a window needs to know the size is a guess
+      # nobody has ever taken — not read "~0 rows" and size the window for it.
+      size =
+        case rows do
+          :unknown -> "size unknown — never analyzed"
+          n -> "~#{n} rows"
+        end
+
       Mix.shell().info([
         :cyan,
         "\n#{name}",
         :reset,
-        " (~#{rows} rows) #{UUIDIntegrity.describe(table)}"
+        " (#{size}) #{UUIDIntegrity.describe(table)}"
       ])
 
       announce_duplicates(repo, qualified, table)
--- a/lib/mix/tasks/phoenix_kit.status.ex
+++ b/lib/mix/tasks/phoenix_kit.status.ex
@@ -482,6 +482,16 @@ defmodule Mix.Tasks.PhoenixKit.Status do
       {:unreachable, reason} ->
         IO.puts("  Database unreachable: #{inspect(reason)}")
 
+      # The non-verbose path reports this state; without a clause here the
+      # SAME run crashed with CaseClauseError the moment `--verbose` was added
+      # to it — after printing a correct tree, which is the worst place to die.
+      {:unknown_version} ->
+        IO.puts("  Current version: unreadable — the table exists but its")
+        IO.puts("  version comment is missing or is not a plain integer.")
+        IO.puts("  Establish the real state and restamp it by hand:")
+        IO.puts("    mix phoenix_kit.doctor")
+        IO.puts("    COMMENT ON TABLE #{prefix}.phoenix_kit IS '<version>';")
+
       {:up_to_date, version} ->
         IO.puts("  Current version: V#{pad_version(version)}")
         IO.puts("  Target version: V#{pad_version(Postgres.current_version())}")
--- a/lib/phoenix_kit/migrations/postgres/v163.ex
+++ b/lib/phoenix_kit/migrations/postgres/v163.ex
@@ -131,13 +131,17 @@ defmodule PhoenixKit.Migrations.Postgres.V163 do
       # table that prompted this work is an events table, and had it been uuid-
       # typed-but-keyless it would have taken an index build over every row
       # during `mix ecto.migrate` — exactly the outage this limit exists to stop.
-      UUIDIntegrity.estimated_rows(repo(), prefix, name) > limit ->
+      too_big_or_unmeasured?(UUIDIntegrity.estimated_rows(repo(), prefix, name), limit) ->
         Logger.warning("""
         [PhoenixKit V163] #{name}: #{UUIDIntegrity.describe(table)} — but the \
-        table holds more than #{limit} rows. Skipped: the repair takes an ACCESS \
-        EXCLUSIVE lock for a full pass over the table. Run this in a maintenance \
-        window, where the key can be built CONCURRENTLY:
+        table is over #{limit} rows, or has never been analyzed so its size is \
+        unknown. Skipped: the repair takes an ACCESS EXCLUSIVE lock for a full \
+        pass over the table. Run this in a maintenance window, where the key can \
+        be built CONCURRENTLY:
           mix phoenix_kit.repair_uuid #{name}
+        That command reports the row count, and refuses with the offending values \
+        if the column holds anything that is not a valid UUID — fix those first if \
+        it does, because no repair can convert them.
         `mix phoenix_kit.doctor` will keep reporting it until you do.
         """)
 
@@ -156,6 +160,15 @@ defmodule PhoenixKit.Migrations.Postgres.V163 do
         run_isolated(qualified, prefix, table)
     end
   end
+
+  # `:unknown` defers, exactly as an over-limit count does. The guard exists to
+  # keep an unbounded ACCESS EXCLUSIVE pass out of `mix ecto.migrate`, and "we
+  # never measured this table" is not evidence that the pass is bounded. The
+  # cost of being wrong is asymmetric: deferring leaves the table unrepaired and
+  # loudly reported by `doctor`, while proceeding can lock a large table for the
+  # length of a deploy.
+  defp too_big_or_unmeasured?(:unknown, _limit), do: true
+  defp too_big_or_unmeasured?(rows, limit) when is_integer(rows), do: rows > limit
 
   # Deleting rows is the only destructive thing V163 does, so it is never
   # silent — an operator reading the deploy log sees the count before the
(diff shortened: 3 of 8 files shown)


timujinne:
## Post-merge review — chain #689 → #694 → #695 → #696 (two-track ultra-review)

Two independent review tracks (3-pass A: broad → deepen → verify/narrow; 2-pass B: broad → deepen), each on the default model, then reconciled. Both tracks read files at chain HEAD `855c17f9` and re-derived the load-bearing claims against source. **No runtime execution** — no Elixir/PG available to either track, so closure claims rest on code-reading + Ecto source inspection.

**Headline:** the two tracks are strongly convergent, not contradictory. They agree on every cross-cutting verdict. One HIGH-severity finding raised in A1 (mentions "drops every @-ping") was **refuted** by both tracks against Ecto source. The one finding worth acting on before release is **MEDIUM, B-only**.

---

### ✅ Verified CLOSED — chain's headline claims hold

**#695 closes all six #694-round findings** (full A/B consensus, re-derived end-to-end):
1. `read_comment/2` nil/corrupt conflation → `probe.ex:94` returns `:unparseable`; `comment_policy.classify(:unparseable,...) -> :comment_unreadable`; `--adopt` cannot stamp over it (routes to error, never to `run_adopt_branch`).
2. `estimated_rows` collapsed `reltuples = -1` to `0` → `uuid_integrity` returns `:unknown`, guarded.
3. `status --verbose` crash on `{:unknown_version}` → handled.
4. V164 post-ADD probe raised mid-run → wrapped in `try/rescue`.
5. `comments_fk_on_delete` read by name alone → anchored on `contype='f'` + arity + column.
6. V163 deferral text → fixed.

No seventh finding dropped. #694's own BUG-HIGH items were fixed in the #694 pass (`common.ex:200`), correctly outside #695's scope.

**#696 closes both F-2 and F-7 from the #691 round, no same-series sibling left fail-open.** Both tracks audited every actor-gated write in `auth.ex`: `update_user_status/3`, `toggle_user_confirmation/2` now have the three-branch shape; `admin_update_user_password/3` already fail-closed; `delete_user/2` fail-closed by a different mechanism. `git grep "_system ->" auth.ex` → nothing. **No sibling left open.**

---

### 🔴 B-#1 · MEDIUM · Differ NOT-NULL masking defeats `--adopt`'s clean gate

`differ.ex:222` `reason_not_null(reasons, %{not_null: true, default: nil}, _observed)` short-circuits and never compares `_observed`. Three sharpenings Track B alone made:

1. **Blast radius is the dominant class, not a corner case:** 632 manifest shapes match `%{not_null:true, default:nil}`, 170 of them `type:"uuid"` — the V56/V57 flush-fallout columns V164 exists to repair.
2. **`--adopt` is defeated:** `floor_verify_clean?/1` (`comment_policy.ex:213`) cannot see the suppressed `:wrong_shape`, so a comment-less DB whose `*_uuid` PKs/FKs are nullable **passes the clean gate and gets the floor stamped over unconverged drift**. This raises it from "verify under-reports" to "the adopt contract can be silently violated."
3. **Masks a known manifest/migration disagreement:** `changed_by_uuid` is deliberately nullable in V164 (`ON DELETE SET NULL`) but the manifest says NOT NULL (`expected_schema.ex:29786`) — the exclusion hides the disagreement rather than surfacing it.

**Fix:** in `differ.ex:222` and/or a "just-created" gate in `repair.ex:587` `presence_present_finding`. The `--adopt` sharpening is the release-impacting part. Also resolve the `changed_by_uuid` manifest/migration disagreement so the differ stops hiding it.

---

### Next (low-effort hardening, batchable)

- **B-N1 · LOW** `postgres.ex:619 plan_up/3` + `:704,:721` parsers lack the `n > 0` guard that `Probe.read_comment/2` (`probe.ex:94`) enforces. A hand-edited comment `'-5'` → `{:run_delta, -4..166}` → `Module.concat([Postgres, "V-4"])` → opaque `UndefinedFunctionError` before any DDL. Status path is safe (`current_version > 0` guard). Notably `probe.ex:90-92` already documents this exact risk on the path it guards.
- **B-N2 / A-N6 · LOW** `mentions.ex:771-776` `claim_for_delivery`: drop `returning: false` *and* collapse the `case` to `{count, claimed} when is_list(claimed)` so the silent-`[]` trap cannot exist under a future Ecto precedence change. Correct today; latent.
- **A-N1 · LOW** `v164.ex:827`: literal `'user_uuid'` should be `'#{@comments_fk_column}'` for consistency with the rest of the query. Identical value today, no functional bug.
- **B-#2 · LOW** `v164.ex:654-655`: FK-retry summary says "re-running does not retry them" — accurate for FK constraints but the NOT-NULL path *does* retry. One-line scoping fix.

### Optional (cleanup / docs / test gaps)

- **A-N3** `access_requests.ex:114,118` `grant/2`/`deny/2` have no in-repo caller and no authz check; a host could wire `grant/2` without authz. `@doc` warning or authz check.
- **A-N5** `mentions.ex:657,759` `notify/2` + `claim_for_delivery/1` have no in-repo caller → the claim/deliver path is untested. Add a regression test so the `select: m.uuid` clause (the thing that makes it correct) can't be silently removed.
- **A-N2** `auth.ex:1305,1329` `admin_confirm_user/1`/`admin_unconfirm_user/1` are unguarded primitives (callers today are gated). Rename or `@doc`-warn.
- **B-#4** `mentions.ex:514-519` `escape_md` escapes only `\ [ ] \`` — `(`,`)`,`_`,`*`,`!` in operator-settable titles can malformed link rendering.
- **B-#5 · NEGLIGIBLE** `uuid_fk_columns.ex:397,401,409,419` docstrings say `V163`, consumer is `v164.ex`. `s/V163/V164/`.

### Watch (open by design — flag at release time)

- **pgcrypto non-public schema:** `verify/1` shows permanent spurious `:wrong_shape` on the uuid function; repair converges via the helper. Noisy, not blocking. Confirm operator runbook note exists.
- **`idx_publishing_posts_group_slug` predicate contradiction** (manifest revs 59/62 `predicate: nil` vs rev 164 `predicate: "(slug IS NOT NULL)"`): self-heals after V164; only affects pre-V164 DBs at version 59-163.
- **Stale `chain_hash` blocker:** 2 failures in `phoenix_kit_release_check_test.exs` since #692 — **predates this chain, not closed by it.** Track separately.

---

### Cross-cutting verdict

**Did #695 really close all #694-round findings? Yes — all six, high confidence.** Two findings #694 deliberately left open (pgcrypto, idx predicate) are open by design, both tracks agree. No finding was dropped or silently partial.

**The residual risk is operational, not logical:** B-#1 (Differ NOT-NULL masking) is a separate, pre-existing verify blindness that #695 did not address and was not asked to address — but it means that on `*_uuid` columns, `verify/1` (and therefore `--adopt`'s clean gate) can report convergence that hasn't happened. That does not contradict "#695 closed its six findings"; it is a distinct gap surfaced by this review for the release decision.

*Full artifact trace (prep, both tracks' iterations, synthesis) in internal review directory. Method: two-track ultra-review, both tracks on the default model, reconciled by a fresh synthesizer.*

PR #707, opened 2026-08-12
integrations: GitHub and Amazon Bedrock providers

timujinne:
Two new built-in integration providers, and richer validation notes for the AWS ones.

## GitHub (`github`, api_key)

A read-only token for GitHub API consumers — org-statistics dashboard widgets and the like. GraphQL requires authentication even for public data, so the least-privilege credential is a fine-grained PAT with public-repo read-only access. Validated with the generic Bearer GET against `/rate_limit`. Instructions walk through fine-grained token creation.

## Amazon Bedrock (`amazon_bedrock`, api_key)

Long-term Bedrock API keys (plain Bearer, no SigV4) plus a region field. Validation is a new `:amazon_bedrock` strategy: `ListFoundationModels` in the configured region — the cheapest call that proves the key, the region, and the `bedrock:CallWithBearerToken` permission at once. It is the same auth header the OpenAI-compatible runtime endpoint accepts, so a green check means the AI completions path can authenticate too.

Declares `:ai_completions` with the primary region as the default `base_url`; other regions override `base_url` per AI endpoint. Two field-tested caveats are spelled out in the card instructions:

- the key's IAM identity must allow `bedrock:CallWithBearerToken` — without it every Bearer call answers 403 even when invoke permissions are in place;
- the OpenAI-compatible endpoint (`…/openai/v1`) currently serves only the `openai.gpt-oss-*` models — Anthropic and the rest answer on the native Converse API with the same key.

## aws_ses Test enrichment

The Test verdict stays the retry-confirmed send-quota probe (its transient-signature logic is what keeps good keys from being called invalid). A passing check now also reports the account id and per-service permissions (SES/SQS/SNS) via `PhoenixKit.AWS.CredentialsVerifier` — the same sweep the Emails settings page runs — so "Test connection" says whose key it is and what it may do. The SES and Bedrock cards gain step-by-step instructions with console/docs links, and the SES one cross-references Admin → Settings → Emails as the home of the delivery pipeline.

Also: the connections table gets the `zebra` variant like every other admin list.

## Verified

Live on a host app: GitHub connection validates and feeds the dashboard widgets; Bedrock connection validates ("38 foundation models visible in eu-central-1") and completes end-to-end through phoenix_kit_ai 0.19.0 against `openai.gpt-oss-20b-1:0` (and through native Converse with the same key). Pre-commit gate (compile --warnings-as-errors, format, credo --strict, sobelow) green on both commits.

### DIFF
--- a/lib/phoenix_kit/integrations/integrations.ex
+++ b/lib/phoenix_kit/integrations/integrations.ex
@@ -1072,6 +1072,9 @@ defmodule PhoenixKit.Integrations do
   defp do_validate(%{auth_type: :api_key, validation: %{strategy: :brevo_api}}, data),
     do: Validators.brevo_api(data)
 
+  defp do_validate(%{auth_type: :api_key, validation: %{strategy: :amazon_bedrock}}, data),
+    do: Validators.amazon_bedrock(data)
+
   defp do_validate(%{auth_type: :bot_token, validation: %{strategy: :telegram}}, data),
     do: Validators.telegram(data)
 
--- a/lib/phoenix_kit/integrations/providers.ex
+++ b/lib/phoenix_kit/integrations/providers.ex
@@ -185,10 +185,12 @@ defmodule PhoenixKit.Integrations.Providers do
       deepseek(),
       xai(),
       elevenlabs(),
+      amazon_bedrock(),
       aws_ses(),
       smtp(),
       brevo_api(),
-      telegram()
+      telegram(),
+      github()
     ]
   end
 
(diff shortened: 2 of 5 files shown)


timujinne:
Two independent reviews ran against this branch (one with full repo access, one diff-only); `8bd5974a` addresses everything substantiated:

- **verdict/enrichment separation** — the CredentialsVerifier sweep now runs in its own probe after the send-quota verdict and is strictly additive; its failure/exit/timeout can no longer flip a passing key into an error (and `{:ok, nil}`, which `record_validation/3` couldn't match, is gone);
- **least-privilege keys** — the permissions note lists only granted management APIs, so a `ses:SendEmail`-only key no longer reads as "SES denied";
- **Bedrock 403 ≠ invalid credentials** — an authenticated key lacking `bedrock:CallWithBearerToken` is told exactly that;
- Bedrock probe wrapped in `Probe.run` with Req retries off; region guard accepts 2–4 letter prefixes (`eusc-de-east-1`) and trims pasted whitespace; the models note stops implying per-account grants; docstrings stop promising cross-region guarantees the static `base_url` cannot keep;
- pure helpers (`aws_note/3`, `bedrock_models_note/2`, `valid_aws_region?/1`) are public and covered by new tests (36 green).

Known and deliberately left as-is:
- `check_permissions/3` still probes EC2 `DescribeRegions` whose result the note discards (one spare AWS call per manual Test click; trimming it means touching `PhoenixKit.AWS.CredentialsVerifier`'s public shape);
- the note now surfaces the AWS account id in `validation_status`/activity payloads — admin-only, non-credential, but new data in logs;
- `priv/gettext/*.pot` has ~700 lines of pre-existing extraction drift; ran `mix gettext.extract` locally and reverted rather than drown this diff — worth a separate catalog catch-up pass.

timujinne:
A third review pass (independent, diff-only) ran after the previous fixes; `ce14a769` addresses what it found:

- **partial permissions map** — `aws_note/3` is public for tests, so a caller handing it a map without one of the service keys would have hit `Map.values/1` on `nil`; missing services now simply read as "not granted";
- **trailing whitespace in the Bedrock key** — trimmed like the region already was, so a key pasted with a newline no longer reads as invalid credentials;
- **403 is ambiguous** — a key issued for a *different region* answers exactly like a missing IAM action, so the message now names the region and offers both explanations instead of sending the operator to edit IAM;
- **China partition** — Bedrock answers on `.amazonaws.com.cn` there; the global host never resolves, which surfaced as "could not reach Bedrock" for a perfectly good key.

Tests: 38 green (two new — partial-map note assembly, partition host). Full gate (compile `--warnings-as-errors`, docs, format, credo `--strict`, dialyzer) green.

Two suggestions from that pass were considered and not taken: the enrichment sweep's cost inside its own probe deadline (it is already strictly additive — a timeout degrades to "no note", never to a failed verdict), and splitting the model-catalog note's parenthetical into a separate gettext string (the single sentence reads better in the connection card).

PR #708, opened 2026-08-13
Add locale and priority options to the country selects

timujinne:
## Problem

`CountryData.countries_for_select/0` returns all 250 countries as English names in one flat alphabetical list. Two consequences for a non-English host:

- The dropdown is English on a fully translated page — an Estonian admin picks **Estonia**, not **Eesti**.
- Their own country is buried: `Estonia` sits between `Eritrea` and `Eswatini`, 60-odd rows down, and the countries they actually invoice are scattered across the list.

`beamlab_countries` has carried per-locale country names all along (`BeamLabCountries.Translations`); nothing in core used them.

## Change

`countries_for_select/1` and `eu_countries_for_select/1` take two options:

- **`:locale`** — country names in that locale, defaulting to the active `PhoenixKitWeb.Gettext` locale (dialects reduce to their base: `ru-RU` → `ru`). A locale with no translation data falls back to the English name **per country**, so a host only ever loses the translation, never the entry.
- **`:priority`** — alpha-2 codes pinned to the top in the order given, defaulting to `config :phoenix_kit, :country_select_priority`. A host that serves one region configures it once instead of patching call sites.

Sorting moves from the English name to the localized one, with accents folded through NFD — plain byte order exiles every accented name past `Z` (`Ühendkuningriik` after `Zimbabwe`), which reads as a bug in any locale that uses them.

```elixir
config :phoenix_kit, :country_select_priority, ~w(EE FI LV LT SE)

CountryData.countries_for_select(locale: "et") |> Enum.take(3)
# [{"🇪🇪 Eesti", "EE"}, {"🇫🇮 Soome", "FI"}, {"🇱🇻 Läti", "LV"}]
```

Both functions keep working with no arguments, so every existing caller — including `phoenix_kit_billing`'s `core_compat` MFA list, which probes `{CountryData, :countries_for_select, 0}` — is unaffected.

## Verification

New `test/phoenix_kit/utils/country_data_test.exs` (11 tests, the module had none): translation, dialect reduction, unknown-locale fallback, no country lost, pin order, no duplicate pinned entry, unknown/lower-case codes in `:priority`, the config default, and that the tail is sorted by the localized name.

```
$ mix test test/phoenix_kit/utils/country_data_test.exs
11 tests, 0 failures

$ mix format --check-formatted && mix credo --strict lib/phoenix_kit/utils/country_data.ex
(clean, no issues)
```

Also exercised end-to-end in a host app (Andi): the order form's country select renders `Eesti / Soome / Läti / Leedu / Rootsi` first, then the rest alphabetically, in `et`, `ru` and `en`.

## Notes

- No `doctest` was added: this module's existing examples are written against the bare `CountryData.` alias, which doctests cannot resolve. The new examples follow the same house style, so they are documentation only.
- Estonian names need `beamlab_countries` with an `et` locale (BeamLabEU/beamlab_countries#9). Without it, `et` simply falls back to English — this PR does not depend on it.
- `CHANGELOG.md` and `@version` are left to the maintainer.

### DIFF
--- a/lib/phoenix_kit/utils/country_data.ex
+++ b/lib/phoenix_kit/utils/country_data.ex
@@ -466,24 +466,34 @@ defmodule PhoenixKit.Utils.CountryData do
   def get_currency_code(_), do: nil
 
   @doc """
-  Get country name.
+  Get country name in the active locale.
+
+  Takes the same `:locale` option as `countries_for_select/1` and falls back
+  to the English name when that locale has no translation for the country.
 
   ## Examples
 
-      iex> CountryData.get_country_name("EE")
+      iex> CountryData.get_country_name("EE", locale: "en")
       "Estonia"
+
+      iex> CountryData.get_country_name("EE", locale: "ru")
+      "Эстония"
 
       iex> CountryData.get_country_name("XX")
       nil
   """
-  def get_country_name(country_code) when is_binary(country_code) do
+  def get_country_name(country_code, opts \\ [])
+
+  def get_country_name(country_code, opts) when is_binary(country_code) do
+    locale = opts |> Keyword.get(:locale, active_locale()) |> normalize_locale()
+
     case get_country(country_code) do
-      %{name: name} -> name
+      %{} = country -> translated_name(country, locale)
       _ -> nil
     end
   end
 
-  def get_country_name(_), do: nil
+  def get_country_name(_, _), do: nil
 
   @doc """
   Get country flag (emoji).
--- a/lib/phoenix_kit/utils/country_data.ex
+++ b/lib/phoenix_kit/utils/country_data.ex
@@ -289,17 +289,45 @@ defmodule PhoenixKit.Utils.CountryData do
   Returns list of tuples {display_name, alpha2_code} for use
   in Phoenix form selects, sorted by the country name in the active locale.
 
+  Names come from `BeamLabCountries.Translations`, not the country struct's
+  `:name` field — the two differ even in English, for 20 of the 250
+  countries (as of beamlab_countries 1.1.0). For example: GB "United
+  Kingdom of Great Britain and Northern Ireland" -> "United Kingdom", US
+  "United States of America" -> "United States", CZ "Czech Republic" ->
+  "Czechia", TR "Turkey" -> "Türkiye", KP "Korea (Democratic People's
+  Republic of)" -> "North Korea" (which also moves its place in the sorted
+  list, from the K's to the N's). A host that never passes `:locale` still
+  gets different strings, and a different order, than a version of this
+  function that read `.name` directly.
+
+  Sorting folds accented letters to their base form before comparing (e.g.
+  "ü" sorts with "u"), which is a deliberate approximation, not proper
+  collation — the BEAM has no ICU. It is correct for most Latin-script
+  locales, but wrong for locales that give diacritics their own place in
+  the alphabet: in Estonian, Ü belongs at the very end (after W, Õ, Ä, Ö),
+  and in Swedish, Å, Ä, Ö belong after Z; folding moves those names out of
+  that position instead of leaving them there. A host that needs strict
+  local collation should sort the returned list itself.
+
   ## Options
 
     * `:locale` — locale for the country names. Defaults to the active
-      `PhoenixKitWeb.Gettext` locale, reduced to its base code. Locales
-      `BeamLabCountries` ships no translations for fall back to English, so a
-      host only ever loses the translation, never the entry.
+      `PhoenixKitWeb.Gettext` locale, reduced to its base code (`"ru-RU"`
+      normalizes to `"ru"`). `BeamLabCountries` 1.1.0 ships translations for
+      `ar`, `de`, `en`, `es`, `fr`, `it`, `ja`, `ko`, `nl`, `pl`, `pt`, `ru`,
+      `sv`, `uk`, `zh` (`BeamLabCountries.Translations.supported_locales/0`);
+      any other locale falls back to the country's English name, and so
+      does an unsupported *value* such as an atom (`locale: :ru`) — a host
+      only ever loses the translation, never the entry.
 
     * `:priority` — alpha-2 codes pinned to the top of the list, in the order
       given; everything else follows alphabetically. Defaults to
       `config :phoenix_kit, :country_select_priority`, so a host that serves
       one region can put its own countries first without touching call sites.
+      A non-list value is treated as `[]`.
+
+  `opts` itself must be a keyword list — a map or a bare string raises
+  `FunctionClauseError` naming this function rather than `Keyword`.
 
   ## Examples
 
@@ -311,17 +339,13 @@ defmodule PhoenixKit.Utils.CountryData do
       ...> |> Enum.take(2)
       [{"🇪🇪 Эстония", "EE"}, {"🇫🇮 Финляндия", "FI"}]
   """
-  def countries_for_select(opts \\ []) do
+  def countries_for_select(opts \\ []) when is_list(opts) do
     locale = opts |> Keyword.get(:locale, active_locale()) |> normalize_locale()
     priority = opts |> Keyword.get(:priority, configured_priority()) |> normalize_priority()
 
-    entries = Enum.map(BeamLabCountries.all(), &select_entry(&1, locale))
+    {pinned, rest} = locale |> sorted_entries(:all) |> split_priority(priority)
 
-    {pinned, rest} = split_priority(entries, priority)
-
-    Enum.map(pinned ++ sort_by_name(rest), fn {_name, display_name, code} ->
-      {display_name, code}
-    end)
+    Enum.map(pinned ++ rest, fn {_name, display_name, code} -> {display_name, code} end)
   end
 
   @doc """
@@ -355,18 +379,16 @@ defmodule PhoenixKit.Utils.CountryData do
   Get list of EU countries for select dropdown.
 
   Takes the same `:locale` and `:priority` options as
-  `countries_for_select/1`.
+  `countries_for_select/1`, including its fallback, leniency, and sorting
+  caveats.
   """
-  def eu_countries_for_select(opts \\ []) do
+  def eu_countries_for_select(opts \\ []) when is_list(opts) do
     locale = opts |> Keyword.get(:locale, active_locale()) |> normalize_locale()
     priority = opts |> Keyword.get(:priority, configured_priority()) |> normalize_priority()
 
-    entries = Enum.map(eu_countries(), &select_entry(&1, locale))
-    {pinned, rest} = split_priority(entries, priority)
+    {pinned, rest} = locale |> sorted_entries(:eu) |> split_priority(priority)
 
-    Enum.map(pinned ++ sort_by_name(rest), fn {_name, display_name, code} ->
-      {display_name, code}
-    end)
+    Enum.map(pinned ++ rest, fn {_name, display_name, code} -> {display_name, code} end)
   end
 
   # {sortable_name, display_name, alpha2} for one country in `locale`.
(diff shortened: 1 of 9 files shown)


timujinne (commented):
## Review summary — two independent reviewers

Reviewed twice, independently, from the same brief: **Claude Opus** (adversarial, A/B against `upstream/main`, plus the live host app via Tidewave) and **GLM-5.2** (`elixir-review`). Both returned **NEEDS-WORK**. Everything they raised is now fixed in `ff6e8917`; the three commits were also reworded to start with `Add` / `Update` / `Fix` per AGENTS.md.

### The two findings that mattered

**1. The comment above `sort_by_name/1` had the collation backwards.** It claimed byte order "would exile every accented name past 'Z' — 'Ühendkuningriik' after 'Zimbabwe' — which reads as a bug". In Estonian, Ü *belongs* at the very end, after W, Õ, Ä, Ö, so that position is correct and the NFD folding is what breaks it: verified on the live app under `locale: "et"`, "Ühendkuningriik" lands at index **240**, inside the U block, between "Uus-Meremaa" and "Valgevene". Swedish is worse — folding puts "Åland" at index **14**, between "Azerbajdzjan" and "Bahamas", and "Österrike" right after "Oman", when Swedish collation puts Å Ä Ö after Z.

Folding stays — it is right for the majority of Latin-script locales and the BEAM ships no ICU — but it is now documented as the approximation it is, in both the comment and the `@doc`, naming the locales it gets wrong so a host needing strict collation knows to sort the result itself.

**2. The English output changed too, and that was undocumented.** `Translations.get_name(code, "en")` differs from the country struct's `:name` for **20 of 250** countries, so a monolingual host that never sets a locale gets different strings *and* a different order out of `countries_for_select/0`:

```
GB  "United Kingdom of Great Britain and Northern Ireland" -> "United Kingdom"
US  "United States of America"                             -> "United States"
CZ  "Czech Republic"                                       -> "Czechia"
TR  "Turkey"                                               -> "Türkiye"
KP  "Korea (Democratic People's Republic of)"              -> "North Korea"   # K -> N in the list
```

Now stated in the `@doc`, examples verified against the resolved dep rather than quoted from the review.

### Also fixed

- **Performance.** `countries_for_select/0` went from **0.902 ms** on `upstream/main` to **5.754 ms** on this branch (300 iterations, warmed) — in a function that runs in LiveView mount, twice per connection. The localized, sorted entry list is now memoized in `:persistent_term`, keyed by locale *and* source so the EU subset cannot collide with the full list: **1.266 ms → 0.018 ms** per warm call. Priority pinning stays outside the cache, so a `:country_select_priority` change still takes effect immediately — checked at runtime, not just by reading. Equivalence with the pre-cache implementation was verified across **144 combinations** (15 supported locales + `xx` + `ru-RU` + nil, × 4 priority lists, × both functions): all identical.
- **Option guards.** `countries_for_select(%{locale: "ru"})` and `get_country_name("EE", "ru")` raised `FunctionClauseError` from inside `Keyword.get/3`, pointing the caller at `Keyword`. `when is_list(opts)` now names this module. `get_country_name/2` needed more than a guard: its catch-all `def get_country_name(_, _), do: nil` would have swallowed the binary-code-plus-bad-opts case, so it is narrowed to `when not is_binary(country_code)`. `get_country_name("EE", "ru")` is the realistic slip here, since the option is `:locale`.
- **Two tests were doing nothing.** The sort test asserted `names == Enum.sort_by(names, <the implementation's own sort key>)` — it re-derived the key it was supposed to be testing and therefore could not fail for any collation, which is exactly why it did not catch finding 1. It now asserts literal neighbours (`Azerbaijan < Åland Islands < Bahamas`, plus two more pairs); dropping the fold in a scratch copy turns it red (`left: 249, right: 15` — "Åland Islands" falls to second-to-last). The "keeps every country" test compared lengths, 250 == 250 by construction of the dataset; it now compares the alpha-2 code sets.
- **Missing coverage.** Nothing exercised the zero-arity contract or the `active_locale/0` default branch. Added, with `Code.ensure_loaded?/1` before `function_exported?/3` — without it the probe answers `false` on a cold VM, which the existing arity test also got wrong.

### Verified and dismissed

- **`get_country_name/1`'s locale sensitivity is deliberate**, and its blast radius was checked rather than assumed: no caller in this repo at all; in `phoenix_kit_billing`, `order_form.ex:372` and `format_company_address/1` both only build strings for display. Nothing persists, compares or caches the value. The one real consequence — the same order printed from a web request and from an Oban job can name the country differently — follows from the feature and is now documented.
- **`code in pinned_codes` inside a reject** is O(n·m) as written but measured harmless: 250-code priority 2.86 ms, 1 code 4.58 ms, empty 4.08 ms (full priority is *fastest*, since it skips sorting). It is now moot anyway — the sort it fed is cached.
- **No country vanishes or duplicates** in any of the 16 supported locales: 250 entries, 250 unique codes, zero duplicate display names each.
- **Two tests do pass against a revert** — "still answers nil for an unknown country" and "keeps its one-argument shape". That is their purpose: they are backward-compat guards. Swapping in `upstream/main`'s implementation fails 13 of 17.

### One thing this PR cannot fix

`:locale` only translates for locales the *installed* `beamlab_countries` ships. The pinned 1.1.0 has 15 and **Estonian is not among them** — in a clean checkout of this branch, `get_country_name("GB", locale: "et")` still answers `"United Kingdom of Great Britain and Northern Ireland"`. Estonian works in our host app only because it overrides the dep with a local checkout carrying two unpublished commits — and that checkout is still stamped `@version "1.1.0"`, identical to the published release, so a `mix deps.get` that ever resolves to Hex would silently drop Estonian with no version change to signal it. The fix belongs in `beamlab_countries` (BeamLabEU/beamlab_countries#9) plus a release; noting it here so the dependency between the two is on record.

`mix format --check-formatted`, `mix credo --strict`, `mix compile --warnings-as-errors` and the module's 17 tests are all green. The full suite could not run in this container (no local PostgreSQL; `pk-test` dies with an environment fault unrelated to this diff), but a targeted `pk-test` over this test file and `test/integration/users/organization_test.exs` — the one in-repo caller — passed 40 tests, 0 failures.

timujinne:
### Second round — the admin setting (`12c9a2c2`, fixed in `b55f861d`)

The branch gained a settings-backed priority list after the first review, so that commit went through its own adversarial pass. Verdict was **NEEDS-WORK** with three real defects, all reproduced against the running app rather than argued. All three are fixed.

**The edit was silently discarded on an unrelated validation error.** `save_country_priority/1` ran inside `save_company_info/2`, which `handle_event("save_company", …)` only reaches once `validate_company_data/1` returns `[]`. A blank company name therefore dropped the operator's country list — and because the error branch skips `load_settings/1`, LiveView sent no diff and the typed text stayed on screen, so nothing signalled the loss. Worst on a fresh install, where it was circular: `@default_company_info` is all empty, so all five required fields fail, and the reason to set a priority list in the first place is to make finding your own country among 250 easier. The setting is independent of the company form and now persists on every submit.

**Total rejection was reported as success.** `"Estonia, USA, Suomi"` left nothing after `known_country_codes/1`, stored blank, silently reverted the dropdown to the config list — under a green "Organization information saved". Rejected codes are now named in the flash, and a wholesale rejection says so rather than implying a partial save.

**Pinning could not be turned off from the UI.** Blank means "fall back to config", so a host with `config :phoenix_kit, :country_select_priority` set could never disable pinning without a deploy — the one edit the setting exists to enable. This is not fixable by storing an empty string: verified against the running app that `Settings.get_setting_cached(key, default)` answers with the **default** for a stored empty string, exactly as for an absent key, so blank and unset are indistinguishable at the settings layer. Hence an explicit `none` sentinel (case-insensitive, trimmed), honoured over the config and round-tripped in the field so the operator can see that pinning is deliberately off. It is checked *before* `known_country_codes/1` — `"none"` is not a country code and would otherwise be dropped like any other unknown one, storing blank and leaving the switch unreachable.

Also fixed from the same pass:

- `Keyword.get(opts, :priority, configured_priority())` evaluated its default eagerly. That was near-free when the default was `Application.get_env/3`; with the setting behind it, it became a cache lookup — falling through to a query on a cold key — on every call, thrown away whenever the caller passed `:priority`. Both defaults now sit behind `Keyword.fetch/2`.
- The `rescue`/`catch :exit` around the settings read was **unreachable**: `get_setting_cached/2` rescues and catches `:exit` itself, and so does the `get_setting/2` it falls through to. Removed — a bare `rescue _ ->` with no log only swallows a future programmer error.
- `update_setting/2`'s `{:error, changeset}` was discarded. All 250 alpha-2 codes joined with `", "` is 998 characters against a 1000-character column, so one more country in the dataset would have turned this into a silent no-save under a success flash.
- The form read the value uncached while the hot path read it cached; and the `parse_priority/1` doc credited the wrong function with dropping unknown codes (`normalize_priority/1` only filters non-binaries, upcases and dedupes — `split_priority/2` drops them by matching no entry).

**Live verification** (setting restored to blank afterwards, config left at `EE FI LV LT SE`):

```
submit with blank company name + "LT, LV"  -> stored "LT, LV"       (was: not stored at all)
submit "Estonia, USA"                      -> :error flash naming ESTONIA, USA; nothing stored
submit " NoNe "                            -> stored "none"; dropdown starts at Afghanistan
```

30 tests in the module, 0 failures; `format`, `credo --strict`, `compile --warnings-as-errors` clean.

One gap worth stating rather than hiding: the LiveView save path itself has no automated test, because `Settings.update_setting/2` writes to the database and this module's suite runs without one. The pure halves (`parse_priority/1`, `known_country_codes/1`, `none_priority?/1`, and the config/setting/sentinel precedence) are covered; the wiring in `organization.ex` was verified by hand against the running app, as quoted above.

timujinne:
### Third round — the settings card (`55c38fd6`, `ca9042ce`, fixed in `f6a1d2ca`)

Reviewed twice again, independently: **Claude Opus** (adversarial, handlers called against the running app) and **GLM-5.2**. Both returned **NEEDS-WORK**; everything they raised is fixed.

### Two crashers, found by calling the handlers rather than reading them

`add_main_country` mapped `String.upcase/1` over a code taken straight from `phx-value`, so `%{"code" => 42}`, `%{"code" => ["EE"]}` and `%{"code" => %{}}` raised `FunctionClauseError` — a payload the browser never sends, but a WebSocket client trivially can. `reorder_main_countries` did the same through `Enum.filter/2` on a non-list `ordered_ids` (`"SE"`, `nil`). Both are guarded now, and `reorder`/`remove` gained the catch-all clause `add` already had: that asymmetry *was* the bug.

**Data integrity was never at risk, and both reviewers verified it rather than assuming.** A forged payload cannot add an unknown country (`"ZZ"` → no-op), cannot duplicate (500× `"EE"` → the stored five unchanged), cannot drop one (`ordered_ids: []` → list preserved), and cannot promote a code that was never pinned (`["US","SE"]` → `US` ignored).

### Drag was the only way to reorder

No keyboard path, and the `SortableGrid` hook fetches SortableJS from `cdn.jsdelivr.net`, so a strict CSP or an offline deploy leaves the operator with a list they cannot order at all — while add and remove keep working. The up/down buttons are back **alongside** the drag handle; drag stays primary. Their msgids were still in all four catalogues from the earlier iteration, so nothing new needed translating.

### The card only updated itself

- The Company card's own country `<select>` sits directly above it and was computed once in `load_settings/1` — add Estonia and the card reorders while the select above still starts at Afghanistan, until a reload.
- No `broadcast_settings_change/1`, although this LiveView subscribes to `"organization:settings"` at mount and every other card on the page broadcasts. Two admins on the page: A adds Estonia, B shows "Nothing chosen yet" indefinitely.

Both fixed on the write path, which now also **stops writing when nothing changed** — an empty Add, a remove of an absent code, a drag that ends where it started all previously rewrote the identical string.

### Smaller

- The suggestion went stale after switching the company country: `country_changed` updated `@company_country` but not the suggestion, which then silently flipped to the new country's neighbours at the next unrelated edit.
- `suggested_priority/2` raised on a non-integer `:limit` — `limit > 0` does not stop `"4"`, since Erlang term ordering puts a binary above any integer.
- Its docstring promised Estonia "Latvia, Finland, Lithuania and Sweden"; the data answers `["EE", "LV", "AX", "FI", "LT"]`. Corrected to match, Åland included.

### Withdrawn after verification

One reviewer asked for an upgrade note about removing `config :phoenix_kit, :country_select_priority`. The other checked `git log -S` and showed the key was **born and removed on this branch** — no released version ever read it, so no host can have set it and there is nothing to announce.

### The gap that let the crashers through

Nothing in `test/` referenced the LiveView. There are now **19 integration tests** against a real database covering the four handlers: forged payloads on every one, the reorder invariants (cannot add / drop / duplicate / promote), dedup and unknown-code rejection on add, the move clamps at both ends, the no-op writes, the Company-select refresh, and the cross-session broadcast (two mounted sessions, edit in A asserted in B).

### Verified independently after the fixes

```
mix test test/phoenix_kit/utils/country_data_test.exs            32 tests, 0 failures
pk-test test/integration/.../organization_main_countries_test.exs 19 tests, 0 failures
mix format --check-formatted / mix credo --strict / mix compile --warnings-as-errors   clean
```

Also checked on the running host app: `suggested_priority("EE")` → `["EE","LV","AX","FI","LT"]`, `("DE")` → `["DE","LU","NL","CZ","BE"]`, `("SG")` → `["SG","MY","ID","KH","BN"]`; `limit: nil | "4" | 2.5` all behave as the default instead of raising.

`CHANGELOG.md` and `@version` are left to the maintainer, per this workspace's convention — the release-note material is in the commit messages.

PR #714, opened 2026-08-14
Rename the built-in SEO module to Crawlers, and grow it into the bot-policy page

mdon:
## Why

Everything the built-in SEO module held was bot policy — the global
noindex/nofollow directive and robots.txt guidance — while its name and card
description promised meta tags, Open Graph and "search optimization" that
never lived there (Open Graph is its own module). Meanwhile actual SEO work
is arriving as the external `phoenix_kit_seo` package, whose natural
`module_key` the built-in was shadowing out of `ModuleRegistry.get_by_key/1`
lookups. The module is renamed to **Crawlers** and extended into the one
place an operator decides who may read the site.

## The rename

`PhoenixKit.Modules.SEO` → `PhoenixKit.Modules.Crawlers`, module key `seo` →
`crawlers`, settings `seo_module_enabled`/`seo_no_index` →
`crawlers_module_enabled`/`crawlers_no_index`, settings page at
`/admin/settings/crawlers`, permission key `crawlers`, layout assign
`:seo_no_index` → `:crawlers_no_index`. A `@deprecated` defdelegate shim
keeps `PhoenixKit.Modules.SEO` callers compiling (deliberately not a
`PhoenixKit.Module`, so it cannot re-enter the registry).

**V170** migrates stored data: settings rows are renamed with a dedupe guard
(the old row's value wins — it is what the running site was honoring), and
roles granted `seo` gain `crawlers`. The old `seo` permission rows are
**kept deliberately**: the ExpectedSchema manifest still lists the V135
`seo` seed as required, so deleting them would set `mix phoenix_kit.repair`
and this migration fighting over the same row. They are inert — no module
registers the key — and their retirement belongs to the next manifest
regeneration. (The migration was V167 during development and was renumbered
after `mdon/main` gained V167–V169 mid-flight; the marker walk-back
procedure in the migration docs covers databases that ran the pre-renumber
version.)

## New on top of the rename

- **Per-bot-group access controls** (`Crawlers.Bots`): search engines / AI
  training scrapers / AI assistants / SEO tool crawlers / web archives, all
  default-allowed. Each bot carries a robots.txt token and separately a
  request-UA fragment — `Google-Extended`/`Applebot-Extended` are
  robots.txt-only opt-outs crawling under the main search UA, so they are
  excluded from UA matching by construction.
- **Generated robots.txt**: the settings page renders the complete file the
  toggles describe, for copy-paste. Core still deliberately does not serve
  the file (host policy; `Plug.Static` answers first).
- **llms.txt** served at the site root and under the kit prefix (same
  reasoning as `/sitemap.xml`): heading from the project title, body from an
  operator-edited setting; 404 while the module is disabled.
- **Search-engine verification metas** (google-site-verification,
  msvalidate.01); pasting the whole `<meta>` tag extracts the content value.
- **`CrawlerMetas` head component** — self-contained settings reads, so it
  works in ANY root layout (most installs render the head in the host app's
  root layout, where core's assigns never arrive; the assign-based noindex
  meta silently reached no page on such installs). Core's two shells use it;
  hosts embed one line. The noindex directive renders whenever set — not
  gated on the module toggle — preserving the long-standing contract pinned
  by `auth_crawlers_no_index_test`.
- **`CrawlerBlocker` plug** — default-off, honestly advisory 403 for
  user-agents of blocked groups, wired into the kit page pipeline and
  documented for host pipelines. Policy-file scopes skip it on purpose.
- **Doctor check "Crawler Visibility"** — warns on noindex ON for a
  production-looking host and on a staging-looking host left indexable.
  Reads settings via direct SQL because the doctor runs in update_mode,
  which short-circuits `PhoenixKit.Settings` reads.

## Review round

An external multi-model review panel was run over the diff with instructions
to refute numbered claims; every finding was reproduced against the running
code before acting. Applied as a result: the three optional string settings
joined `@optional_settings` (first write of an empty value on a missing row
failed validation, half-applying the verification form — reproduced live),
the compat shim, hostname downcasing in the doctor, a narrower `down/1`
module-tag retag, and the noindex-contract fix above. Notable false
positives rejected with evidence: atom/string mismatch in `ua_fragments`
(blocked groups are always atoms), robots.txt stanza separation (present),
and a claimed test failure that predated the component (real — caught and
fixed as the noindex contract).

## Verified

- Full flow on max-dev2.don.ee: rename visible in Modules/settings/sidebar,
  V170 applied (after the documented marker walk-back from the pre-renumber
  167), group toggle → regenerated robots.txt, llms.txt 200/404 by module
  state, GPTBot 403 vs Googlebot 200 with enforcement on, policy files
  fetchable by blocked bots, verification meta in the served head, noindex
  meta + empty sitemap while the directive is on, doctor check reporting
  real state.
- `mix precommit` green (format, credo --strict, dialyzer, JS tests);
  targeted suite 23/0 including the prefixed-schema chain oracle over
  V135..V170; manifest `chain_hash` restamped (`verify.exs --scenario s7,s8`
  not run — the manifest body is unchanged apart from the hash).

## Follow-ups (deliberately out of scope)

- Manifest regeneration: add a `crawlers` permission seed and retire the
  `seo` seed (also removes the kept `seo` rows' one sharp edge: a host
  registering a custom `"seo"` key would find old grants attached).
- `mix phoenix_kit.update` backfill of `<CrawlerMetas.crawler_metas />` into
  existing host root layouts (the favicon precedent).
- `robots_txt_present?` checks `priv/static/robots.txt` relative to CWD —
  pre-existing behavior, wrong under releases; worth a release-aware path.
- llms.txt is `public, max-age=3600`, so a shared cache can serve it up to
  an hour after the module is disabled — same trade-off as the sitemap.

---

*This PR is opened from `mdon:main` and therefore also carries the fork's
pending work that preceded it on the branch: V167 (unique post slugs), V168,
V169, the slug-generation move into core, and the function-component
undeclared-read guard. The Crawlers work is the six commits from `536c1648`
through the `cc6ee34a` merge.*


ddon:
Conflicts with `main` in `lib/phoenix_kit/migrations/expected_schema.ex`.

Cause: this branch was cut before core **2.4.0** landed (PRs #711 + #712, published earlier today). 2.4.0 added V167/V168/V169 to the chain and restamped `@chain_hash` over the now-35-file `v*.ex` set, so any branch carrying its own `expected_schema.ex` edit diverges there.

Leaving the resolution to you rather than guessing at it — a manifest conflict is not one to settle from the outside, and `restamp_chain_hash.exs` deliberately refuses to be the answer when the change could move the schema.

Rebase on `main` (now at 2.4.0) and re-run `mix run dev_docs/squash/restamp_chain_hash.exs` to see whether a restamp is appropriate for your change, per that script's own rule. I'll pick this up on the next sweep.

mdon:
Conflict resolved — mergeable again.

While this PR was open, upstream took **V170** for the notification-collapsing
indexes (#713) and the fork then took **V171** for the shop-slug projections,
so the Crawlers migration renumbered twice and now lands as **V172**. Chain
order after the merge: V170 notifications (upstream) → V171 shop slugs →
V172 Crawlers. Both markers moved with it (up stamps '172', down '171'),
`@current_version` is 172, and the manifest chain_hash is restamped.

Verified: fresh chain into a clean database ends at marker 172 with all three
migrations applied; suite 3581 tests, 0 failures; `mix precommit` passes. A
dev install that had run the Crawlers content under its old number was
repaired by the documented marker walk-back and replay — V172's `up` is
re-runnable by construction, so the replay is a no-op over the applied shape.

PR #728, opened 2026-08-17
Fix integrations encryption: decouple key, add rotation, warn

timujinne:
## Summary

Integration credentials (`phoenix_kit_settings`, `module = "integrations"`) are encrypted with a key derived from `secret_key_base` when no dedicated key is configured. That's a silent, undocumented fallback with no operator-facing signal — `secret_key_base` is a session/CSRF secret, shared far more broadly than an encryption-at-rest key should be, and the installer never stamps a dedicated `integrations_encryption_key`, so this is the common case, not an edge case.

This PR:

- Adds a proper `key_status/0` (`:dedicated`, `:legacy_secret_key_base`, `:disabled_no_key`, `:disabled_explicit`) and surfaces it as a visible banner on the Integrations settings page, plus a one-shot boot-time warning (`PhoenixKit.boot/1`, so it runs after the host's Endpoint has started).
- Adds `mix phoenix_kit.integrations.rotate_key` to re-encrypt every stored connection under a new key — covers both genuine key rotation and first-time adoption of a dedicated key on an install whose credentials are currently plaintext or legacy-keyed.
- A configured dedicated key under 20 characters is rejected (with a distinct warning, not silently treated as "no key configured").

## Review

Independently reviewed end-to-end against a real database across five rounds (two genuine defects found and fixed along the way — a write-path bug that could permanently erase a credential that failed to decrypt once, and a regression in an intermediate round that caused the rotation task to skip plaintext-only rows). Full review text in the first comment below.

## Test plan

- `mix test` — full suite, integration tags included: 3799 tests, 0 failures.
- `mix format --check-formatted`, `mix credo --strict`: clean.
- `mix dialyzer`: clean (run standalone/unpiped — this container's pre-commit hook OOM-kills a chained `mix quality` run under its 4GB memory limit; several commits in this branch's history were made with `--no-verify` for that reason, each with the reasoning in the commit message, after verifying format/credo/dialyzer individually).


timujinne:
## Independent end-to-end review — PASS

Reviewed across five rounds against a real PostgreSQL database (not just the diff). Two genuine defects were found and fixed along the way; both are described below alongside what was verified.

### Round 4 findings (both fixed, verified fixed in round 5)

**1. Write path could permanently erase a credential (`encryption.ex`).** A field that failed to decrypt once was dropped from the in-memory map to keep it from leaking as a usable value to readers — but that same map fed every write path, including automatic ones (OAuth token-refresh), and `save_integration/2` writes `value_json` wholesale. So the very next ordinary write — no operator action required — permanently deleted the ciphertext. Before this fix, the same situation was recoverable once the correct key came back; after, it was not.

Fixed with a precise, non-heuristic approach: `Encryption.decrypt_fields_with_failures/1` returns `{safe_map, failed_field_names}`; every merge-based write path threads that list through and restores the untouched ciphertext only into what gets persisted, never into what's returned to a caller. `disconnect/3` deliberately does not participate — it already expresses removal intent unambiguously via an allowlist, and there is no way to distinguish "absent because undecryptable" from "absent because disconnect means to remove it" without that explicit list.

Verified end-to-end:
```
stored ciphertext present:                     true
api_key still in DB after 3 unrelated writes:  true   (record_validation, rename_connection, save_setup)
returned map exposes ciphertext:                false
credential recovered once the correct key is restored: <original value>
disconnect still clears it:                     true
```

**2. The boot-time warning misfired on the most common install shape (`supervisor.ex`).** The check ran from a `PhoenixKit.Supervisor` child, which starts *before* the host's Endpoint in a stock Phoenix tree. `Endpoint.config/1` raised before the endpoint was up, was rescued to `nil`, and the check reported `:disabled_no_key` (a false "credentials are in plaintext" alarm) on any install relying on the endpoint-secret fallback — which, per the code's own comment, is most installs, since the installer never stamps a dedicated key.

Fixed by moving the check into `PhoenixKit.boot/1`, which runs after `Supervisor.start_link/2` returns — i.e. after every child of the host application, Endpoint included.

### Round 4 → 5: a regression in one of the accompanying minor fixes

An unrelated, non-blocking cleanup in the same round (making the rotation task's reported count exact) accidentally changed *which rows get rewritten*, not just the count — it filtered on whether the raw stored value already carried the `enc:v1:` prefix. That meant a row whose secret was still **plaintext** (exactly the "first adoption" scenario the rotation task exists for) was excluded from the write set entirely: `rotate/2` reported `{:ok, %{rotated: 0}}` and left the plaintext untouched.

Fixed in round 5 by testing presence against the *decrypted* view instead of the raw one. Verified:
```
before: rotate on a plaintext row → rotated: 0, value unchanged
after:  rotate on a plaintext row → rotated: 1, value is now ciphertext, decrypts back to the original under the new key
a row with no sensitive fields at all is still correctly skipped (no spurious UPDATE)
```

### Also verified, no issues
- Config-key decoupling itself: the dedicated key is tried first; the legacy `secret_key_base`-derived tier is a real, working fallback, not a silent bypass.
- Rotation's row-locking and dry-run behavior match their documentation.
- The settings-page banner renders correctly for all three non-`:dedicated` statuses.
- A dedicated key under 20 characters is rejected with a message that names *why* (distinct from "no key configured"), rather than silently falling through with a misleading warning.
- `mix format`, `mix credo --strict`, `mix dialyzer` all clean; full suite (43 doctests / 3799 tests, integration tags included) passes.

Two nitpicks remain open by agreement, neither blocking: per-field decrypt-failure logging fires on every render of a busy admin panel rather than being deduplicated, and `mix phoenix_kit.integrations.rotate_key --new-key=<secret>` puts the new key on argv (shell history / `ps` visibility) rather than accepting it via stdin or an env var.

PR #732, opened 2026-08-18
Encrypt Storage bucket credentials, add an Integrations-backed source

timujinne:
## Summary

`PhoenixKit.Modules.Storage.Bucket` stores `access_key_id`/`secret_access_key` for cloud providers (`s3`, `b2`, `r2`, `tigris`). The schema docstring and README both say these are encrypted at rest. Neither was: `changeset/2` just cast the fields, and nothing in `lib/modules/storage/` ever called any encryption. Create a bucket through this API and the secret lands in the database in plain text.

## What this does

- `secret_access_key` is now actually encrypted before it's persisted, reusing `PhoenixKit.Integrations.Encryption` rather than adding a second crypto implementation — two new single-value functions, `encrypt_value/1`/`decrypt_value/1`, wrap its existing AES-256-GCM machinery. `access_key_id` is intentionally left alone; it's an identifier, not a secret (same treatment `aws_ses`'s `access_key` already gets).
- Buckets can instead reference a `PhoenixKit.Integrations` connection by UUID (`integration_uuid`, bare, no FK — same pattern as `phoenix_kit_newsletters`' `send_profiles.integration_uuid`). A changeset with both `integration_uuid` and a direct key set is a validation error, not a silent pick-one — two live sources for the same secret is exactly the failure mode this PR closes elsewhere.
- Decryption happens once, at the one place credentials are actually used (`Providers.S3.resolve_credentials/1`), not in the general-purpose `get_bucket/1` accessor, which now returns the stored value as-is.
- Already-encrypted-at-rest is opportunistic, not a backfill migration: a legacy plaintext `secret_access_key` gets encrypted on the next save of *any* field on that bucket (via `get_field/2`, not `get_change/2`, so an unrelated edit still triggers it). A real SQL-level bulk migration was considered and rejected — reimplementing this app's KDF+AES-GCM outside Elixir risks producing ciphertext its own `decrypt_value/1` can't read, and the blast radius is narrow (only installs that already configured direct cloud credentials, not the common `local`-bucket case). Existing rows stay plaintext until then; this is stated in the schema docstring and README, not left implicit.
- `secret_access_key` is also encrypted-length-aware now: the column was `varchar(255)`, but `enc:v1:` + base64 runs ~1.6x plaintext length, so any real secret past ~158 characters raised a raw, unvalidated Postgres error. Widened to `text` in the same not-yet-shipped V175 (folded in rather than a new version, since nothing has installed V175 yet).
- Three previously-silent failure paths now log (never the secret value, only the bucket and the reason): a missing/undecryptable integration, a decrypt failure, and `encrypt_value/1` falling back to plaintext when no key is resolvable — the last one is the same defect this PR exists to fix, one layer down, and was silent before this change too.

## What we did **not** verify — said plainly, not left for you to find

- `verify.exs --scenario s7,s8` — the harness this project's own manifest conventions point to for proving hand-declared objects against a live database — reports `SKIP:needs-generated-manifest` on this checkout regardless of a database being available (confirmed via the offline `--check` harness). We do not believe this is something this PR broke; it appears to be a pre-existing gap. In its place, we manually replayed `V175.up/1` against a scratch schema cloned from the live pre-V175 table structure and diff'd the catalog against every field the manifest declares (type, position, nullability, default, and the partial index's definition byte-for-byte) — full detail in the review comment below — plus ran the `create:`/repair-idempotence check the same scenarios would perform. That is not a substitute for the real harness; flagging the gap so it isn't silently assumed closed.
- Dialyzer could not be run to completion in our environment (an unrelated container memory limit kills the analysis phase before it produces output); `mix format`, `mix credo --strict`, and `mix compile --warnings-as-errors` are all clean.
- The dedicated test database we used for the 38 DB-backed tests (`phoenix_kit_test_bucket_enc`) no longer exists in our environment at review time, so that count is reported, not independently re-confirmed by every reviewer — the live-schema catalog verification above covers the same ground a different way.

## About the `chain_hash` restamp commit

One commit here (`8a8ef07b`) only restamps `PhoenixKit.Migrations.ExpectedSchema.@chain_hash`. Please read this plainly, not as a formality: **`chain_hash` is a desync detector between the shipped `postgres/v*.ex` files and the hand-written manifest body, not an integrity guarantee.** It's computed over whatever those files currently contain — anyone who edits a migration file and reruns the same script gets a hash that matches their own edit. What actually catches a bad change is the same thing that catches any other bug in a PR: a human reading the diff. To reproduce independently rather than take our word for it, run (no write):

```
mix run dev_docs/squash/restamp_chain_hash.exs
```

against this branch — it should report up to date. The restamp is its own commit with nothing else in it specifically so this one line is easy to review in isolation from the actual content changes, which land in the commits before it.

## Test plan

- `mix test` (relevant suites, DB-backed where applicable): 55+ tests, 0 failures.
- `mix format --check-formatted`, `mix credo --strict`, `mix compile --warnings-as-errors`: clean.
- `mix phoenix_kit.release_check`: `Migration Version Sync` — PASS (current_version/0 == v175.ex, floor contiguity V135..V175, chain_hash matches). The two remaining FAILs (`Git Branch`, `Tag Collision`) are hex-release readiness checks, not applicable to a PR.


timujinne:
## Independent end-to-end review

Reviewed across multiple rounds against a real PostgreSQL database (not just the diff). Two genuine gaps were found and fixed along the way, described below alongside what was verified — and, explicitly, what was not.

### Encryption path — verified against live behavior

- `secret_access_key` is genuinely encrypted before persisting (AES-256-GCM via `PhoenixKit.Integrations.Encryption`, the same machinery `#728` uses — this PR adds `encrypt_value/1`/`decrypt_value/1`, single-value wrappers around the existing cipher, not a second implementation). `access_key_id` is deliberately left alone, consistent with how `aws_ses`'s `access_key` is already treated as an identifier, not a secret.
- Decryption happens at exactly one place, `Providers.S3.resolve_credentials/1` — confirmed by grepping the whole tree for readers of `secret_access_key`. `get_bucket/1` returns the stored (possibly encrypted) value as-is; it does not decrypt.
- Mutual exclusion between `integration_uuid` and direct keys is a real changeset error, not a silent pick-one: tested by submitting a changeset with both set.
- Legacy plaintext migrates opportunistically: exercised against a real database — an existing bucket with a plaintext `secret_access_key`, saved with an unrelated field changed, and the secret came back encrypted afterward. No bulk migration; this is stated plainly in the docstring and README, not left implicit.
- Column overflow: reproduced end-to-end against real PostgreSQL in an isolated scratch schema. `enc:v1:` + base64 runs ~1.6x plaintext length, so a plaintext secret past ~158 characters overflowed the original `varchar(255)` with an unvalidated Postgres `22001`. Verified 159/200/2000-char secrets all round-trip after the column was widened to `text`; the full `up` → `down` → `up` cycle was also verified (down deliberately does not narrow the column back — shrinking a widened column risks truncating data this migration never touched).
- Three previously-silent failure paths in `resolve_credentials/1` (missing/undecryptable integration, decrypt failure, credentials map missing the expected keys) now log the bucket and the reason — confirmed the ciphertext itself never appears in the log output.
- `encrypt_value/1` falling back to plaintext when no key is resolvable — the same defect this PR exists to fix, one layer down — now logs a warning; confirmed the warning does not contain the secret.
- `redact: true` on the schema field: confirmed `inspect/1` on a bucket no longer shows the `enc:v1:` value.
- The `test_connection` LiveView path: the underlying `Storage.build_probe_bucket/1` helper correctly threaded `integration_uuid`, but its only production caller (`bucket_form.ex`) initially didn't pass it through — the helper's own tests were green while the real admin-panel button stayed broken. Caught by an independent re-review pass and fixed; the params map now carries `name`/`integration_uuid`.

### The manifest — what convinced us the body is correct

`chain_hash` (see the dedicated note below) only proves the shipped migration files and the recorded hash agree with each other — it says nothing about whether the hand-declared manifest objects describe what `V175.up/1` actually builds. That second claim needed separate proof, and the repository's own pointer for it — `verify.exs --scenario s7,s8` — reports `SKIP:needs-generated-manifest` on this checkout even with a database available (confirmed via the offline `--check` harness alongside s9/s10/s13/s17/s19/s20). That appears to be a pre-existing gap in this project's own verification story, not something introduced here — flagging it rather than letting the manifest header's repeated citation of s7/s8 imply it ran.

In its place: cloned the real pre-V175 `phoenix_kit_buckets` structure into an isolated scratch schema, applied `V175.up/1` verbatim, and diffed the resulting catalog against every field the manifest declares — type, position, nullability, default, and the partial index's definition byte-for-byte (`indexdef` matched modulo the schema token). Then ran the s7 tamper-and-repair check by hand, restricted to V175: dropped each of the three objects in turn and re-ran the manifest's own `create:` string verbatim. All three repaired correctly — critically, `secret_access_key` came back as `text`, not `varchar(255)`, which is exactly the defect an incomplete manifest entry would have reproduced. Re-running each `create:` against a healthy object was a clean no-op (the s8 idempotence property).

This covers the V175 delta. The V135..V174 manifest body predates this PR and was out of scope.

### About the `chain_hash` restamp commit

One commit here only restamps `PhoenixKit.Migrations.ExpectedSchema.@chain_hash`. Read this plainly, not as a formality: **`chain_hash` is a desync detector between the shipped `postgres/v*.ex` files and the hand-written manifest body, not an integrity guarantee.** It's computed over whatever those files currently contain — anyone who edits a migration file and reruns the same script gets a hash that matches their own edit. What actually catches a bad change is the same thing that catches any other bug in a PR: a human reading the diff. To reproduce independently rather than take this on trust, run (no write):

```
mix run dev_docs/squash/restamp_chain_hash.exs
```

against this branch — it should report up to date. The restamp is its own commit with nothing else in it, specifically so that one line is trivial to review in isolation from the content changes, which land in the commits before it.

### What we did **not** verify

- Dialyzer could not be run to completion in our environment — an unrelated container memory limit kills the analysis phase before it produces output. `mix format`, `mix credo --strict` (11058 mods/funs, no issues), and `mix compile --warnings-as-errors` are all clean.
- `verify.exs --scenario s7,s8` proper (see above) — SKIP on this checkout regardless of database availability; the manual replay described above covers the same ground a different way, but is not the same harness.
- The dedicated test database used for the DB-backed test runs no longer exists in this environment at final-review time, so the exact passing-test count from earlier in the process is reported, not independently re-run by every reviewer at every step — the live-schema catalog verification above re-covers the encryption/column/index claims directly against Postgres regardless.

### Gates at HEAD

`mix format` clean · `mix credo --strict` 11058 mods/funs, no issues · `mix compile --warnings-as-errors` clean · non-integration suite: 58 tests, 0 failures · `mix phoenix_kit.release_check`: `Migration Version Sync` PASS (current_version/0 == v175.ex, floor contiguity V135..V175, chain_hash matches). The two remaining FAILs (`Git Branch`, `Tag Collision`) are hex-release readiness checks, correctly out of scope for a PR.

PR #734, opened 2026-08-18
Add canonical/hreflang to the stable auth pages

timujinne:
Auth pages in core carry no canonical/hreflang assigns at all. On a multi-domain, multi-language install that means several near-duplicate versions of each are served with nothing telling a search engine which is preferred.

This adds `PhoenixKitWeb.Users.AuthSEO` and wires it into three of them:

- `/users/register` (`Users.Registration`)
- `/users/log-in` (`Users.Login`)
- `/users/register/magic-link` (`Users.MagicLinkRegistrationRequest`)

**Scope — corrected from the first version of this description.** That version claimed these were "the only public LiveViews without canonical/hreflang". That was wrong, and I would rather say so than let it stand. Five more public auth routes have a stable, token-free path and qualify by exactly the same criterion:

| route | module |
|---|---|
| `/users/magic-link` | `Users.MagicLink` |
| `/users/qr-login` | `Users.QrLogin` |
| `/users/reset-password` | `Users.ForgotPassword` |
| `/users/confirm` | `Users.ConfirmationInstructions` |
| `/users/referral` | `Users.ReferralGate` |

They are not covered here, and deliberately not added to this PR. The three above are the registration and sign-in entry points — the pages a multi-domain install actually competes with itself on. Whether the remaining five want SEO assigns at all is a separate question (several are functional forms that arguably should not be indexed in the first place), and it deserves its own change with its own reasoning rather than being folded into a reviewed one. Tracked on our side.

**Deliberately excluded, and these should stay excluded:** every route carrying a one-time token — `/users/confirm/:token`, `/users/reset-password/:token`, `/users/register/complete/:token`, `/users/register/verify/:token`, `/users/magic-link/:token`, `/users/qr-login/scan/:token`, `/users/qr-login/finish/:token`. Canonical/hreflang there is meaningless (nothing to consolidate) and actively wrong (the advertised link dies with the token). Noted in the module docstring so it does not get "fixed" later.

Note that `/users/reset-password` and `/users/reset-password/:token` are different modules — `Users.ForgotPassword` (the request form, path-stable) and `Users.ResetPassword` (token-bearing). The first version of this description said "reset-password" was deliberately excluded, which read as covering both. Only the token-bearing one is.

**Design.** `AuthSEO` mirrors `PhoenixKitEcommerce.Web.SEOHelpers`'s `canonical_host_resolver` contract (`config :phoenix_kit, :canonical_host_resolver, {mod, fun}`), so a multi-domain host wires one resolver for both the shop and auth surfaces. Setting the assigns is enough — rendering `@canonical_url`/`@hreflang_links` into `<head>` is the host app's root layout, same as it already is for shop pages.

One deliberate divergence from that reference: `AuthSEO` does not copy `strip_language_prefix/2`, which splits the path on `/` and takes the first segment as the locale. It uses `Routes.path(url_path, locale: :none)` instead. A test surfaced why: when `url_prefix` is non-empty (e.g. `/phoenix_kit/en/...`) the locale is no longer the first segment and the split picks up the prefix. `phoenix_kit_ecommerce` carries that bug for installs with a non-empty prefix — out of scope here, but worth its own issue.

`hreflang_links` is `[]` when Languages is disabled or fewer than two languages are enabled; one language is not a hreflang set. `x-default` is not emitted here, matching `SEOHelpers` — hosts that want it add it in their own rendering layer.

**Test plan**
- `mix test test/phoenix_kit_web/users/auth_seo_test.exs test/phoenix_kit_web/users/auth_seo_mount_test.exs` — 12 new tests
- `mix test` — full suite, 43 doctests + 3858 tests, 0 failures
- `mix format --check-formatted`, `mix credo --strict`, `mix compile --warnings-as-errors --all-warnings` — clean
- Dialyzer not run: no PLT built in this clone. Say if you want it before merge.

### DIFF
--- a/lib/phoenix_kit_web/users/auth_seo.ex
+++ b/lib/phoenix_kit_web/users/auth_seo.ex
@@ -1,0 +1,113 @@
+defmodule PhoenixKitWeb.Users.AuthSEO do
+  @moduledoc """
+  Canonical / hreflang assigns for the auth LiveViews whose path is stable
+  and token-free (registration, log-in, magic-link registration request).
+
+  Mirrors `PhoenixKitEcommerce.Web.SEOHelpers`'s contract so a multi-domain
+  host wires one resolver for both the shop and auth surfaces: the canonical
+  host per language comes from the workspace-wide
+  `config :phoenix_kit, :canonical_host_resolver, {mod, fun}` MFA (nil or
+  absent ⇒ the site-wide `PhoenixKit.Utils.Routes.base_url/0`, today's
+  behavior). On a language's own resolved host the locale segment is dropped
+  (`Routes.path(url_path, locale: :none)`) — the domain itself carries the
+  language, so a `/en` (or any other) prefix on top of it would be redundant.
+
+  Deliberately excludes every token-bearing auth page (email confirmation,
+  password reset, magic-link verify, QR-login confirm): a one-time token in
+  the URL makes canonical/hreflang meaningless — there is no duplicate-content
+  signal to consolidate — and actively wrong, since the link a search engine
+  would be told about goes dead the moment the token is spent.
+
+  Setting these assigns is enough on its own: the root layout that renders
+  `@canonical_url` / `@hreflang_links` into `<head>` is the HOST app's, per
+  the same cross-package contract `phoenix_kit_ecommerce`'s catalog pages
+  already rely on (see `PhoenixKitEcommerce.Web.SEOHelpers`). Nothing here
+  renders anything itself.
+  """
+
+  require Logger
+
+  alias PhoenixKit.Modules.Languages
+  alias PhoenixKit.Modules.Languages.DialectMapper
+  alias PhoenixKit.Utils.Routes
+
+  @doc """
+  Canonical + hreflang assigns for a stable, token-free auth path, e.g.
+  `"/users/register"`.
+
+  `hreflang_links` is `[]` when Languages is disabled or fewer than two
+  languages are enabled — one language is not a hreflang set.
+  """
+  @spec seo_assigns(String.t()) :: %{canonical_url: String.t(), hreflang_links: [map()]}
+  def seo_assigns(url_path) when is_binary(url_path) do
+    %{
+      canonical_url: canonical_absolute_url(current_language(), url_path),
+      hreflang_links: hreflang_links(url_path)
+    }
+  end
+
+  defp hreflang_links(url_path) do
+    links =
+      enabled_languages()
+      |> Enum.map(&DialectMapper.extract_base(&1.code))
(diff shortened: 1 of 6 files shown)


timujinne:
## Review — independent verification

Posting the full review rather than a summary, per our process. Everything below was re-run and re-checked independently of the author, on a clean clone of this branch.

### Verified

| check | result |
|---|---|
| branch base | `6a98179e` on top of `upstream/main` (`fa42b5e6`), not on the fork's stale `main` |
| diff size | 6 files, +381/−2 — helper 113 lines, 3 call sites, 252 lines of tests |
| new tests | `12 tests, 0 failures` (re-run) |
| full suite | `43 doctests, 3858 tests, 0 failures` in 106.9s, exit 0 (re-run) |
| format / credo / compile | clean, `--warnings-as-errors --all-warnings` |

First re-run of the new tests failed to compile with `PhoenixKitWeb.ConnCase is not loaded`. That was an environment artefact on our side — `MIX_ENV=dev` was set, so `test/support` was not compiled. With `MIX_ENV=test` the numbers match the author's report exactly. Recording it so the next person does not read it as flakiness.

### Deliberate divergence from the reference implementation

`AuthSEO` is modelled on `PhoenixKitEcommerce.Web.SEOHelpers`, but it does **not** copy that module's `strip_language_prefix/2`, which splits the path string on `/` and takes the first segment as the locale. It uses `Routes.path(url_path, locale: :none)` instead.

This is not a style preference — a test surfaced the reason. When `url_prefix` is non-empty (e.g. `/phoenix_kit/en/...`), the locale is no longer the first segment and the split-based approach picks up the prefix instead. The `Routes.path/2` form is correct in both cases.

**Worth flagging separately:** this means `phoenix_kit_ecommerce` carries a latent bug of the same class for installs that use a non-empty `url_prefix`. It is out of scope here and not touched, but it is real and probably wants its own issue.

The rescue path also handles `:exit` in addition to raises, which the reference does not; both fall back to the site base URL and log a warning, following `Routes.get_default_language_base/0`.

### Scope

Included: the three auth pages with a stable, token-free path (`/users/register`, `/users/log-in`, `/users/register/magic-link`).

Excluded on purpose: confirmation, reset-password, magic-link verify, QR-login confirm. Each carries a one-time token in the URL, so there is nothing to consolidate and the advertised link dies with the token. This is stated in the module docstring so it does not get "fixed" later.

Not changed: the core root layout does not render these assigns — it does not render the ecommerce ones for shop pages either, so this is consistent, not an omission. Host apps with their own layout already read the pair.

### Not done, named explicitly

Dialyzer was not run — no PLT is built in this clone, and a first build is disproportionate for a change of this size. If the maintainer wants it before merge, say so and it will be run separately.

PR #737, opened 2026-08-19
Add V177: catalogue item-to-attribute-set attachments table

mdon:
One migration: `phoenix_kit_cat_item_attribute_sets` — the join table behind the catalogue's attribute-sets rework (catalogue PR: BeamLabEU/phoenix_kit_catalogue#74).

- Composite PK `(item_uuid, set_uuid)`, `position` for per-item ordering, reserved `data` JSONB for per-attachment state (currently `selected_value_slugs`, the per-item value selection).
- `item_uuid` FKs to catalogue items; `set_uuid` deliberately has **no FK** — it references `phoenix_kit_entities.uuid` across module boundaries. Integrity is owned by the modules instead: entities consults the catalogue's registered delete guard before deleting a set blueprint (fails closed), and the catalogue prunes orphans off entities PubSub events. Rationale in the migration moduledoc.

**Filed as V176; renumbered to V177** after upstream took V176 for the A002 FK-validation pass — resolved by merging upstream/main (2.13.1), keeping upstream's V176 exactly as shipped. Every statement is IF-NOT-EXISTS idempotent, so installs that ran the DDL under the old number re-run it as a clean no-op (verified live on a box that had the pre-renumber deploy: 176 → 177 migrated cleanly over the existing table). The expected-schema manifest gains hand-declared, catalog-exact V177 objects (the V175-buckets precedent) and a restamped chain hash; full suite (3907 tests), release_check, and fixture conformance green against the V177 chain.

Companion PRs: BeamLabEU/phoenix_kit_entities#31 (Managed blueprints + FieldInput) and BeamLabEU/phoenix_kit_catalogue#74 (the rework itself). Merge this first.

https://claude.ai/code/session_0164bhZcmjp9N98XXTtZbrxv


timujinne:
Thanks for the migration — the shape of the table looks right, and the reasoning in the moduledoc is unusually thorough. Two things need to change before this can land, and one of them is a hard blocker.

## Blocker: V176 is already taken on `main`

`v176.ex` in this PR is a `new file`, and `@current_version` moves 175 → 176. But `main` already carries a V176, merged before this branch was opened:

```
18.08 12:36  V176 merged into upstream/main  (FK validation for existing NOT VALID constraints)
18.08 14:55  main → 2.13.1, @current_version 176
19.08 01:24  this PR opened, adding a different V176
```

Merging as-is would put the catalogue table where the FK-validation migration lives. Git will flag the file collision, but that only helps if whoever resolves it notices the semantic clash — and the worse half is silent: databases where our V176 has **already been applied** (two of ours have) would keep a `schema_migrations` row saying V176 is done, while the code under that number is now something else entirely. Schema and migration history would drift apart without a single error.

Fix is mechanical: rebase on current `main` and renumber to **V177**. Nothing in the content needs to change.

## The delete guard this design leans on does not exist

The moduledoc justifies skipping the FK on `set_uuid` this way:

> the entities blueprint delete path consults the owning module's delete guard (`PhoenixKitEntities.Managed`)

We traced the entities side against a local `main` checkout:

- `grep -rn "Managed" lib/` and `git log --all -p -S "Managed"` both return nothing. `PhoenixKitEntities.Managed` does not exist in that codebase under any branch.
- The actual delete path is `PhoenixKitEntities.delete_entity/2`, which calls `repo().delete(entity)` and broadcasts on PubSub. Nothing can veto the delete, and a raw-SQL delete skips even the broadcast.

The argument against a hard FK is sound in principle — a cascade would take catalogue data with a blueprint, and a cross-module FK would tie entities' lifecycle to the catalogue. We are not asking for the FK. We are pointing out that the safety net named in the moduledoc is not there yet, so today the integrity story is "nothing stops it, and a catalogue-side PubSub cleanup may or may not exist" (we could not verify the catalogue side — that repo is not reachable from here).

Either land the guard alongside this, or reword the moduledoc to describe what actually protects the data today. The comment will outlive the memory of what was planned.

## Smaller notes

- **`position` has no uniqueness constraint** on `(item_uuid, position)`. Concurrent attaches can produce duplicate positions. This matches the precedent in V173's sibling tables, so it is consistent rather than wrong — flagging it as a known gap, not a change request.
- **`data` JSONB** — the diff reads nothing from it, which matches the "designed, not yet read" claim.
- **Down migration** is present and correctly drops the fresh table. It is missing the "lossy rollback" doc comment that V173 carries; worth adding for symmetry.

PR #739, opened 2026-08-19
Fix Ecto.CastError on OAuth registration with geolocation tracking enabled

timujinne:
## Summary
- `register_oauth_user/3` builds `attrs` with atom keys (`:email`, `:password`, `:first_name`, `:last_name`) for the OAuth registration path.
- With geolocation tracking enabled, `register_user_with_geolocation/2` merged its own string-keyed fields (`"registration_ip"`, `"registration_country"`, ...) into that map without normalizing key types first, producing a mixed atom/string-keyed map.
- `Ecto.Changeset.cast/3` raises `Ecto.CastError` on a mixed-key map, so **every** OAuth sign-in/registration crashed on any host with geolocation tracking enabled and a non-localhost client IP.
- Fix: normalize `attrs` to string keys at the top of `register_user_with_geolocation/2` before merging in the IP/geolocation fields, so the merge is always homogeneous regardless of the caller's key style.

## Test plan
- [x] Added a regression test reproducing the exact production crash (atom-keyed OAuth attrs + `track_geolocation: true` + a real IP), using `127.0.0.1` to stay network-free (hits `Geolocation.lookup_location/1`'s early-return branch).
- [x] `pk-test test/integration/users/oauth_email_verification_test.exs` — 16 tests, 0 failures.
- [x] `pk-test test/integration/users/` (full users integration suite) — 390 tests, 0 failures.
- [x] Verified live via Tidewave against a running host app pinned to `phoenix_kit 2.13.1` (same code path) — reproduced the crash before the fix, confirmed `{:ok, %User{}}` after.

### DIFF
--- a/lib/phoenix_kit/users/auth.ex
+++ b/lib/phoenix_kit/users/auth.ex
@@ -554,10 +554,7 @@ defmodule PhoenixKit.Users.Auth do
   end
 
   defp stringify_keys(attrs) do
-    Map.new(attrs, fn
-      {key, value} when is_atom(key) -> {Atom.to_string(key), value}
-      {key, value} -> {key, value}
-    end)
+    Map.new(attrs, fn {key, value} -> {to_string(key), value} end)
   end
 
   @doc """
--- a/lib/phoenix_kit/users/auth.ex
+++ b/lib/phoenix_kit/users/auth.ex
@@ -515,6 +515,12 @@ defmodule PhoenixKit.Users.Auth do
       {:error, %Ecto.Changeset{}}
   """
   def register_user_with_geolocation(attrs, ip_address) when is_binary(ip_address) do
+    # `attrs` may arrive atom-keyed (e.g. the OAuth registration path) or
+    # string-keyed (e.g. form params). Normalize to string keys before
+    # merging in any of our own string-keyed fields below — Ecto.Changeset.cast/3
+    # raises Ecto.CastError on a map with mixed atom/string keys.
+    attrs = stringify_keys(attrs)
+
     # Start with the IP address
     enhanced_attrs = Map.put(attrs, "registration_ip", ip_address)
 
@@ -545,6 +551,13 @@ defmodule PhoenixKit.Users.Auth do
   def register_user_with_geolocation(attrs, _invalid_ip) do
     # Invalid IP provided, register without geolocation data
     register_user(attrs, nil)
+  end
+
+  defp stringify_keys(attrs) do
+    Map.new(attrs, fn
+      {key, value} when is_atom(key) -> {Atom.to_string(key), value}
+      {key, value} -> {key, value}
+    end)
   end
 
   @doc """
--- a/test/integration/users/oauth_email_verification_test.exs
+++ b/test/integration/users/oauth_email_verification_test.exs
@@ -246,6 +246,27 @@ defmodule PhoenixKit.Integration.Users.OAuthEmailVerificationTest do
       assert created.confirmed_at
       assert confirm_tokens_for(created) == 0
     end
+
+    test "is created when geolocation tracking is enabled, even though the OAuth attrs arrive atom-keyed" do
+      # `register_oauth_user/3` builds `attrs` with ATOM keys (:email,
+      # :password, ...). `Auth.register_user_with_geolocation/2` used to
+      # `Map.put(attrs, "registration_ip", ip)` — a STRING key — producing a
+      # mixed atom/string-keyed map that `Ecto.Changeset.cast/3` refuses
+      # outright with `Ecto.CastError`. This is the real production path:
+      # `handle_oauth_callback/2` runs with `track_geolocation: true` and a
+      # real IP whenever the host app turns geolocation tracking on.
+      #
+      # "127.0.0.1" hits `Geolocation.lookup_location/1`'s early-return
+      # branch, so this stays network-free while still exercising the exact
+      # mixed-key merge that crashed in production.
+      email = unique_email()
+
+      assert {:ok, created, :created} =
+               OAuth.find_or_create_user(oauth_data(email), true, "127.0.0.1")
+
+      assert created.email == email
+      assert created.registration_ip == "127.0.0.1"
+    end
   end
 
   defp confirm_tokens_for(user) do


lib/phoenix_kit/users/auth.ex
```diff
@@ -515,6 +515,12 @@ defmodule PhoenixKit.Users.Auth do
       {:error, %Ecto.Changeset{}}
   """
   def register_user_with_geolocation(attrs, ip_address) when is_binary(ip_address) do
+    # `attrs` may arrive atom-keyed (e.g. the OAuth registration path) or
+    # string-keyed (e.g. form params). Normalize to string keys before
+    # merging in any of our own string-keyed fields below — Ecto.Changeset.cast/3
+    # raises Ecto.CastError on a map with mixed atom/string keys.
+    attrs = stringify_keys(attrs)
```
timujinne, line 522: **Altitude:** the normalization is applied only at this one merge site inside `register_user_with_geolocation/2`, not at the shared `register_user/2` → `do_register_user/1` choke point that *every* registration path (including this one) funnels through before `User.registration_changeset/3`'s `cast/3`.

Because the invariant "`attrs` is string-keyed before `cast/3`" is enforced here rather than at the shared entry point, a future `Map.put(attrs, "some_field", val)` added anywhere else in the registration pipeline (or a new caller merging string keys onto atom-keyed attrs) can reintroduce the exact same `Ecto.CastError` class this PR just fixed. Consider normalizing once in `do_register_user/1` right before the changeset call instead, so the whole pipeline is robust regardless of which merge site grows next.

lib/phoenix_kit/users/auth.ex
```diff
@@ -547,6 +553,13 @@ defmodule PhoenixKit.Users.Auth do
     register_user(attrs, nil)
   end
 
+  defp stringify_keys(attrs) do
```
timujinne, line 556: **Reuse:** this `stringify_keys/1` duplicates three already-existing, nearly identical private helpers elsewhere in this codebase instead of extracting a shared utility:

- `lib/modules/languages/languages.ex:1420` — byte-for-byte the same two-clause `is_atom(key)` guard implementation.
- `lib/phoenix_kit/notifications/notifications.ex:596` and `lib/phoenix_kit/theme_config.ex:700` — same normalization, implemented with `to_string/1`.

No shared `PhoenixKit.Utils.Map` (or similar) module exists to hold one canonical version — this is now the 4th copy of essentially the same 5-line function. A future improvement to the normalization logic (e.g. handling nested maps, or the collision edge case below) has to be found and applied in 4 places instead of 1.

lib/phoenix_kit/users/auth.ex
```diff
@@ -547,6 +553,13 @@ defmodule PhoenixKit.Users.Auth do
...
+    Map.new(attrs, fn
+      {key, value} when is_atom(key) -> {Atom.to_string(key), value}
```
timujinne: **Simplification:** the `is_atom(key)` guard branch is unnecessary — `to_string/1` already handles atom and binary keys identically, and it's what the codebase's own analogous helper (`notifications.ex:596`) uses:

```elixir
defp stringify_keys(attrs) do
  Map.new(attrs, fn {k, v} -> {to_string(k), v} end)
end
```

Same output for every reachable input here, one fewer clause to read.

lib/phoenix_kit/users/auth.ex
timujinne, line 522: **Efficiency:** `stringify_keys(attrs)` runs unconditionally on every call, including from the two non-OAuth callers whose `attrs` is always already string-keyed — a full `Map.new` rebuild that changes nothing on the more common registration path.

`lib/phoenix_kit_web/users/registration.ex:185` and `lib/phoenix_kit/users/magic_link_registration.ex:183` both pass LiveView/controller form params (always string-keyed) through `register_user_with_geolocation/2`; every such registration now pays for a copy that is a guaranteed no-op. Low impact since registration is infrequent, but worth noting — e.g. skip the copy when `Enum.all?(Map.keys(attrs), &is_binary/1)`, or restrict the normalization to the OAuth call path where the mixed-key risk actually originates.

lib/phoenix_kit/users/auth.ex
```diff
@@ -515,6 +515,12 @@ defmodule PhoenixKit.Users.Auth do
       {:error, %Ecto.Changeset{}}
   """
   def register_user_with_geolocation(attrs, ip_address) when is_binary(ip_address) do
```
timujinne, line 517: **Conventions:** this repo's CLAUDE.md/AGENTS.md states "Commit messages: start with Add, Update, Fix, Remove, Merge." Commit `ca1a16d9` is titled `fix: normalize attrs keys before geolocation merge in register_user_with_geolocation/2` — lowercase conventional-commit-style `fix:`, not `Fix ...`. Minor process nit, not a code issue (anchored here for lack of a better line).

lib/phoenix_kit/users/auth.ex
timujinne: **Latent robustness gap (low severity, not currently reachable):** `Map.new/2`'s last-write-wins collision behavior here means that if `attrs` ever carried *both* an atom and a string version of the same key (e.g. `:email` and `"email"` as separate entries), one would be silently dropped depending on the map's hash-based enumeration order — non-deterministic across inputs/OTP versions.

No current caller can produce this: the OAuth path builds `attrs` fresh with only atom keys, and the other two callers pass plain string-keyed form params. Flagging only as a documentation/robustness note in case a future caller merges attrs from two sources before calling this.

timujinne:
Pushed a follow-up commit (e38ac38e) addressing the review's simplification finding: `stringify_keys/1` now just does `Map.new(attrs, fn {k, v} -> {to_string(k), v} end)` — the `is_atom` guard branch was redundant since `to_string/1` on a binary is already a no-op.

On the rest of the review's findings:
- **Commit message convention**: noted, but since this repo squash-merges PRs and the PR title (`Fix Ecto.CastError on OAuth registration with geolocation tracking enabled`) already opens with `Fix`, the individual commit message won't survive to `main` — not worth a history rewrite for.
- **Reuse (4th `stringify_keys` duplicate)**, **altitude (choke-point placement)**, and **efficiency (unconditional normalization)**: agreed these are real, but fixing them means touching call sites/behavior beyond this bug fix's blast radius (a shared `Map.new`-style helper module, or moving normalization into `register_user/2` itself, which every registration path funnels through). Would rather do that as a separate, focused follow-up PR than widen this one.
- **Map.new/2 last-write-wins on colliding atom+string keys**: not reachable from any current caller, agreed it's documentation-only.

Re-verified live against a running host app pinned to `phoenix_kit 2.13.1` (this repo's `main`-equivalent release) after both commits — reproduces cleanly before, `{:ok, %User{}}` after.

PR #740, opened 2026-08-19
Add crawlers_sitemap_exempt_from_no_index: decouple sitemap content from the noindex directive

timujinne:
## Why

`crawlers_no_index` currently gates the sitemap's content and the robots
meta directive together: while noindex is active, `/sitemap.xml` (and
every per-module file) collapses to an empty `<urlset>`, unconditionally.

That makes the sitemap feed impossible to verify on an install that is
intentionally marked noindex (a pre-prod stand, say) without briefly
lifting `crawlers_no_index` for the whole site. We hit this directly: the
only way to check the sitemap on a closed pre-prod stand was to disable
`crawlers_no_index` site-wide for a window — 37 seconds, in our case — and
for the duration of that window a stand carrying 627 real product URLs was
completely undirected: no noindex meta, no empty-sitemap guard, nothing.
For those 37 seconds the stand looked exactly like a production site to
anything crawling it in that instant.

This PR adds a second, independent flag so that state is reachable without
the tradeoff: **directives stay up, sitemap stays full.**

## What changes

- `PhoenixKit.Modules.Crawlers.sitemap_exempt_from_no_index?/0` (default
  `false`) + `update_sitemap_exempt_from_no_index/1` — same shape as the
  existing `no_index_enabled?/0`/`module_enabled?/0` pair: a
  `Settings.get_boolean_setting/2` reader, default via the second
  argument, same module.
- `Generator.do_generate_all/2`'s noindex branch is now gated
  `crawlers_no_index?() and not sitemap_exempt_from_no_index?()` — the new
  flag can only *narrow* that branch, never fire it on its own, so it is
  backward compatible by construction rather than by claim: an install
  that never sets the new flag gets bit-for-bit today's behavior.
- A checkbox on `/admin/settings/crawlers`, in the Robots Directive card
  below the noindex toggle.

## Boundary of applicability

**This flag does not lift noindex.** It does not touch the robots meta
tag, and nothing in this change affects response headers. An operator who
enables it is still fully closed to indexing — the *only* thing that
changes is that the sitemap stops publishing an empty `<urlset>` and goes
back to listing real URLs. Search engines are still told `noindex` on
every page; they just also see a populated sitemap if they go looking for
one.

## Scope

Scoped to `Generator.generate_all/1`'s dispatch (`do_generate_all/2`).
`generate_html/1`'s own noindex short-circuit (the HTML sitemap view) is
deliberately untouched — separate call site, not covered by this change.

## Test plan

- `test/integration/sitemap/no_index_test.exs` — four states: flag
  off/on crossed with noindex off/on. The flag-on + noindex-on case is
  the one this PR exists for (sitemap stays full despite noindex).
- `test/integration/phoenix_kit_web/live/settings/crawlers_sitemap_exemption_test.exs`
  — the new checkbox: default state, toggling both ways, persistence
  across mount, and that toggling it never flips `no_index_enabled?/0`.

Full verbatim run output and what was/wasn't verified by mutation testing
are in the review comment below, not restated here.

### DIFF
--- a/lib/modules/crawlers/crawlers.ex
+++ b/lib/modules/crawlers/crawlers.ex
@@ -33,6 +33,7 @@ defmodule PhoenixKit.Modules.Crawlers do
 
   @module_enabled_key "crawlers_module_enabled"
   @no_index_key "crawlers_no_index"
+  @sitemap_exempt_from_no_index_key "crawlers_sitemap_exempt_from_no_index"
   @block_at_app_key "crawlers_block_at_app"
   @google_verification_key "crawlers_google_verification"
   @bing_verification_key "crawlers_bing_verification"
@@ -98,6 +99,38 @@ defmodule PhoenixKit.Modules.Crawlers do
     # The directive flips what the sitemap must advertise. Drop the cached
     # sitemap and regenerate so `/sitemap.xml` reflects the new state instead of
     # serving a stale file. Best-effort — never let it break the toggle.
+    refresh_sitemap()
+
+    result
+  end
+
+  @doc """
+  Whether the sitemap's *content* is exempt from `no_index_enabled?/0`.
+
+  `no_index_enabled?/0` still drives the robots meta directive unconditionally
+  — this flag only lets the sitemap keep publishing its real URLs while that
+  directive is active, e.g. to verify the feed independently of the indexing
+  block. Default `false`: an install that never sets this keeps today's
+  behavior, where the sitemap blanks in lockstep with `no_index_enabled?/0`.
+  """
+  def sitemap_exempt_from_no_index? do
+    Settings.get_boolean_setting(@sitemap_exempt_from_no_index_key, false)
+  end
+
+  @doc """
+  Sets whether the sitemap is exempt from `no_index_enabled?/0` (see
+  `sitemap_exempt_from_no_index?/0`).
+  """
+  def update_sitemap_exempt_from_no_index(exempt?) when is_boolean(exempt?) do
+    result =
+      Settings.update_boolean_setting_with_module(
+        @sitemap_exempt_from_no_index_key,
+        exempt?,
+        @module_name
+      )
+
+    # Same reasoning as update_no_index/1: this flips what the sitemap must
+    # advertise, so the cached/stale file must not outlive the new state.
     refresh_sitemap()
 
     result
--- a/lib/modules/sitemap/generator.ex
+++ b/lib/modules/sitemap/generator.ex
@@ -90,7 +90,7 @@ defmodule PhoenixKit.Modules.Sitemap.Generator do
     xsl_enabled = Keyword.get(opts, :xsl_enabled, true)
 
     cond do
-      crawlers_no_index?() ->
+      crawlers_no_index?() and not sitemap_exempt_from_no_index?() ->
         do_generate_no_index(xsl_style, xsl_enabled)
 
       Sitemap.flat_mode?() ->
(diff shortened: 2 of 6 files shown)


timujinne:
## Review notes (full, not a paraphrase)

### What was mutation-tested vs. what wasn't

**`generator.ex` / `crawlers.ex` (the actual gating logic) — mutation-tested.**
Procedure: wrote the four-state test first against the *unmodified* generator, confirmed
the "flag ON + noindex ON → full sitemap" case failed for the expected reason (sitemap
still empty), implemented the fix, confirmed green, then reverted *only* the fix
(`git stash push -- lib/modules/sitemap/generator.ex lib/modules/crawlers/crawlers.ex`,
keeping the new tests in place) and re-ran: same test failed the same way. Restored
with `git stash pop` and verified the restored files are byte-identical to what passed
green via `sha256sum -c` (both `OK`) — not just "tests pass again," but the exact same
file.

**Settings-page checkbox — RED→GREEN only, no mutation/checksum pass.** Lower stakes
(a UI wrapper around the already-mutation-tested reader/writer, not new gating logic),
so I stopped at: wrote the 5 tests first, confirmed all 5 failed against the
unmodified page (`5 tests, 5 failures`, element not found — checkbox didn't exist),
implemented, confirmed `5 tests, 0 failures`. I did not revert-and-reconfirm this half.
Flagging this explicitly rather than letting "all green" imply the same rigor was
applied to both halves.

**Also explicitly not covered by this PR:** `generate_html/1`'s own noindex
short-circuit (the HTML sitemap view) is a separate call site and is untouched —
it still blanks under noindex regardless of the new flag.

### Verbatim run output

RED, before the generator fix (correct failure — not an empty pass, not an error):
```
1) test generate_all/1 with crawlers_no_index active still generates a full sitemap when the sitemap is exempt (PhoenixKit.Integration.Sitemap.NoIndexTest)
     test/integration/sitemap/no_index_test.exs:47
     Assertion with >= failed
     code:  assert total >= 1
     left:  0
     right: 1
     stacktrace:
       test/integration/sitemap/no_index_test.exs:58: (test)
4 tests, 1 failure
```

GREEN, after the generator fix:
```
Running ExUnit with seed: 301000, max_cases: 32
....
Finished in 0.09 seconds (0.00s async, 0.09s sync)
4 tests, 0 failures
```

Mutation RED — fix reverted via `git stash`, same test file, same failure:
```
1) test generate_all/1 with crawlers_no_index active still generates a full sitemap when the sitemap is exempt (PhoenixKit.Integration.Sitemap.NoIndexTest)
     Assertion with >= failed
     code:  assert total >= 1
     left:  0
     right: 1
4 tests, 1 failure
```

Restore + checksum verification:
```
$ git stash pop
$ sha256sum -c postfix_checksums.txt
lib/modules/sitemap/generator.ex: OK
lib/modules/crawlers/crawlers.ex: OK
```

Settings-page checkbox, RED (before the checkbox existed):
```
Running ExUnit with seed: 398873 [pre-fix run]
5 tests, 5 failures
```
(all five failing on `element/2` finding no `input[phx-click='toggle_sitemap_exempt_from_no_index']` — correct failure mode, not a compile error)

Settings-page checkbox, GREEN:
```
Running ExUnit with seed: 398873, max_cases: 32
.....
Finished in 1.1 seconds (0.00s async, 1.1s sync)
5 tests, 0 failures
```

Full suite, current HEAD (`bd4c62b8`), fresh run immediately before opening this PR:
```
Finished in 107.0 seconds (43.7s async, 63.3s sync)
43 doctests, 3914 tests, 0 failures
```
(3909 baseline + 5 new checkbox tests; the 4 generator-level tests replace/extend the
pre-existing 2 in `no_index_test.exs`, net +2 there.)

Quality gates on every changed file, all clean:
```
$ mix format --check-formatted <changed files>
EXIT:0
$ MIX_ENV=test mix compile --warnings-as-errors --force
Generated phoenix_kit app   (no warnings)
$ mix credo --strict <changed files>
found no issues.
```

### Why the condition is backward-compatible by construction, not by claim

```elixir
crawlers_no_index?() and not sitemap_exempt_from_no_index?() ->
  do_generate_no_index(xsl_style, xsl_enabled)
```
The new flag only ever *narrows* this branch — it cannot make it fire on its own, since
it's ANDed against the pre-existing condition, not ORed or substituted. An install that
never sets `crawlers_sitemap_exempt_from_no_index` gets `not false == true` here, i.e.
the exact original single-condition branch. Both new getters default to `false` on a
lookup error too (mirroring the existing `crawlers_no_index?/0` defensive wrapper), so
an infra hiccup fails toward the pre-existing blank-sitemap behavior, never toward
silently publishing a full one.

PR #743, opened 2026-08-21
Federated party references for the catalogue (V178–V180)

mdon:
Three migrations that let a catalogue supplier or manufacturer BE a CRM party
instead of necessarily a row in the catalogue's own directory.

CRM owns party identity — "supplier" and "manufacturer" are roles on a company.
The catalogue's references were foreign keys into its local tables, so a party's
uuid physically could not be stored in them. Each migration replaces one of those
FKs with the `{uuid, source}` shape `phoenix_kit_cat_item_supplier_info` has used
since V149.

### V178 — cross-reference columns

`cat_manufacturers.crm_company_uuid`, plus partial unique indexes on
`crm_company_uuid WHERE NOT NULL` for both directory tables, so one party cannot
be claimed by two local rows.

### V179 — an item's manufacturer

Adds `manufacturer_source` (CHECK-backed) and `manufacturer_name_snapshot`, drops
`phoenix_kit_cat_items_manufacturer_uuid_fkey`. The snapshot is a TOMBSTONE, read
only when a reference resolves to nothing — never the display source.

### V180 — the manufacturer↔supplier graph, and one current supplier per pair

`phoenix_kit_cat_manufacturer_suppliers` carried hard FKs onto **both** local
tables, so with parties in CRM the graph simply stopped being built and
warehouse's manufacturer-based supplier resolution had nothing to read. Both
sides now carry a source and both FKs are dropped.

Riding along (rather than spending a version number of its own): the partial
unique index `..._current_pair_uniq` on
`(item_uuid, supplier_uuid) WHERE valid_to IS NULL`, so one item cannot list the
same supplier twice with two live prices. **Partial on purpose** — several rows
per pair are exactly what a price revision produces. Existing duplicates are
CLOSED rather than deleted (they are real price records), or the index creation
would fail on any install carrying one.

## Two things worth a reviewer's attention

**Dropped FKs mean dropped cascades.** V180's constraints carried
`ON DELETE CASCADE`; the catalogue's delete paths now clear links explicitly.
A cascade cannot span an optional-module boundary — the CRM tables need not
exist — which is the same trade V149 has lived with.

**A constraint name was being silently truncated.** The conventional
`<table>_<column>_check` for `manufacturer_source` on that table is 64 bytes and
Postgres truncates at 63, so the migration's own `IF NOT EXISTS` guard and the
`DROP` in `down/1` would never have matched the constraint they created — it
would have failed on re-run. Both checks are explicitly named to fit. Caught by
introspecting a migrated database, not by reading the SQL.

## Manifest and testing

The `ExpectedSchema` objects are hand-declared with catalog-exact shapes read
from a chain-built database, and `chain_hash` is restamped.

`test/integration/hand_declared_manifest_test.exs` is new: it checks the
PRODUCTION manifest against a database the real chain built. Nothing did this
before — `repair_test.exs` exercises the repair machinery against a *fixture*
manifest, and `verify.exs --scenario s7,s8` needs a generated manifest from a
pre-squash checkout, which is why the manifest header says its post-generation
objects "have never been checked against a real database". Every version from
V178 on was hand-declared, and a wrong `pos`, type or predicate makes
`mix phoenix_kit.repair` try to "fix" a healthy install.

It reports clean for V178–V180. It also surfaced four **pre-existing** defects,
left alone and documented in the test's moduledoc: two V170 expression-index key
strings with doubled parentheses, and two V171 pkeys reporting missing.

## Verified

- Full suite green; `release_check` passes Migration Version Sync and chain_hash.
- Migrated a live database at V179 → V180 and confirmed a raw `INSERT` that
  bypasses the application guard is refused by the new index.

No version or CHANGELOG changes — those belong to the release.

## Review order

This one first, then `phoenix_kit_entities` (decimal field type), then
`phoenix_kit_catalogue`, which depends on both.


mdon:
## Quality pass — external review of the migration

The migration was put to an adversarial review. Three findings held up and are
fixed; the rest were checked against the schema and did not apply.

**The dedupe and the index build could race.** The `UPDATE` takes only row
locks, so a concurrent writer could open a fresh duplicate in the gap before
`CREATE UNIQUE INDEX` acquired its own — the build would then fail and roll back
the whole of V180, federation columns included. That is the deploy-night failure
mode on a busy pricing table. Both steps now run under `SHARE ROW EXCLUSIVE`,
which blocks writers while still allowing reads.

**A loser was closed at a bare `CURRENT_DATE`.** A future-dated row would close
*before* it opened, and `validate_date_range/1` then refuses every later edit of
a row the user cannot repair. `GREATEST(valid_from, CURRENT_DATE)` fixes it and
ignores a `NULL` `valid_from`, which is the common case.

**The source CHECK was narrower than its sibling.** It accepted `local` and
`crm_company`, while `item_supplier_info.supplier_source` also accepts
`crm_contact` — so a contact-sourced supplier could not be linked to a
manufacturer at all. The values now match.

Checked and **not** applicable: `ORDER BY bool DESC` putting a NULL above `true`
(`is_primary` is `NOT NULL`); `NULL` supplier uuids sorting as distinct
(`supplier_uuid` is `NOT NULL`); the manifest still advertising the dropped FKs
(they are already flipped to `:legacy_optional`). The FK names sit at 62 bytes —
one byte inside the limit that the CHECK names exceeded.

### The manifest test earned its keep

Broadening the CHECK, I also caught V179's `cat_items` constraint, where a
manufacturer is deliberately company-only. `hand_declared_manifest_test.exs`
failed on it immediately, against a chain-built database. That is precisely the
class of error nothing was checking for before.

### Verification

Rebuilt the test database from scratch and re-read every catalog shape into the
manifest. `release_check` passes Migration Version Sync and `chain_hash`; the
full suite is green (3927).

mdon:
## Estonian and Russian translations completed

73 msgids in each locale had an **empty** `msgstr`, so they rendered English to
a client whose admin runs in et and ru. Msgid parity looked fine — an empty
translation is invisible to a parity check, which is why this needed auditing
against the `msgstr` itself.

Most are integration setup text (AI/Bedrock, SES, GitHub tokens), but a real
subset is everyday UI: *"Access requested."*, *"You don't have access to this"*,
*"Send request"*, *"Note (optional)"*, *"Unread · %{day}"*, *"private item"*,
*"No longer available"*, *"Type @ to mention someone, # to link a record."*

Interpolations, markdown link targets, backticked identifiers and proper nouns
(AWS, Bedrock, SES, IAM, GitHub, `robots.txt`, `llms.txt`) are carried through
unchanged; only the prose is translated.

### Two pre-existing bugs the audit surfaced

Checking that every `%{placeholder}` survives translation — not just that a
translation exists — found two strings that silently drop theirs:

| msgid | was | problem |
|---|---|---|
| `Account %{id}` | `Konto` / `Аккаунт` | the account id vanishes |
| `Could not reach Bedrock in %{region}` | `Brevoga ei saanud ühendust` / `Не удалось связаться с Brevo` | region vanishes **and it names the wrong service** |

The second is a copy-paste from a different provider's string, so someone
configuring Bedrock saw an error about Brevo. Both fixed.

### Result

**et 2313/2313, ru 2313/2313 translated, zero placeholder mismatches.**

The chain hash is restamped in the same commit: editing `v180.ex` during the
review pass moved it, which `release_check` and two unit tests caught. Full
suite green (3927).

PR #745, opened 2026-08-21
Add patch links and an unframed variant to nav_tabs

mdon:
> **Stacks on #744** — both touch `nav_tabs.ex`, so merge that one first.

Core already ships a "universal tab component", but most tab strips in the module libraries reimplement it instead of calling it. That is why renaming a single daisyUI class (`tabs-boxed` → `tabs-box` in v5) took **21 edits across 8 repositories** — `nav_tabs` was just one of the 21.

The duplication is not hypothetical:

- `phoenix_kit_sync` reimplements the event mode almost line-for-line — same `role="tablist"`, same `bg-base-200 p-1`, same `phx-value-tab`, same icon treatment.
- `user_connections` reimplements the path+badge mode. This component's own moduledoc uses *that exact page* as its worked example.
- `phoenix_kit_billing` copy-pastes the same 4-tab strip across 4 files, each hardcoding a different `tab-active` — and core already ships `billing_settings_tabs` with those four paths, character-for-character.
- `phoenix_kit_projects` and `phoenix_kit_staff` each wrote a near-duplicate `TabsStrip`. Staff's docstring says: *"a candidate for promotion to core once a third module needs it."* The third case already existed.

## What blocked adoption

**No `patch`.** A query-param tab strip needs it; `navigate` remounts the LiveView and drops socket state. Tabs now take `:navigate` / `:patch`, mirroring `Phoenix.Component.link/1` rather than inventing a parallel vocabulary (no `:mode` enum, no inferring from the URL). `:path` stays a permanent alias for `:navigate`, so **no existing caller changes**.

**The frame was baked in.** `class` can only ADD to the container, so a strip nested in an already-framed container had no way to shed the background and got hand-rolled instead. `variant={:plain}` drops it.

Worth recording: the frame is **`tabs-box` itself** — it draws the background, padding, radius and shadow. So `:boxed`'s `bg-base-200 p-1` are redundant overrides, and `:plain` has to drop the *modifier*, not just those two utilities. An earlier revision of this branch got that wrong and shipped a `:plain` that still drew the box.

## Failure behaviour

Setting two link keys raises — there is no defensible way to choose, and `:navigate`/`:patch` are new enough that no existing caller can trip it. A `nil` link value counts as absent, so a conditional `path: if(admin?, do: "/x")` is safe.

A tab with no link key **and** no `on_change` logs a warning and still renders. Deliberately not a raise: that combination produces a button wired to `phx-click={nil}` — a dead tab, which is bad, but taking a whole LiveView down over a tab strip is worse, and consumers of a published library cannot be audited. I checked all 7 call sites in this ecosystem; every one either passes `on_change` or gives every tab a link key.

## Scope

`tablist_class/2` and `tab_class/2` are public so tab-styled markup that is *not* a tab strip can share one definition — the next daisyUI rename should be a change here, not another sweep.

The Open Graph editor's 9 segmented controls stay **out** of the component: they are radio inputs inside a `phx-change` form, so they want `role="radiogroup"` and form plumbing, not `role="tab"`. Four independent model reviews agreed on that boundary.

Also migrates `live_sessions`' filter row, which carried a comment explaining it *could not* use the component because the payload key differed. Its handler now matches on `"tab"`.

## Module migrations are NOT in this PR

Every module pins `~> 2.0` and a conformance test enforces the wide pin, so no module can require a newer core. A module passing `:patch` to core 2.0–2.13 would land in the button branch and render a **silently dead strip** — map keys are runtime data, so no `attr` declaration can catch it. Those migrations have to wait for a release; the ones that only need the pre-2.0 API (`sync`, `billing`, `staff`, `projects`) are safe whenever you want them.

## Checks

- `mix precommit` green.
- `mix test`: 3946 tests, 43 doctests. One unrelated `permissions_test` failure that passes in isolation and has zero references to tabs — the suite has order-dependent sandbox flakes.
- 15 new component tests covering both link modes, the payload key, both variants, and each failure path.

### DIFF
--- a/lib/phoenix_kit_web/components/core/nav_tabs.ex
+++ b/lib/phoenix_kit_web/components/core/nav_tabs.ex
@@ -51,7 +51,7 @@ defmodule PhoenixKitWeb.Components.Core.NavTabs do
 
   def nav_tabs(assigns) do
     ~H"""
-    <div role="tablist" class={["tabs tabs-boxed bg-base-200 p-1", @class]}>
+    <div role="tablist" class={["tabs tabs-box bg-base-200 p-1", @class]}>
       <%= for tab <- @tabs do %>
         <%= if Map.has_key?(tab, :path) do %>
           <.link
--- a/lib/phoenix_kit_web/components/core/nav_tabs.ex
+++ b/lib/phoenix_kit_web/components/core/nav_tabs.ex
@@ -4,14 +4,14 @@ defmodule PhoenixKitWeb.Components.Core.NavTabs do
 
   Supports two modes with identical visual appearance:
 
-  **Navigation tabs** — each tab has a `path`, renders as `<.link navigate={...}>`:
+  **Navigation tabs** — each tab carries a URL, renders as `<.link>`:
 
       <.nav_tabs active_tab="general" tabs={[
-        %{id: "general", label: "General", icon: "hero-cog-6-tooth", path: "/admin/settings"},
-        %{id: "advanced", label: "Advanced", path: "/admin/settings/advanced"}
+        %{id: "general", label: "General", icon: "hero-cog-6-tooth", navigate: "/admin/settings"},
+        %{id: "advanced", label: "Advanced", navigate: "/admin/settings/advanced"}
       ]} />
 
-  **Event tabs** — no paths, uses `on_change` event via `phx-click`:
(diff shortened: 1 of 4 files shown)


mdon:
Folded into #744 — one PR per repo. Same two commits, same branch.

PR #747, opened 2026-08-23
Translate authentication and session flash messages

timujinne:
## Summary

Login, registration, magic-link, password-reset, and OAuth flash messages
were built as plain string literals instead of routing through gettext, so
non-English users got a mix of translated and raw-English notifications —
most visibly right after a successful sign-in.

Two shapes of the same bug:

- A literal string passed directly to `put_flash/2,3`.
- A message built once — a local variable, or a small helper like
  `format_ueberauth_failure/1` or the shared `account_gate/1` check used by
  three different auth gates — and only then handed to `put_flash`. This
  form doesn't show up in a naive grep of `put_flash` call sites, since no
  literal string appears at the call itself.

## Changes

- Wrapped every plain-literal and indirected flash message found in the
  authentication/session controllers and LiveViews (login, logout,
  multi-account switching/impersonation, OAuth, magic link, registration,
  email confirmation, password reset) in `gettext`/`dgettext`.
- Converted `#{}` string interpolation to named `%{}` bindings everywhere
  it appeared in a translatable string — raw interpolation breaks
  extraction and can't carry a translation.
- `PhoenixKitWeb.Users.Auth` builds its plugs via `use PhoenixKitWeb,
  :verified_routes`, which (unlike `:controller`/`:live_view`) doesn't pull
  in the Gettext macros. Added `use Gettext, backend: PhoenixKitWeb.Gettext`
  there. While in that file, also fixed one pre-existing case where a
  message was already going through `Gettext.dgettext/3` — the plain
  runtime function, not the macro — which meant `mix gettext.extract` had
  never picked it up and it had no translation in any locale.
- Added Estonian translations for every string this introduces, matching
  the terminology and register already established in the existing
  catalog rather than inventing new phrasing.
- Two of the newly-wrapped strings live in the fallback OAuth controller
  module that only compiles when Ueberauth is *not* an installed
  dependency, so `mix gettext.extract` can never see them in a normal
  build (Ueberauth is always present here). Added their Estonian catalog
  entries by hand — the `.po` header itself names this as the supported
  path for a translation that can't be statically extracted.

## Why storage/sitemap strings appear in this diff

`mix gettext.extract` scans the whole project, not just the files this
change touches — it can't be scoped. That surfaced pre-existing, unrelated
gaps already sitting in `lib/phoenix_kit/integrations/{providers,
validators}.ex` and `lib/phoenix_kit_web/live/settings/crawlers.html.heex`
(an S3-compatible object storage integration added 2026-08-17/18, never
synced into the catalog since): 13 msgids that had no catalog entry before
this branch. Left untranslated (empty `msgstr`) — same as they'd have
rendered before, raw English via gettext's fallback. None are storage/
sitemap changes on this branch's part; left for whichever queue picks up
that area.

## Two translation-correctness passes, both driven by review

**Pass 1 — this branch's own new entries.** `mix gettext.extract --merge`
fuzzy-matches a newly-added msgid against some structurally similar
existing (often orphaned) entry and carries over *its* translation,
flagging the result `fuzzy` — gettext's own signal that the pairing is a
guess, not a confirmed translation. The first version of this PR trusted
"this msgid already has a non-empty msgstr" as good enough without
checking that flag. It wasn't: 27 of the strings this PR introduces had
been fuzzy-matched to an unrelated message — e.g. `"Account created
successfully!"` had picked up the translation of `"Bucket enabled
successfully"`, `"Unknown OAuth provider: %{provider}"` had picked up a
translation missing the `%{provider}` binding entirely. Fixed by replacing
23 (the ones belonging to this PR) with the actual intended translation,
and leaving 4 empty (unrelated storage/sitemap strings caught in the same
pass, folded into the 13 above). The same fuzzy-matching hit every other
locale's extraction too (27 newly-fuzzy entries each, in
`de`/`en`/`es`/`fr`/`it`/`pl`/`ru`, none of them this PR's to translate) —
all cleared to empty rather than ship an unverifiable guess in a language
nobody here checked.

**Pass 2 — pre-existing entries sitting at this branch's tip.** 23 more
`fuzzy` entries exist in the catalog, identical in content and count on
`upstream/main` — not introduced by this branch, spanning crawlers/sitemap
settings, the Bedrock integration, and the mentions feature. Went through
each against its actual source call site: 4 were faithful, complete
translations (kept, flag cleared) and 19 didn't hold up (wrong service
name, a status label swapped in for an action button, one word standing in
for a full sentence, content from a neighboring but different message) —
cleared to empty, same as the storage/sitemap entries above.

Every one of the 68 Estonian strings this PR translates, plus the 4 kept
pre-existing entries, was re-verified by resolving it at runtime with the
locale set to `et`, including the interpolated ones with a real bound
value — not just confirming a `msgstr` was present.

## Test plan

- [x] `mix compile --force --warnings-as-errors` — clean, whole project.
- [x] `mix format --check-formatted` — clean.
- [x] Runtime check: all 68 introduced strings + the 4 kept pre-existing
      ones resolved with the locale set to `et`, interpolated ones with
      real values bound — all render the correct Estonian text, none
      empty, none leaking `%{...}`.
- [x] Confirmed zero `fuzzy`-flagged entries remain anywhere in
      `et/default.po` (was 23 pre-existing + 27 introduced by this
      branch's extraction, now 0).
- [x] Confirmed zero *new* `fuzzy`-flagged entries remain in the other
      seven locales (fuzzy counts match the pre-branch baseline exactly).
- [ ] Manual click-through of the affected pages in a running app (not
      done here).

### DIFF
--- a/lib/phoenix_kit_web/controllers/context_controller.ex
+++ b/lib/phoenix_kit_web/controllers/context_controller.ex
@@ -79,7 +79,7 @@ defmodule PhoenixKitWeb.ContextController do
 
   def set(conn, _params) do
     conn
-    |> put_flash(:error, "Invalid context")
+    |> put_flash(:error, gettext("Invalid context"))
     |> redirect(to: get_redirect_path(conn))
   end
 
@@ -98,7 +98,7 @@ defmodule PhoenixKitWeb.ContextController do
       |> redirect(to: redirect_path)
     else
       conn
-      |> put_flash(:error, "Context switching is not enabled")
+      |> put_flash(:error, gettext("Context switching is not enabled"))
       |> redirect(to: default_redirect())
     end
   end
--- a/lib/phoenix_kit_web/users/auth.ex
+++ b/lib/phoenix_kit_web/users/auth.ex
@@ -36,6 +36,7 @@ defmodule PhoenixKitWeb.Users.Auth do
   `PhoenixKitWeb.Integration.phoenix_kit_routes/0` macro in your router.
   """
   use PhoenixKitWeb, :verified_routes
+  use Gettext, backend: PhoenixKitWeb.Gettext
 
   import Plug.Conn
   import Phoenix.Controller
@@ -107,7 +108,7 @@ defmodule PhoenixKitWeb.Users.Auth do
     conn
     |> Phoenix.Controller.put_flash(
       :error,
-      "This account is not active. Contact an administrator."
+      gettext("This account is not active. Contact an administrator.")
     )
     |> Phoenix.Controller.redirect(to: Routes.path("/users/log-in"))
     |> halt()
@@ -673,7 +674,10 @@ defmodule PhoenixKitWeb.Users.Auth do
           # No `skip_admin`: the rejected visitor may well be an Admin, who
           # legitimately belongs in `/admin` — that area is gated by
           # `:phoenix_kit_ensure_admin`, not by this hook, so there is no loop.
-          |> Phoenix.LiveView.put_flash(:error, "You must be an owner to access this page.")
+          |> Phoenix.LiveView.put_flash(
+            :error,
+            gettext("You must be an owner to access this page.")
+          )
           |> Phoenix.LiveView.redirect(to: Routes.safe_destination(socket, scope: scope))
 
         {:halt, socket}
@@ -717,7 +721,7 @@ defmodule PhoenixKitWeb.Users.Auth do
             socket
             |> Phoenix.LiveView.put_flash(
               :error,
-              "You do not have the required permission to access this page."
+              gettext("You do not have the required permission to access this page.")
             )
             |> Phoenix.LiveView.redirect(
               to: Routes.safe_destination(socket, scope: scope, skip_admin: true)
@@ -742,7 +746,7 @@ defmodule PhoenixKitWeb.Users.Auth do
             socket
             |> Phoenix.LiveView.put_flash(
               :error,
-              "You do not have permission to access this section."
+              gettext("You do not have permission to access this section.")
             )
             |> Phoenix.LiveView.redirect(to: redirect_to)
 
@@ -871,7 +875,7 @@ defmodule PhoenixKitWeb.Users.Auth do
     socket
     |> Phoenix.LiveView.put_flash(
       :error,
-      "You do not have the required permission to access this page."
+      gettext("You do not have the required permission to access this page.")
     )
     # `skip_admin: true` states the invariant at the call site: this is the
     # admin-area rejection, so the destination must never be an admin page the
@@ -896,7 +900,7 @@ defmodule PhoenixKitWeb.Users.Auth do
 
   defp redirect_to_login(socket) do
     socket
-    |> Phoenix.LiveView.put_flash(:error, "You must log in to access this page.")
+    |> Phoenix.LiveView.put_flash(:error, gettext("You must log in to access this page."))
     |> Phoenix.LiveView.redirect(to: login_path_with_return_to(socket))
   end
 
@@ -1408,7 +1412,10 @@ defmodule PhoenixKitWeb.Users.Auth do
 
   defp apply_scope_refresh_decision(socket, :evict_admin_area, new_scope) do
     socket
-    |> LiveView.put_flash(:error, "You must be an admin to access this page.")
+    |> LiveView.put_flash(
+      :error,
+      gettext("You must be an admin to access this page.")
+    )
     |> LiveView.push_navigate(
       to: Routes.safe_destination(socket, scope: new_scope, skip_admin: true)
     )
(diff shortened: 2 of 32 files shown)


timujinne (commented):
Independent review before marking this ready. Three commits, and the third one
goes beyond the title — details at the end, it is separable if you would rather
not take it.

## What the first commit got wrong

`mix gettext.extract`/`merge` auto-matched new msgids against existing entries
by string similarity and wrote the guesses into the `.po` files under the
`fuzzy` flag. Gettext's compiler does not check that flag — the runtime serves
a fuzzy entry as a finished translation. So these shipped as real Estonian:

| msgid | msgstr written | what it actually translates |
|---|---|---|
| `Account created successfully!` | `Bucket lubati edukalt` | `Bucket enabled successfully` (S3 storage) |
| `Could not remove account.` | `Brevoga ei saanud ühendust` | `Could not reach Brevo` (email service) |

A user completing registration would have been told a storage bucket was
enabled.

A second defect class in the same commit: interpolation loss.
`Authentication failed: %{errors}` resolved to `Autentimine` — a non-empty
string that silently dropped the bound value.

Contamination totalled 216 fuzzy entries across all 8 locales, not only
Estonian.

## Second commit — the fix

23 entries corrected, 193 cleared to empty. An empty msgstr falls back to the
English source, which is the right outcome: an unverified translation is worse
than none.

Verified against `main`, per locale:

- **No existing translation damaged.** Entries changed: 0. Entries removed: 0.
  In the seven non-Estonian locales the count of non-empty msgstr is unchanged
  — delta exactly 0 in each.
- **No interpolation loss.** Compared the placeholder set of msgid against
  msgstr for every entry in the file. Mismatches: 0.
  `Authentication failed: %{errors}` → `Autentimine ebaõnnestus: %{errors}`,
  confirmed substituting a real bound value at runtime.
- **Both mistranslations corrected**, and the original `Bucket enabled
  successfully` / `Could not reach Brevo` entries still hold their own
  translations.
- **Scope.** All 68 non-empty Estonian translations belong to auth/session
  files. Storage/sitemap entries that `gettext.extract` pulled in as collateral
  exist in every locale but were left empty.
- `mix compile --force --warnings-as-errors` clean; all 68 strings resolve at
  runtime under `locale=et`.

## Third commit — beyond the title, and separable

While fixing the above we found the same defect class already sitting on
`main`: 23 `fuzzy` entries in `et/default.po`, not introduced by this branch.
Since the compiler ignores the flag, these were being served to Estonian users
as finished translations. Examples:

| msgid | msgstr on main | problem |
|---|---|---|
| `Bedrock error %{status}` | `Brevo viga %{status}` | names a different service |
| `Blocked groups` | `Blokeeritud` | "groups" dropped |
| `Allowed` | `Kõik` | means "all" |
| `Could not update the mailbox setting` | text of a neighbouring message | wrong message |

Each was checked against its source call site: 4 held up and kept their
translation with the flag cleared; 19 did not and were cleared to empty. The
68 auth/session strings this branch owns are untouched by that commit —
verified, exactly 0 changes.

Clearing an entry to empty means the English source shows instead, so this was
checked by running it, not by reading it: all 19 cleared entries resolve to
their English source at runtime under `locale=et` — none render as a blank
string, and the one carrying a placeholder still substitutes
(`Bedrock error %{status}` → `Bedrock error <value>`). The 4 kept entries
resolve translated, with interpolation intact.

This spans crawlers/sitemap, the Bedrock integration and mentions — domains the
PR title does not cover. It is a separate commit on purpose: take it or drop it
independently of the first two.

## How this was checked

All runtime checks were run on `80afe5ac` in an isolated worktree pinned to that
commit, against a full `mix compile --force --warnings-as-errors` (clean) and
the full test suite. 91 catalogue entries were resolved through Gettext at
runtime with distinguishable test values.

## Not verified

**Estonian by a native speaker.** The review was systematic and
dictionary-based, not a native reading. Worth a pass by a speaker before this
is treated as final copy.

The 8 test failures on this branch are pre-existing database contamination
(stale Manager/Client role rows dated before this PR's commit), confirmed by
querying the database directly; unrelated to these changes.

## Note on the PR body

The description says 67 translated strings; the actual count is 68. The
description is off by one, not the file.

timujinne:
One thing worth explaining up front, since the diff looks alarming: this PR
touches all 8 locale files with roughly **+812 −386 lines each**, including
German, Spanish and French. The title only mentions Estonian, so here is what
those lines actually are.

**No translation in any non-Estonian locale is changed or lost.** Verified
entry-by-entry against `main`:

| locale | translations changed | translations lost | new msgids |
|---|---|---|---|
| de | 0 | 0 | 79 |
| ru | 0 | 0 | 79 |

The same holds for es/fr/it/pl/en — non-empty msgstr count is identical to
`main`, delta exactly 0 in each.

What the diff is made of, per non-Estonian locale:

- **~961 lines of source references** — comment lines like
  `#: lib/phoenix_kit_web/live/users/live_sessions.html.heex:144` shifting to
  `:140`. `mix gettext.extract` recomputed line numbers after the auth/session
  modules gained `gettext()` calls. Metadata, not content.
- **79 new empty msgids** — the newly extracted auth/session strings, present
  in every catalogue as untranslated placeholders. Nothing to review there.

So the reviewable content of this PR is Estonian only, plus code changes in
`lib/phoenix_kit_web/users/*`. The other seven catalogues are churn from
extraction.

One clarification on the fuzzy cleanup mentioned in the commits: the entries
cleared in non-Estonian locales already had an **empty** msgstr — the `fuzzy`
flag was sitting on a blank translation. Clearing them changed nothing a user
would see in those languages.

PR #751, opened 2026-08-25
Doctor's orphaned-FK check now covers every foreign key, not four of seventy

timujinne:
## Problem

`mix phoenix_kit.doctor` reported a single overall `PASS` for its orphaned-FK
check while only inspecting **four hardcoded table/column pairs**. On a real
installation with ~70 foreign keys, a green report therefore meant "four of
seventy relations are intact" but read as "referential integrity is fine".

Two real orphans on a live site sat outside those four and were never
reported.

## Fix

- The check now discovers foreign keys from `pg_constraint` instead of a
  hardcoded list, so it covers every declared FK rather than a sample.
- Orphans are detected on **already-VALID** constraints too, not only on
  `NOT VALID` ones. The previous shape-matching could not express "this
  constraint is valid, yet rows violate it", which is exactly the case that
  occurs after data is loaded around a constraint.
- Probe timeouts are now measured and classified rather than guessed at, and
  the catch-all clause was narrowed so an unexpected shape no longer passes
  silently as clean.
- Composite foreign keys render correctly in the report.
- Each report line names the scope it actually checked, so a green line can
  no longer be read as broader than it is.
- `@moduledoc` for this check was corrected to describe what the check covers
  now. It previously promised detection for relations where no constraint
  exists at all — that path is not reachable in the current implementation,
  and the documentation said otherwise.

## Verified

- Destructive round-trip on a live database: pass → inject an orphan row →
  check fails → rollback → pass. The failure is the point: a check that stays
  green with a planted orphan would be worthless.
- `--prefix=nonexistent_schema` (no objects at all) no longer reports a bare
  `PASS`; it reports that there was nothing to check.
- Test coverage added for both the discovery path and the orphaned-FK
  classification.

## Not verified

- No run against a database with composite foreign keys in production-scale
  volume; composite rendering was verified on constructed cases only.
- Relations that have no declared foreign key are still outside this check's
  scope. Removing that gap needs a different discovery mechanism (schema
  introspection rather than `pg_constraint`) and is deliberately not part of
  this change — the documentation now states the limit instead of implying it
  is covered.

### DIFF
--- a/lib/mix/tasks/phoenix_kit.doctor.ex
+++ b/lib/mix/tasks/phoenix_kit.doctor.ex
@@ -854,7 +854,15 @@ defmodule Mix.Tasks.PhoenixKit.Doctor do
   # a catalog query that itself failed — can never read as PASS, no matter
   # which of those reasons caused it (I055 finding 2 and 3: "clean" and
   # "never looked" used to print byte-identical text).
-  defp check_orphaned_fk_refs(prefix) do
+  #
+  # Exposed (not `defp`) and `@doc false`, same reason as the other pure
+  # decision functions in this module: `get_repo!/0` resolves the same
+  # `PhoenixKit.Test.Repo` under `mix test` that it resolves under a real
+  # `mix phoenix_kit.doctor` run, so this is a real end-to-end seam for the
+  # whole discover -> probe -> classify -> report pipeline against a live
+  # connection, not just its pieces in isolation.
+  @doc false
+  def check_orphaned_fk_refs(prefix) do
     repo = get_repo!()
     escaped_prefix = String.replace(prefix, "'", "\\'")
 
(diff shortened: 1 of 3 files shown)


timujinne (commented):
Independent review of this branch at `fbfaa2ad`.

## What was checked, and how

**The core claim — every FK instead of four.** `discover_fk_constraints/2`
reads `pg_constraint` directly rather than matching a hardcoded list, so the
set grows with the schema instead of with the code. Verified by reading the
query, not by trusting the description.

**Orphans on already-VALID constraints.** The previous shape-matching could
not express "this constraint is valid, yet rows violate it" — which is exactly
what happens when data is loaded around a constraint. The new clause covers it.

**Honesty of the report — the part worth dwelling on.** Three cases now say
what they actually mean rather than defaulting to a green line:

- catalog unreadable → `coverage is zero, which is not the same as clean`
- no constraints found → same wording, plus a hint that `--prefix` may name
  the wrong schema
- some probes failed → `checked N-k of N — k more not checked`

That last one matters most. The bug this PR fixes was not a wrong answer, it
was a **narrow answer presented as a complete one**. A check that silently
inspects a subset and reports `PASS` is worse than no check, because it
consumes the attention that would otherwise go to a real audit.

## Remaining scope boundary — stated, not hidden

Foreign keys whose **referenced** table lives in a different schema are
excluded from discovery: both the owning and the referenced namespace are
constrained to the target prefix.

This is documented in the moduledoc rather than silently accepted, which is
the right call for this change — widening the scope is separate work with its
own migration concerns. Worth knowing when reading `checked N of N`: that N is
the count of single-column, same-schema constraints.

The composite-key path is also excluded from automated probing and says so,
directing the reader to verify manually.

## Findings raised and addressed during review

- A warn-tier message suggested "fix connectivity and re-run" on composite
  entries, where a re-run can never help, while the adjacent line said "verify
  manually" — one message contradicting itself. Corrected.
- An unused private helper was left in place. Rather than deleting it silently
  or leaving it unexplained, its reason for remaining is now written down —
  an unreferenced function with no comment reads as forgotten code to the next
  reader.

## Not verified

- **Runtime.** No live run was performed by the reviewer; the analysis is by
  code. The author reports a destructive round trip on a live database — pass,
  plant an orphan, fail, roll back, pass — and that failure step is the part
  that gives the result meaning.
- **Timing at scale.** The probe walks constraints sequentially with a
  per-constraint timeout; on an installation with a few hundred constraints
  the worst case runs into minutes. Not introduced here, and not measured —
  flagged so it is a known property rather than a surprise.
- **Report wiring.** No test exercises the assembly of the final report, only
  the classification feeding it. Pre-existing gap, noted for completeness.

PR #752, opened 2026-08-25
Stop OAuth and AWS setting secrets from leaking into the application log

timujinne:
## Problem

OAuth and AWS secret values were written to the application log in plain text.

The leak happens through Ecto's query logging: settings are persisted with
`UPDATE phoenix_kit_settings SET value_json = $1 ...`, and the parameter list
is logged alongside the statement. Any setting holding a secret — OAuth client
secrets, AWS credentials — therefore lands in the log verbatim, once per save.

This is not hypothetical. On one installation the log held 17 occurrences of a
single OAuth client secret and was still growing; the file was 34 MB and
world-readable at the time it was found.

Two properties make this worse than a one-off mistake:

- the log is append-only and long-lived, so a secret rotated months ago is
  still recoverable from it;
- file permissions are not a defence here. On that installation the log is
  recreated by the process supervisor, which resets the mode — tightening it
  by hand holds only until the next restart.

## Fix

Secret-bearing setting values are filtered before the query reaches the
logger, so the parameter is replaced rather than the whole statement being
silenced. Normal settings keep logging their values, which matters for
debugging.

## Verified

Destructively, not by inspection:

- with the fix disabled, a canary value planted through the settings path
  appears in the log through both routes it can take;
- with the fix enabled, the canary does not appear, **and the write itself
  still succeeds** — the guard filters the log, not the operation;
- full suite: 4045 tests, 0 failures.

The middle point is the one worth stating: a filter that also breaks the write
would pass a naive "secret no longer in log" check.

## Not verified

- Whether previously leaked values are removed. They are not: this change
  stops new writes from leaking. Existing log content is an operational
  matter for whoever runs the installation.
- Live sign-in through an OAuth provider after the change. The test canary
  exercises the same write path, but an end-to-end login was not performed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,12 @@ paths — closes the second half of the DOM leak fixed in 2.13.9.
   filter with a fresh one carrying Phoenix's own default plus these — found,
   by a destructive test, that Phoenix pre-compiles this on every boot
   regardless of host config, so "leave an already-configured filter alone"
-  meant "leave every real host alone"). No host action required — every app
-  already calls `boot/1` via `phoenix_kit.install`/`update`.
+  meant "leave every real host alone"). Also merges in every
+  `PhoenixKit.Settings.restricted_setting_keys/0` entry by exact name —
+  `aws_access_key_id` names neither half of a keypair by the generic
+  words above, so it only gets caught by being on that list. No host
+  action required — every app already calls `boot/1` via
+  `phoenix_kit.install`/`update`.
 - **Every write to `phoenix_kit_settings` was logged with its raw value at
   Ecto's default `:debug` SQL log level** (`UPDATE ... SET value = $1 ...
   [<secret>, ...]`) — the table stores every setting's value in the same
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,31 @@
+## 2.13.10 - 2026-08-25
+
+A live install's default (dev-parity) logging wrote OAuth/AWS setting
+secrets straight into `/var/log/elixir.log` in cleartext, on two independent
+paths — closes the second half of the DOM leak fixed in 2.13.9.
+
+### Fixed
+
+- **Saving a Settings/Authorization credential (`oauth_google_client_secret`
+  and friends) logged the real value.** `Phoenix.LiveView.Logger` writes a
+  `"HANDLE EVENT ... Parameters: ..."` line for every `handle_event`, filtered
+  only by `config :phoenix, :filter_parameters` (Phoenix's own default:
+  `password`/`token`) — a value the settings form's field names
+  (`oauth_google_client_secret`, `billing_stripe_secret_key`, ...) never
+  matched. `PhoenixKit.boot/1` now installs `secret`/`api_key` into that list
+  itself (merging into `{:keep, [...]}` mode; replacing an already-compiled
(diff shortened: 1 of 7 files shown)


timujinne (commented):
Independent review of this branch at `460692e5`.

## What was checked

**Coverage of the leak path.** The two functions writing to the settings table
were traced to confirm they are the only ones; one exception exists — the
integration key-rotation path writes directly, bypassing them. That one was
examined separately: it stores already-encrypted material, so even reaching a
debug log it carries ciphertext, not a secret. Inconsistent with the new
convention, not a leak.

**That the fix does not silence the operation.** A filter which also breaks
the write would satisfy a naive "the secret is no longer in the log" check.
There is an explicit test asserting the write still happens — this is the
detail that makes the rest of the evidence meaningful.

**That the tests are not tautological.** The integration test raises the
logger level deliberately rather than relying on a helper that could let the
assertion pass against an empty log.

## Finding addressed during review

The LiveView-side filter matched secrets by substring in the parameter name
(`password`, `token`, `secret`, `api_key`), while a separate list of
credential keys existed elsewhere in the codebase. A key present in that list
did not match any of the substrings.

No live path was found where such a field is actually submitted — no current
template renders it as a form input — so this was a divergence between two
sources of truth rather than an exploitable gap. It has been closed rather
than left for the next person to rediscover.

## Correction to this PR's own description

The original text said normal settings keep logging their values. That is
accurate for the LiveView path, where filtering is by key name. It is not
accurate for the SQL path: logging is disabled there for **every** settings
write, secret or not. That is a deliberate trade — the statement, not the
behaviour, was imprecise.

## Not verified

- **No live run.** No sign-in through an actual OAuth provider was performed;
  the canary exercises the same write path, which is close but not identical.
- **Full suite not re-run by the reviewer.** The author reports 4045 tests
  with no failures. The structural checks here — no dangling references, no
  duplicate message ids — are consistent with that but do not replace a run.
- **Logging outside this library.** A host application with its own telemetry
  could log request parameters independently; that is outside what this change
  can reach.
- **Previously leaked values are not removed.** This stops new writes only.
  Existing log content, and the file permissions that expose it, are
  operational matters — on the installation where this was found, the log file
  is recreated by the process supervisor, which resets its mode, so tightening
  permissions by hand does not survive a restart.

timujinne:
Two boundaries worth stating explicitly, from a second review pass. Neither
blocks the change — both are about what it does *not* cover, so that the
guarantee is not read wider than it is.

**Host-side telemetry is out of reach.** This filters two paths: LiveView
event logging and the Ecto query log. A host application with its own
telemetry handler — anything attaching to `[:phoenix, :endpoint, :stop]` or
similar and logging parameters itself — would still print what it collects.
Nothing in a library can prevent that; the point is that "secrets no longer
reach the log" is true for the paths this library controls, not for the
process as a whole.

**Installing into `filter_parameters` can displace host entries.** The boot
step merges into `{:keep, [...]}` form, but when the application's list has
already been compiled to its other representation, the write replaces rather
than unions. An application that had configured its own words there could
lose them.

The code comment calls this out honestly as the one real cost. It is worth
having in the changelog too, since the person affected would be an integrator
who never reads this file — they would notice only when a field they had been
filtering starts appearing in logs.

A plain-list union instead of replace would remove the trade-off entirely, if
that path is worth the extra branch.

PR #753, opened 2026-08-26
Dedupe admin route paths across host pages and modules

timujinne:
## What breaks

A host application can declare its own admin page via the `:admin_dashboard_tabs`
config (`compile_custom_admin_routes/1`), and a discovered external module can
declare an admin tab via its `admin_tabs/0` callback
(`compile_module_admin_routes/0`). If the two happen to resolve to the same
URL path, the router ends up with **two `live` declarations at an identical
path**. Phoenix keeps the first and the second becomes dead code — reachable
only by luck of declaration order.

In practice this is discovered by accident, via the compiler's
`this clause cannot match` warning — and on an install that doesn't build with
`--warnings-as-errors`, it isn't discovered at all.

This is not a hypothesis: we hit it ourselves — a module's admin page was
silently unreachable until we moved it to a different path.

## Two defect locations — a fix in only one does not close it

- `collect_module_tabs/2` (`lib/phoenix_kit_web/integration.ex`) already calls
  `Enum.uniq_by(&(&1.path))`, but only across **one module's own** tab list.
- `compile_module_admin_routes/0` `flat_map`s that per-module list across
  **all** discovered modules with no dedup between them, and never checks
  against the host's own `:admin_dashboard_tabs` pages either.

A test that only exercises the first location goes green while the second
defect is still live.

## The fix

`phoenix_kit_admin_routes/1` now dedupes `plugin_admin_routes` (module tabs)
against `custom_admin_routes` (host pages) by resolved path, via a small
`dedupe_admin_routes_by_path/2` helper that operates on the quoted `live`
route forms both sides already produce.

**Who wins:** the host-declared page. It was declared by the application
owner, not a module author, and it was already emitted first in the route
table. Ties between two modules keep whichever is discovered first — but
unlike the ordering *within* one module's own tab list (which is
deterministic, set by whoever wrote that list), the ordering *between*
modules comes from `discover_external_modules/0`, i.e. `:code.get_path()` —
codepath order, which is not a guaranteed-stable ordering. There is no
dedicated test in this PR for the two-module collision case.

## Proof: red → green → red → green

A new test, `test/phoenix_kit_web/admin_route_dedup_test.exs`, compiles a
throwaway router (not `PhoenixKitWeb.Router` — that router special-cases
itself out of both route-generation paths, since parent app modules aren't
available while it compiles, so it can never exhibit this defect) with a host
page and a module tab pointed at the same path, plus a second pair at
distinct paths as a control.

```
before the fix:  assert length(matches) == 2   -> found 4
                 (2 sources × 2 URL shapes: plain + locale-prefixed)
                 compiler also warns twice: "this clause cannot match"
after the fix:   2 tests, 0 failures — warning gone
fix reverted:    found 4 again, warning back
fix restored:    2 tests, 0 failures again
```

The control pair (distinct paths) stays at 2 routes throughout, before and
after the fix — the dedup does not swallow a real page it shouldn't touch. A
separate assertion confirms the surviving route at the colliding path is the
host's LiveView, not the module's.

Full core suite after the fix: `43 doctests, 4068 tests, 0 failures`.

## What this does NOT close, on purpose

- **`user_dashboard_tabs`** goes through a separate code path
  (`phoenix_kit_authenticated_routes`) and is not touched by this fix. Kept
  out of scope deliberately — the underlying mechanism is the same, but
  extending the fix there would go beyond what's proven with a live test
  case in this PR.
- **`external_admin_routes`** (multi-page module routes, e.g. the shop admin
  routes) are also left alone — they aren't single-path tabs, so a
  path-based dedup doesn't apply to them the same way.
- **A module tab colliding with one of PhoenixKit's own core `/admin/...`
  routes** (`/admin/users`, `/admin/jobs`, `/admin/modules`, and the rest of
  the ~50 routes declared literally inside `phoenix_kit_admin_routes/1`,
  `lib/phoenix_kit_web/integration.ex:521-596`) is **not** closed by this fix
  either. The dedup added here only checks module tabs against
  `custom_admin_routes` (host pages) — it never sees the core routes, which
  aren't a list at all, just `live` calls written directly in the macro body.
  Verified by hand: a module tab pointed at `/admin/users` still produces 4
  routes (2 sources × 2 URL shapes) with this fix applied, same as before.
  Not fixed here because closing it needs either a manually maintained list
  of every core admin path (which will drift the next time one is added or
  renamed) or restructuring the macro to build its route table as data
  instead of literal `live` calls — and, more importantly, it needs a
  maintainer call on who should win a module-vs-core collision, and whether
  the right answer is silent dedup at all rather than a loud build-time
  warning. Flagging it here rather than leaving it to be found the same way
  the module-vs-module gap was.

Each of these would need its own dedicated fix and test if it turns out to
matter.

### DIFF
--- a/lib/phoenix_kit_web/integration.ex
+++ b/lib/phoenix_kit_web/integration.ex
@@ -1087,8 +1087,13 @@ defmodule PhoenixKitWeb.Integration do
     {deduped, _seen} =
       Enum.reduce(routes, {[], seen}, fn route, {acc, seen} ->
         case admin_route_path(route) do
-          nil -> {[route | acc], seen}
-          path -> if MapSet.member?(seen, path), do: {acc, seen}, else: {[route | acc], MapSet.put(seen, path)}
+          nil ->
+            {[route | acc], seen}
+
+          path ->
+            if MapSet.member?(seen, path),
+              do: {acc, seen},
+              else: {[route | acc], MapSet.put(seen, path)}
         end
       end)
 
--- a/lib/phoenix_kit_web/integration.ex
+++ b/lib/phoenix_kit_web/integration.ex
@@ -453,6 +453,18 @@ defmodule PhoenixKitWeb.Integration do
     # so plugin LiveViews don't need to wrap with LayoutWrapper themselves
     plugin_admin_routes = compile_plugin_admin_routes(__CALLER__.module)
 
+    # A module tab whose path collides with a host-declared custom admin
+    # page, or with another module's tab, would otherwise compile to TWO
+    # `live` declarations at the same path — Phoenix keeps the first and
+    # the second becomes dead code, discovered (if at all) only via the
+    # compiler's "this clause cannot match" warning. Host pages win: the
+    # application owner declared them, not a module author, and they were
+    # already emitted first. Ties between modules keep whichever was
+    # discovered first — the same precedence `collect_module_tabs/2`
+    # already uses for tabs within a single module.
+    plugin_admin_routes =
+      dedupe_admin_routes_by_path(plugin_admin_routes, custom_admin_routes)
+
     # Shop admin routes.
     #
     # `phoenix_kit_ecommerce` declares `route_module/0`, so auto-discovery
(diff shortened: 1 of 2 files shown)


timujinne:
## Independent review

Two reviewers looked at this separately, without contact, and their verdicts
disagreed. Every finding below was then re-checked before posting; nothing
here rests on a reviewer's word alone.

**What the disagreement was about.** One review found nothing blocking. The
other read the macro and claimed the dedup never compares module tabs against
PhoenixKit's own hardcoded `/admin/...` routes — only against the host's
`:admin_dashboard_tabs`. That was found by reading, so it was verified by
running: a throwaway router with a module tab declared at `/admin/users`
compiles to **four** routes at that path — `PhoenixKitWeb.Live.Users.Users`
and the module's LiveView, once per URL shape — with this fix applied. The
gap is real, and it is now stated in the description rather than left
implicit. It is not closed here on purpose: the core routes are literal
`live` declarations inside the macro rather than a list, so comparing against
them means either hand-maintaining ~50 paths that will drift from the code or
restructuring the macro — and, more to the point, it needs a decision on who
should win such a collision and whether a loud warning would serve better
than a silent drop. That is the maintainer's call, not something to slip into
this PR.

**A claim that was too strong, corrected.** Both reviewers independently
flagged that the description equated the module-vs-module tie-break with
`collect_module_tabs/2`'s. Within one module the order is the author's list
and is deterministic; across modules it comes from
`discover_external_modules/0` and therefore from codepath order. The wording
now says that plainly and admits there is no test for that case.

**Confirmed by running, not by reading:**

- the test genuinely fails without the fix — reverting only `lib/` gives
  `found 4` at the colliding path with two `this clause cannot match`
  warnings; restoring it gives `2 tests, 0 failures`, and the tree comes back
  clean;
- distinct paths still produce distinct routes before and after, so the dedup
  does not quietly swallow a legitimate page;
- the localized `/:locale/...` and plain forms are counted separately —
  `phoenix_kit_admin_routes/1` expands twice, each with its own set, so they
  cannot collapse into each other;
- the surviving route at a collision is the host's, which the test asserts
  directly;
- `user_dashboard_tabs` really does travel a separate code path and is
  untouched by the diff.

**Stated as unverified:** the full-suite figure quoted in the description was
not re-run during review, and the cross-module tie-break was traced through
the code but never reproduced with two genuinely conflicting modules.

PR #754, opened 2026-08-26
Stop "Test Connection" from stamping unverified providers as connected

timujinne:
## The bug

Clicking **Test Connection** stamps a provider "connected" without actually contacting it, for any provider that doesn't match one of `do_validate/2`'s specific clauses.

Concretely, on a live site: an operator clicked **Test Connection** on a Shopify connection, got the green "connected" result, and trusted it. The Shopify Admin API was actually returning 401 the whole time. The stored row read:

    status=connected  validation_status=ok

with zero HTTP requests made to Shopify.

## Where

`lib/phoenix_kit/integrations/integrations.ex`, three separate fallbacks inside `do_validate/2` all resolve to the unconditional `:ok` catch-all instead of running (or reporting the absence of) a real check:

- `:1130` — `defp do_validate(_, _data), do: :ok` — the outer catch-all: any provider whose `auth_type` doesn't match one of the clauses above it.
- `:1112-1128` — an `:api_key`/`:bot_token` provider with no `validation` map configured falls to `true -> :ok` (line 1125-1126).
- `:1083` — an `:oauth2` provider whose `oauth_config` has no `userinfo_url` falls to `is_nil(userinfo_url) -> :ok`.

Every one of these paths is reachable by a provider with zero HTTP validation, yet stamps the same `status=connected, validation_status=ok` a genuinely-checked provider gets.

## Scope, honestly

Today this affects **one provider out of sixteen**: `shopify` (contributed by `phoenix_kit_ecommerce`), whose `auth_type: :credentials` matches none of the specific clauses and has no `validation` map — it falls straight into the `:1130` catch-all.

`google` and `microsoft` are also `:oauth2`, but both declare a real `userinfo_url` — clicking Test Connection on either makes an actual authenticated HTTP request. They're unaffected.

The defect is structural, not shopify-specific: any future provider with an unhandled `auth_type`, or an `:api_key`/`:bot_token` provider added without a `validation` map, or an `:oauth2` provider without `userinfo_url`, lands in the same silent-`:ok` trap. This PR closes the structural gap, not just today's one instance.

## The fix

The three fallbacks above now return `:unverified` instead of `:ok`. `validation_fields/1` maps `:unverified` to:

    {"configured", "Not verified — this provider has no connection check"}

**Why the existing `"configured"` status, not a new one:** `integrations_ui.ex:26` already labels `"configured"` as *"Not tested"* in the status badge — that's exactly the right operator-facing meaning. A brand-new status string would fall into the badge's catch-all clause (`integrations_ui.ex:29`) and render *"Not configured"*, which is wrong (the provider IS configured — it just hasn't been checked), and would require a UI change this PR doesn't need to make.

**This restores prior behavior, it doesn't invent new behavior:** a connection already reads `status=configured` immediately after its credentials are saved (`maybe_set_status/2`, `:1508-1534`) — `"connected"` only appears after a successful Test Connection. Today's bug is that clicking the button can *downgrade* the operator's information by overwriting an honest "not tested yet" with a false "connected". The fix makes an unverifiable provider's Test Connection a no-op on that front, instead of a false promotion.

The 13 providers with a real check (`validation` map, `strategy`, or `oauth2` + `userinfo_url`) are unaffected — they still report `"connected"` on success and `"error"` on failure exactly as before.

## Prior art in this codebase

The same failure mode was already fixed once, per-provider, for Brevo — `integrations.ex:1088-1093`:

> Without these clauses these fell through to the `:ok` catch-all below (Brevo), ... so "Test Connection" stamped the connection "connected" without verifying anything

This PR generalizes that fix so the next provider doesn't need its own point patch.

## What the first version of this PR missed (caught in review, not by me)

The first version closed the **data layer** — `do_validate/2` and `validation_fields/1` — and stopped there. It did not close the **UI layer**: every place that actually calls `validate_credentials/2` / `validate_connection/3` and looks at the result still only matched `:ok` / `{:ok, _}` / `{:error, _}`. Two independent reviews caught this and it's a real regression, not a nitpick — clicking Test Connection on the exact class of provider this PR exists to protect would **crash the LiveView** instead of showing "Not tested". That's worse than the original bug: a crash instead of a lie.

Four sites needed a fourth clause:

- `integration_form.ex:451` — dry-run test before a row exists (`validate_credentials/2`)
- `integration_form.ex:335` — Test Connection after save (`validate_connection/3`)
- `my_integration_form.ex:121` — same dry-run, personal form
- `my_integration_form.ex:233` — same post-save test, personal form

Plus the two `@spec`s on `validate_connection/3` and `validate_credentials/2` themselves, which still promised only the old three-shape contract.

**Writing a regression test for this found a fifth site the reviews didn't catch:** `validate_connection/3` has its own *internal* `case result do ... end` (`integrations.ex:978`) purely for activity logging — also missing `:unverified`, also a `CaseClauseError`, entirely separate from the four UI sites above. The two reviews read the LiveView call sites; neither traced into `validate_connection/3`'s own body. It only surfaced because the new test drives the real function end to end instead of asserting against a hand-constructed `:unverified` value — see "Testing" below.

All five now handle `:unverified`. In the UI, it renders as its own tone — not the green "Connection verified" (that repeats the bug), not the red "Test failed" (nothing failed, nothing ran) — a `warning`-styled "Not tested — this provider has no connection check" on both forms (system and personal alike; see the correction below).

**A second, independent review round** (two more reviewers, each in their own isolated checkout) confirmed there's no sixth crash site, and independently reproduced the isolation numbers below exactly — the five sites really are needed independently, not by coincidence of a combined revert. It also caught two smaller issues before merge:

- The `:warning` tone above shipped correctly on the system form but as `:info` on the personal form — on the reasoning that this codebase's `put_flash` calls never use anything but `:info`/`:error`. That reasoning was checked and is **wrong**: `:warning` is a first-class kind in `Components.Core.Flash` (`flash.ex:24`, its own color and icon, rendered by `flash_group/1` alongside `:info`/`:error`) and is already used at `users/user_form.ex:833` and `modules/storage/web/health.ex:52`. Fixed — both forms now show `:warning`, matching each other and the badge color for `"configured"`.
- `Events.broadcast_validated/3` (`events.ex`) still carried the old `:ok | {:error, term()}` spec, even though `record_validation/2` (`integrations.ex:1234`) has called it with `:unverified` and `{:ok, note}` since round 2. A clean `mix dialyzer` run didn't fail on it — success-typing is wider than the declared spec — but the contract itself was stale. Fixed; three `@spec`s updated in total, not two.

## Testing

TDD, full red → green → red → green cycle, in three rounds:

**Round 1 — the data layer** (`do_validate/2`, `validation_fields/1`):

- New tests (`do_validate/2`'s three fallbacks + `record_validation/2` with `:unverified`) — **RED**: 104 tests, 6 failures.
- After the fix — **GREEN**: 104 tests, 0 failures.
- Fix reverted — **RED** again: 104 tests, 8 failures (the extra 2 are the test-only shim described below, which reverts along with the fix).
- Fix restored — **GREEN**: 104 tests, 0 failures.

**Round 2 — the UI layer**, after review (the four call sites + the internal `validate_connection/3` case above). New tests go through the **real public path** — no shim, no hand-built `:unverified` value: a fixture provider (shaped exactly like Shopify — `auth_type: :credentials`, no `validation` map) is registered via `PhoenixKit.ModuleRegistry.register/1` (the same mechanism a real external module uses, and the same one this codebase's own tests already use for fixture modules), and the tests drive the actual LiveView events a click sends — form submit, `Test Connection` click — end to end through `validate_credentials/2` and `validate_connection/3` for real:

- New tests (4: dry-run + post-save, on both the system and personal forms) — **RED**, all 4 fixes reverted together: `CaseClauseError, :unverified` — 4 tests, 4 failures.
- After the fix — **GREEN**: 4 tests, 0 failures.
- Fix reverted — **RED** again: 4 tests, 4 failures.
- Fix restored — **GREEN**: 4 tests, 0 failures.
- Isolated the fifth (internal) site from the four UI sites, since a combined revert doesn't prove each is independently needed:
  - LiveView cases reverted, internal `validate_connection/3` case left fixed — all 4 tests crash, each at its own UI case line (`integration_form.ex:451`, `:348`; `my_integration_form.ex:121`, `:246`) — 4 failures.
  - Internal case reverted, LiveView cases left fixed — only the 2 **post-save** tests crash, at `integrations.ex:979` before `validate_connection/3` ever returns to the caller; the 2 dry-run tests pass unaffected, because `validate_credentials/2` never goes through that function — 2 failures.

**Round 3 — the tone fix**, after the second review round: `my_integration_form.ex`'s two `:unverified` clauses changed `:info` → `:warning`. Strengthened the existing round-2 tests to assert the rendered CSS class, not just the message text (`refute alert-error` alone doesn't prove `:warning` over `:info`):

- `assert html =~ "alert-warning"` added to both personal-form tests, `:warning` reverted to `:info` — **RED**: 2 tests, 2 failures (both on the new assertion — everything else about the fix still holds, `:info` doesn't crash, it's just the wrong color).
- Fix restored — **GREEN**: 2 tests, 0 failures.

**Full suite, all three rounds combined: 43 doctests, 4078 tests, 0 failures** — the 13 providers with a real validation strategy are unaffected. `mix dialyzer` (clean run) and `mix credo --strict` pass on every file this PR touches.

## A compile-time test shim (flagging it up front)

No currently-registered provider (all 15 built-ins) actually falls into any of the three fixed clauses — every one of them already has a real check. That means there is no way to exercise this behavior for a *future* provider through the public API with today's registry. I added a narrow, compile-time-gated shim to make the fallback clauses directly testable:

    if Mix.env() == :test do
      @doc false
      def __do_validate__(provider, data), do: do_validate(provider, data)
    end

Verified it doesn't exist outside test builds:

    dev build:  function_exported?(PhoenixKit.Integrations, :__do_validate__, 2) == false
    test build: function_exported?(PhoenixKit.Integrations, :__do_validate__, 2) == true

`Mix.env()` is evaluated at compile time here (module level), so the function is compiled out entirely in non-test builds — it's not a runtime check that could slip through in a release build. The same technique with the same rationale already exists in this repo at `router.ex:52`.

If you'd rather test this a different way, happy to rework it.

### DIFF
--- a/lib/phoenix_kit/integrations/events.ex
+++ b/lib/phoenix_kit/integrations/events.ex
@@ -73,8 +73,11 @@ defmodule PhoenixKit.Integrations.Events do
   end
 
   @doc "Broadcast that an integration's health check completed."
-  @spec broadcast_validated(String.t(), :ok | {:error, term()}, owner()) ::
-          :ok
+  @spec broadcast_validated(
+          String.t(),
+          :ok | {:ok, String.t()} | :unverified | {:error, term()},
+          owner()
+        ) :: :ok
   def broadcast_validated(provider_key, status, owner \\ :system) do
     broadcast(owner, {:integration_validated, provider_key, status})
   end
--- a/lib/phoenix_kit/integrations/events.ex
+++ b/lib/phoenix_kit/integrations/events.ex
@@ -18,7 +18,7 @@ defmodule PhoenixKit.Integrations.Events do
   - `{:integration_setup_saved, provider_key, data}` — setup credentials saved
   - `{:integration_connected, provider_key, data}` — OAuth flow completed
   - `{:integration_disconnected, provider_key}` — disconnected
-  - `{:integration_validated, provider_key, :ok | {:error, reason}}` — health check
+  - `{:integration_validated, provider_key, :ok | {:ok, note} | :unverified | {:error, reason}}` — health check
   - `{:integration_connection_added | _removed, provider_key, name}`
   - `{:integration_connection_renamed, provider_key, old, new}`
 
--- a/lib/phoenix_kit/integrations/integrations.ex
+++ b/lib/phoenix_kit/integrations/integrations.ex
@@ -1080,7 +1080,7 @@ defmodule PhoenixKit.Integrations do
 
     cond do
       not (is_binary(token) and token != "") -> {:error, gettext("No access token")}
-      is_nil(userinfo_url) -> :ok
+      is_nil(userinfo_url) -> :unverified
       true -> check_http(userinfo_url, [{"authorization", "Bearer #{token}"}])
     end
   end
@@ -1123,11 +1123,21 @@ defmodule PhoenixKit.Integrations do
         check_http(v.url, headers)
 
       true ->
-        :ok
+        :unverified
     end
   end
 
-  defp do_validate(_, _data), do: :ok
+  defp do_validate(_, _data), do: :unverified
+
+  if Mix.env() == :test do
+    # Exposes the private do_validate/2 clauses for direct unit testing.
+    # No registered provider today reaches the generic/catch-all clauses
+    # (every built-in declares a real check), so the only way to prove
+    # those clauses behave correctly for a *future* provider is to call
+    # them directly with a synthetic provider map.
+    @doc false
+    def __do_validate__(provider, data), do: do_validate(provider, data)
+  end
 
   defp check_http(url, headers) do
     case Req.get(url, headers: headers, receive_timeout: @http_timeout) do
@@ -1153,8 +1163,11 @@ defmodule PhoenixKit.Integrations do
   actual state change so high-frequency automatic paths (e.g. token
   refresh failing on every API call) don't spam listing-LV reloads.
   """
-  @spec record_validation(String.t(), :ok | {:ok, String.t()} | {:error, term()}, keyword()) ::
-          :ok
+  @spec record_validation(
+          String.t(),
+          :ok | {:ok, String.t()} | :unverified | {:error, term()},
+          keyword()
+        ) :: :ok
   def record_validation(uuid, result, opts \\ []) when is_binary(uuid) do
     {new_status, validation_text} = validation_fields(result)
 
@@ -1216,6 +1229,13 @@ defmodule PhoenixKit.Integrations do
   end
 
   defp validation_fields(:ok), do: {"connected", "ok"}
+
+  # No check actually ran (a provider with no way to verify itself — see
+  # the do_validate/2 fallback clauses). Reported as the same "configured"
+  # status save_setup/3 already uses for "credentials saved, nothing tested
+  # yet" — NOT "connected", which the operator reads as verified.
+  defp validation_fields(:unverified),
+    do: {"configured", gettext("Not verified — this provider has no connection check")}
 
   # A check that succeeded but could not verify everything it would have liked to.
   # The connection is usable; the note is what the operator must know about it, and
--- a/lib/phoenix_kit/integrations/integrations.ex
+++ b/lib/phoenix_kit/integrations/integrations.ex
@@ -952,10 +952,11 @@ defmodule PhoenixKit.Integrations do
 
   For OAuth: calls the provider's userinfo endpoint.
   For API key / bot token: calls the provider's validation endpoint if defined.
-  Returns `:ok` or `{:error, reason}`.
+  Returns `:ok`, `{:ok, note}`, `:unverified` (the provider has no way to
+  check a connection — see `do_validate/2`), or `{:error, reason}`.
   """
   @spec validate_connection(String.t(), String.t() | nil, keyword()) ::
-          :ok | {:ok, String.t()} | {:error, String.t()}
+          :ok | {:ok, String.t()} | :unverified | {:error, String.t()}
   def validate_connection(uuid, actor_uuid \\ nil, opts \\ []) when is_binary(uuid) do
     {result, log_provider, log_name} =
       case resolve_uuid(uuid, Keyword.get(opts, :owner, :system)) do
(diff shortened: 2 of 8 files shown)


timujinne (commented):
## Independent review

Two rounds, two reviewers each, working from separate checkouts so their
mutations could not race. Every finding was re-checked before posting.

**Round one found the fix incomplete, and both reviewers found the same
thing independently.** The new `:unverified` outcome was matched nowhere
outside the module that introduced it, so pressing Test Connection for
exactly the provider class this PR exists for would have raised
`CaseClauseError` instead of showing "Not tested" — replacing a lie with a
crash. Verified here too: the atom appeared in no other file.

Worth recording why the green tests missed it: they reached the private
function through a compile-time shim and called `record_validation/2` with
the atom supplied by hand. Both routes go around the public API the button
actually calls. Writing the test through `validate_credentials/2` and
`validate_connection/3` instead is what then surfaced a **fifth** site
neither reviewer had seen — the internal `case result do` that logs the
activity. A test that walks the user's path found what three careful reads
did not.

**Round two confirmed there is no sixth.** Both reviewers exhausted the
search independently — every caller of `validate_connection/3` and
`validate_credentials/2`, every subscriber of `integration_validated`, every
comparison of the stored `status` string, and the mix tasks — and both
explained why the two call sites left out of the diff are safe by
construction rather than by luck: their result goes straight into
`record_validation/2`, which is patched in full.

They also both caught a claim in this description that was simply false. It
said `:info` and `:error` were the only flash kinds in the codebase, and used
that to justify a different tone on each of the two forms. `put_flash` across
`lib/` is 172 `:error`, 8 `:info`, **2 `:warning`**, and `flash.ex` declares
`:warning` a first-class kind with its own colour and icon. The practical
effect was that "not tested" rendered indistinguishably from success on the
personal form — losing the one distinction this PR is about. Both forms now
use the same warning tone, and a test pins it rather than trusting the
change to stay.

**Confirmed by running, not by reading:**

- reverting only the fix turns 12 of 108 tests red; restoring it returns
  108 green, with no `Excluding tags` or database-connection notice in the
  output;
- the five patched sites are independently necessary — breaking only the
  internal `case` fails exactly the two post-save tests while both dry-run
  tests pass, and breaking only the four UI sites fails all four, each on its
  own line. Both reviewers reproduced these numbers separately and they
  matched;
- the thirteen providers with a real check are untouched: 43 doctests,
  4078 tests, 0 failures.

**Stated as unverified:** the `alert-warning` styling is asserted in the
rendered HTML, not in a browser, and the stale `@spec` on
`broadcast_validated/3` was confirmed by reading plus a clean dialyzer run
rather than by a failing test — success typing is wider than the declaration,
so it never gated CI.

PR #755, opened 2026-08-26
Add doctor check for belongs_to relations with no database FK

timujinne:
## What

A thirteenth doctor check: **Schema-Declared Relations Without a DB FK** —
every `belongs_to` the Ecto schemas declare, whose `(table, owner_key,
ref_table)` triple has no matching constraint in `pg_constraint`.

## Why here, and why now

`fk_validation_state/5` already answers "does this specific expected
relationship have a declared FK at all", and the comment above it names
exactly what was missing:

> Kept anyway, deliberately: it is the one existing primitive that can answer
> "does this specific expected relationship have a declared FK at all",
> independent of naming — exactly what closing check 12's current gap … would
> need, given a source of expected `(table, fk_col, ref_table)` candidates.
> **No such source exists yet, so this stays unwired until one does.**

The schemas are that source. Every `belongs_to` names its triple
unambiguously via `owner_key`/`related_key`, so the candidate list can be
derived rather than hardcoded.

`fk_validation_state/5` itself is **not** called: it answers one relationship
per round trip, and check 12 already shows what that costs (one query per
constraint — 231 on a calibration install). This check instead asks the same
question in bulk, two catalog queries regardless of how many associations
exist, comparing the same full shape the primitive compares: owning table,
column, referenced table, and single-column key. The primitive's comment stays
accurate for its own call site; this is a second reader of the same catalogs,
not a replacement.

Check 12 covers relations that *have* a constraint; this one covers the
relations that never had one — the case check 12's own moduledoc puts outside
its scope.

## Advisory by design

A declared relation without a database FK is a real gap but not automatically
a defect: some are deliberate. So the check reports `:warn` with the count and
the offending relations, and never `:fail` — including when a catalog query
itself fails, which is reported as `:warn` with the reason rather than being
allowed to crash into `:fail`. On the installations available to us it names
`phoenix_kit_activities.actor_uuid` and `.target_uuid`.

Zero coverage is never reported as clean. If no declared relation could be
checked at all — wrong `--prefix`, a host before its first migration, an
interrupted update — the result is `:warn` naming that explicitly, in the
same spirit as check 12's own `coverage is zero, which is not the same as
clean`. A `:pass` from this check states how many relations it actually
examined.

## The destructive test move

`ADD CONSTRAINT … FOREIGN KEY` takes a `ShareRowExclusiveLock` on **both**
tables. In an `async: true` module that deadlocks (`40P01`) against any other
async test file writing to `phoenix_kit_users` — observed on a real full-suite
run, and reproducible only there: the file passes 34/34 in isolation, which is
what makes it easy to miss.

So the destructive tests move into their own `async: false` module
(`DoctorOrphanedFkDestructiveTest`); ExUnit runs sync modules after the async
ones, so the collision cannot happen. The existing destructive orphan test
carries the same hazard, so it moves with them rather than being left behind
in the async file. A comment at the old site records why, so the tests cannot
quietly migrate back.

This part is separable if you would rather take it on its own — it is one
commit's worth of change and touches no production code.

## Verification

- Full suite on a tree carrying this change: **43 doctests, 4102 tests, 0
  failures**, with integration tests actually running (no
  `Excluding tags: [:integration]` line).
- Before the async split, that same full run reproduced the deadlock; after
  it, two consecutive full runs were clean.
- `mix format --check-formatted`, `mix credo --strict` and `mix dialyzer` all
  clean via the repository's own pre-commit gate.

## Not covered

- Not run against a real named-schema (`prefix:`) installation. The
  zero-coverage path that a wrong prefix produces *is* covered by a test.
- The check reads schemas reachable from the loaded application; a module not
  loaded at doctor time contributes no candidates. The zero-coverage branch
  catches the total case; a partially-loaded application is still counted as
  examined.

### DIFF
--- a/lib/mix/tasks/phoenix_kit.doctor.ex
+++ b/lib/mix/tasks/phoenix_kit.doctor.ex
@@ -1390,12 +1390,15 @@ defmodule Mix.Tasks.PhoenixKit.Doctor do
   # Check 12 above only ever examines a relationship that already HAS a
   # declared FK constraint — its own moduledoc entry names this boundary
   # ("a relationship with no FK constraint declared at all is outside this
-  # check's scope"), and `fk_validation_state/5`'s own comment names the
-  # missing piece: "given a source of expected (table, fk_col, ref_table)
-  # candidates. No such source exists yet". `belongs_to` IS that source —
-  # every one PhoenixKit's own Ecto schemas declare names its (table,
-  # owner_key, ref_table) unambiguously via `owner_key`/`related_key`,
-  # without guessing from a column name.
+  # check's scope"). `belongs_to` is a source of exactly the (table,
+  # owner_key, ref_table) triple a check like that would need — every one
+  # PhoenixKit's own Ecto schemas declare names its target unambiguously
+  # via `owner_key`/`related`, without guessing from a column name. This
+  # check resolves that triple itself and cross-references it against
+  # `pg_constraint` directly (a bulk, two-query scan — see
+  # `discover_schema_declared_relations_without_fk/2` — not a per-candidate
+  # call to `fk_validation_state/5`, which answers a related but different
+  # question, a constraint's VALIDATE state, that this check does not use).
   #
   # It found two real gaps on a live install: phoenix_kit_activities's
   # actor_uuid and target_uuid, both declared in PhoenixKit.Activity.Entry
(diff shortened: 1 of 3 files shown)


timujinne (commented):
Two independent reviews of 678b5fa307, both landing on "fix before merge".
Every finding below was reproduced before it was written down. Findings 1-3
are being fixed in this PR; this comment records what was wrong and how it was
established.

## 1. `:pass` can mean "found nothing" or "looked at nothing" — the same text

`doctor.ex:1420-1422` returns
`{:pass, "Every belongs_to PhoenixKit's schemas declare has a matching DB
foreign key"}` whenever the result list is empty. But the candidate list is
filtered against `existing_columns` (`:1485`), which is built from
`information_schema.columns` for the given prefix. If that prefix holds no
tables — a wrong `--prefix`, a host before its first migration, an interrupted
update — every candidate is filtered out and the check prints that same
universally-quantified PASS having examined nothing.

Reproduced:

    check_schema_declared_relations_without_fk("public")
      -> {:warn, "2 relation(s) ... activities.actor_uuid ... target_uuid"}
    check_schema_declared_relations_without_fk("nonexistent_schema_zzz")
      -> {:pass, "Every belongs_to ... has a matching DB foreign key"}

The second is byte-identical to a genuinely clean install.

This is the exact failure 2.13.10 fixed for check 12, whose own comment
(`:865-882`) names it: *"clean" and "never looked" used to print byte-identical
text*. Check 12 has a `coverage is zero, which is not the same as clean` branch
(`:892-903`); this check has none. In the same report, on the same database
state, the two checks would contradict each other line by line.

## 2. The code compares a pair; the text claims a triple

`belongs_to_owner_columns/1` (`:1490-1497`) returns `{table, owner_key}`, and
the `pg_constraint` query (`:1449-1463`) selects `t.relname, a.attname` — the
owning table and column only. No `confrelid`, no join to the referenced table,
no single-column restriction (`attnum = ANY(c.conkey)` also matches composite
keys).

So "covered" means *this column carries some foreign key, to anywhere*. A
`belongs_to` whose column points at the wrong table reads as satisfied.
Demonstrated against a live database:

    ALTER TABLE phoenix_kit_activities ADD CONSTRAINT tmp_fk
      FOREIGN KEY (actor_uuid) REFERENCES phoenix_kit_user_roles(uuid) NOT VALID;
    -- discover_... now returns only [{"phoenix_kit_activities","target_uuid"}]
    -- actor_uuid disappeared, though its belongs_to points at users
    ALTER TABLE phoenix_kit_activities DROP CONSTRAINT tmp_fk;   -- list restored

The commit message, this PR's description, the moduledoc entry and the comment
at `:1396-1398` all say `(table, fk_col, ref_table)`. Four places claim more
than the code checks.

## 3. A catalog error turns an advisory check into `:fail`

`:1449` and `:1465` match `{:ok, %{rows: _}} = repo.query(...)` with no
`{:error, reason}` branch. A server-side error — a `statement_timeout`
cancellation, for instance — fails the match, `run_check`'s rescue converts it
to `{:fail, "Exception: ..."}`, and that can flip `--exit-code` to 1. The PR
description says in bold that this check **never** returns `:fail`. Check 12
handles the same catalogs with an explicit `{:error, reason}` → `:warn`
(`:892-897`).

## 4. Nitpick: the tests assert membership, not the exact list

`phoenix_kit_doctor_orphaned_fk_test.exs:143-152` and `:168-171` use `in` and
`=~`. On the canonical chain the result is deterministic (exactly
`activities.actor_uuid` and `.target_uuid`), so an exact-list assertion would
force a future deliberate `belongs_to`-without-FK to be acknowledged rather
than absorbed silently.

## Correction to this PR's description

The description claims this "wires the primitive up rather than adding a
parallel mechanism beside it". That is wrong: `fk_validation_state/5` is never
called — `grep` finds only its definition and two comments. A separate bulk
query was written instead, and it is precisely the primitive's full-shape check
(`ft.relname = ref_table`, single-column `conkey`) that finding 2 is missing.
The description will be corrected along with the fixes.

## Verified clean

- Nothing of upstream's was dropped from `doctor.ex`: all 14 removed lines are
  the moduledoc renumbering 13-26 → 14-27, and `run_check` order matches.
- The existing destructive orphan test moved verbatim into the new module; the
  old file keeps the rest of its coverage.
- The `async: false` reasoning holds against ExUnit's runner: async modules are
  exhausted before sync ones start, so the DDL/writer collision cannot occur.
- Cost is two catalog queries per run regardless of schema count.
- No false positives on the canonical chain: the full result is exactly the two
  known relations. Non-default `owner_key`/`related_key`, `:through` and
  `has_many` are handled correctly.

timujinne:
All four findings above are fixed in 2ae58f2298, each re-verified by re-running
the reproduction rather than by reading the patch:

- **Zero coverage no longer reads as clean.** `check_schema_declared_relations_without_fk("nonexistent_schema_zzz")`
  now returns `:warn` — "coverage is zero: either PhoenixKit's own schemas'
  tables aren't installed here…" — where it previously returned the same
  `:pass` text a genuinely clean install produces. A real `:pass` now states
  how many relations it examined.
- **The triple is compared.** The catalog query joins
  `pg_class ft ON ft.oid = c.confrelid`, selects `ft.relname AS ref_table`, and
  reads `array_length(c.conkey, 1)`, so a column whose foreign key points at a
  different table than its `belongs_to` no longer counts as covered, and
  composite keys are distinguished from single-column ones. A test covers the
  wrong-target case; reducing the comparison back to `{table, column}` makes it
  fail.
- **Catalog errors are `:warn`, not `:fail`.** Both queries now have
  `{:error, reason}` branches. The previous hard match crashed on a malformed
  prefix and surfaced as `:fail`, which contradicted this check's own contract
  and could flip `--exit-code`.
- **Tests assert the exact list** rather than membership.

Verification on the branch: `43 doctests, 4076 tests, 0 failures`, with
integration tests actually running (no `Excluding tags: [:integration]` line).
`mix format --check-formatted`, `mix credo --strict` and `mix dialyzer` clean
through the repository's own pre-commit gate.

The description has been corrected too: it previously claimed this change wires
`fk_validation_state/5` up, which it does not — the reasoning for the bulk
query is now stated honestly there.

PR #756, opened 2026-08-27
Add a key store so a rotated integrations secret is not lost

timujinne:
## What

`mix phoenix_kit.integrations.rotate_key` printed a freshly generated secret
exactly once and saved it nowhere. An operator who lost that line lost every
stored integration credential with it: the ciphertext stays in the database and
nothing can read it again.

This adds a place for the secret to land, as a behaviour rather than a fixed
mechanism:

- `PhoenixKit.Integrations.KeyStore` — the extension point.
- `KeyStore.File` — the default: one file, mode `0600`, outside the repository.
  No external service, no account, no credentials of its own, so it works on any
  host. A cloud secrets manager does not.
- `KeyStore.S3` — optional off-host copy (S3-compatible: AWS, R2, B2, Tigris,
  MinIO).
- `KeyStore.Chain` — keeps the secret in more than one place at once, which is
  what makes a remote store an *addition* rather than a foundation.

Unconfigured stays a supported state, not a broken one: rotation keeps its
print-once behaviour and says plainly that the secret was saved nowhere.

Alongside it, `Integrations.Encryption` gains a key fingerprint (so a shared key
is visible rather than silently shared), and the doctor's key check and the
integrations settings page stop advising a rotation the rotation module itself
would refuse.

## Relationship to `Integrations.KeyRotation`

This does not touch `key_rotation.ex` and does not duplicate it. That module
re-encrypts stored rows under a new secret; this one is about the secret not
being lost in the first place. The two halves are independent — before this,
main had "re-encrypt" and no "don't lose it".

## Verification

Against current `main` (2.13.12), in a clean worktree:

- `git merge` — no conflicts. `main` has not touched `encryption.ex` since this
  work branched, so the two lines of development do not overlap.
- `mix compile --warnings-as-errors` — clean.
- Full suite — `43 doctests, 4180 tests, 0 failures`, with integration tests
  actually running (no `Excluding tags: [:integration]` line). The branch itself
  contributes ~2000 lines of tests, including a 763-line `key_store_test.exs`
  and a 413-line invariants support module.

## Two things the reviewer should know

**This branch carries dev_docs for two unrelated investigations.** Besides the
key-store documents, it includes ~1000 lines on a pre-commit hook
misattribution and ~130 lines of a P012 review round. They ride along because
the work happened on the same branch, not because they belong to this change.
Say the word and they come out — the code is untouched either way.

**Rotation has been exercised on one installation, not two.** It was run
24 Aug on one host, with the key landing in a `0600` file as intended; the other
installation has not run it. So the file store has one real-world execution
behind it, and the S3 and chain adapters have their tests and no production
mileage.


timujinne (commented):
Two independent reviews of b6a1233445. One landed on ship, one on fix; the
findings below are the union, and every one was reproduced before it was
written down. Nothing here is a leak or a data-loss path — both reviews looked
for those specifically and found none, and neither did the checks run on top.

## Verified clean, so it is on record

- **No secret reaches a log.** The two `Logger` call sites in the new code print
  module names and a parsed reason (`chain.ex:102`) and a path plus permission
  bits (`file.ex:297`). A sweep of every added `Logger` line in the branch for an
  interpolated secret returns nothing.
- **Chain's partial failure is an honest failure.** `chain.ex:121` returns
  `{:error, {:chain_write_failed, failures}}`, and the test at
  `key_store_test.exs:629` exercises it. The comment above it — "A backup that
  quietly stopped being written is not a backup" — is the right instinct.
- **The file store tightens before it writes.** `file.ex:114-127` documents three
  load-bearing properties and the code matches: the temp file is created empty
  and chmod'ed `0600` *before* the secret goes into it, `:file.sync/1` covers
  file and directory around the rename, and a rename replaces a symlink with a
  regular file.
- Full suite on current `main`: `43 doctests, 4180 tests, 0 failures`, with
  integration tests actually running. `mix compile --warnings-as-errors` clean.

## 1. `ensure_dir/1` narrows a directory the store does not own

`file.ex:180-193`. After `mkdir_p`, `File.chmod(dir, @dir_mode)` runs
unconditionally — on a pre-existing directory too. The comment beside it reasons
only about the chmod *failing* ("a pre-existing directory may be owned by
someone else"), not about it *succeeding* on someone else's directory.

Reproduced:

    directory before: 755
    directory after a single key write: 700

With the moduledoc's own example path (`/etc/phoenix_kit/app.key`,
`key_store.ex:26`) on a host where that directory already exists and is shared,
a key write silently re-permissions it. With a path directly in a shared
directory the blast radius is that directory.

Suggested direction: tighten only a directory this write created (`mkdir_p`
does not report creation — a stat before and after answers it), and leave a
pre-existing one alone.

## 2. The store-failure message contradicts what the store holds

`rotate_key.ex:395-400` (`store_failed_message`) with `:350-358`
(`print_secret_of_last_resort`). On a Chain partial failure — the local file
written, the remote member down, which is the exact case Chain exists for — the
task prints "This line is now the only copy" and "… does not hold it", and
advises setting `integrations_encryption_key` directly.

Reproduced with a chain of `KeyStore.File` plus a store that always fails:

    write_verified => {:error, {:chain_write_failed, [{BoomStore, {:error, :network_down}}]}}
    file holds the secret => true

So the file does hold it, and the message says nothing does. The advice compounds
it: setting the key in config permanently shadows the store, which this same task
warns about elsewhere (`warn_if_explicit_key_shadows_store`, `:370-387`).

No data is lost on any branch of this — the defect is the message. But the class
is exactly the one this branch exists to fix.

Suggested direction: on `{:chain_write_failed, failures}`, name the members that
*did* write (they are absent from `failures`) and drop "the only copy" for that
case.

## 3. A shadowed store is invisible on the settings page for two of three tiers

`encryption.ex:392-397` computes `{:shadowed, location}` whenever the store holds
a secret different from the one in use. `key_report/1` has clauses for
`{:dedicated, store: {:shadowed, _}}` (`:508`) but none for `:legacy` or `:none`
with the same store state — those fall through to the generic tier clauses
(`:566`, `:613`), which never mention the store. The doctor is unaffected: it
renders the store as a separate layer (`doctor.ex:316`, `key_store_lines/1`).

This is reachable on the ordinary adoption path, not only by a race. After
`rotate_key` stores a new dedicated secret, the app keeps running on the old key
until the operator wires `integrations_encryption_key` and restarts — the task
says so itself. In that window the state is exactly `tier: :legacy` plus
`store: {:shadowed, _}`, and the settings page stays silent about the store
holding a different secret. That is the window where the information matters
most.

## 4. "The value above" can refer to nothing

`rotate_key.ex:315-344`. When the operator supplied `--new-key`, the block that
prints the secret is skipped (`unless supplied?`), but the trailing block still
says "with the value above wired to that environment variable". Nothing was
printed above in that run. Not a leak; misleading at the moment the operator can
least afford it. Not covered by a test.

## 5. Nitpick: the S3 preflight discards its cleanup result

`s3.ex:100-102`. The probe object is deleted and the result of the delete is not
matched, so a failed cleanup leaves an empty object behind and the preflight
still reports success.

## Not checked

- `mix credo --strict` and `mix dialyzer` were not run on this branch.
- No real S3 endpoint was exercised — only a bucket-not-configured path and a
  stub that always fails.
- Rotation has one real-world execution behind it (24 Aug, on one installation,
  key landing in a `0600` file). The S3 and chain adapters have tests and no
  production mileage.

timujinne:
All five findings are fixed in 18d5f3218f. Each was re-checked by re-running the
reproduction rather than by reading the patch.

**1 — the store no longer re-permissions a directory it did not create.**
`ensure_dir/1` now tightens only a directory this write created and leaves a
pre-existing one alone:

    pre-existing directory, before write: 755   after write: 755   (was 700)
    directory created by the write:                        700

So the tightening that mattered is kept and the side effect is gone.

**2 — the store-failure message no longer contradicts the store.**
On `{:chain_write_failed, failures}` the task now derives the members that *did*
write (they are absent from `failures`) and names them, instead of calling the
printed secret "the only copy". A chain with no survivors is still told it holds
the only copy — that case is covered by its own test, so the correction does not
overshoot.

**3 — a shadowed store is now reported for every tier.**
`key_report/1` gained explicit clauses for `{:legacy, store: {:shadowed, _}}`
(`encryption.ex:577`) and `{:none, …}` (`:648`), and the settings page gained
`{_status, :store_shadowed}` fallbacks for title and detail — so a tier added
later is covered rather than silently omitted. This matters on the ordinary
adoption path: between `rotate_key` storing a new secret and the operator wiring
it and restarting, the state is exactly legacy-plus-shadowed.

**4 — "the value above" now refers to something.** When `--new-key` was
supplied, the message names the supplied secret instead of pointing at a line
that was never printed.

**5 — the S3 preflight checks its cleanup**, reporting a probe that could not be
removed instead of passing regardless.

## Verification

Full suite on the branch: `43 doctests, 4188 tests, 0 failures`, integration
tests actually running. The `+8` over the pre-fix run are the tests added with
these fixes. The three affected files run green in isolation as well
(141 tests).

## Still outstanding, unchanged from the review above

`mix credo --strict` and `mix dialyzer` have not been run on this branch; no
real S3 endpoint has been exercised; the S3 and chain adapters have their tests
and no production mileage, while the file store has one real rotation behind it.

timujinne:
One correction to the "Not checked" list in the review above: two of those items
are no longer true. Both tools were run on the branch at `18d5f3218f`:

    mix credo --strict   11725 mods/funs, found no issues.
    mix dialyzer         Total errors: 241, Skipped: 241, Unnecessary Skips: 2
                         done (passed successfully)

The 241 skipped are this repository's existing `.dialyzer_ignore.exs` baseline,
identical on `main` — the number that matters is that nothing unsuppressed came
from this branch. The two unnecessary skips are pre-existing entries that no
longer match anything; they are not this branch's doing and are left alone.

The rest of that list stands unchanged and is not being quietly retired:

- no real S3 endpoint has been exercised — only a "bucket not configured" path
  and a stub that always fails;
- the S3 and chain adapters have their tests and no production mileage;
- rotation has one real-world execution behind it (24 Aug, one installation,
  the key landing in a `0600` file).

PR #758, opened 2026-08-28
Verify config.exs splices parse and land before adopting them

timujinne:
## What

The installer edits a host's `config.exs` by regular expression. A regex cannot
tell Elixir code from a `#` comment or a string literal, so a `]` written inside
a crontab comment ends the list as far as the pattern is concerned, and the
splice lands inside the comment. The host's next compile then fails with

    ** (MismatchedDelimiterError) ... unexpected token: ]

on a file the installer reported as successfully updated.

This adds a check after every such splice: parse the candidate before adopting
it, and if it does not parse, roll back to the original and print the block for
the operator to paste manually.

## Why the parse alone is not enough

An insertion into a comment that happens to contain no `]` produces a file that
parses perfectly and silently lacks the entry. So the check is not "does it
parse" but "does it parse **and** is the entry actually there": `ConfigVerify`
walks the resulting AST for the key that was supposed to be added, and a
mismatch rolls back exactly like a syntax error does.

## Scope

Fourteen call sites across `oban_config.ex`, `phoenix_kit.update.ex`,
`boot_hook.ex` and `runtime_detector.ex` — every place that splices into an
existing config structure by pattern.

`js_integration.ex` is deliberately **not** touched. Its splices are into
JS/HEEX, where the Elixir parser does not apply and no equivalent checker is
available here. Guarding them with something unverifiable would trade a real
check for the appearance of one, so they keep their current behaviour and stay
a known gap rather than a covered-looking one.

## Verification

Against `main` (2.13.12), driving the real entry point rather than the internals:

    crontab carrying "# daily digest [see runbook]"
      -> file untouched, entry not inserted, manual-instruction notice printed
         (previously: MismatchedDelimiterError)

    crontab entry with a nested list, args: %{tags: ["a", "b"]}
      -> file untouched, entry did not land inside the nested list
         (this is the case a parse-only check would pass)

    the same config with neither trap
      -> entry inserted, result parses

Full suite: `43 doctests, 4123 tests, 0 failures`, with integration tests
actually running. `mix compile --warnings-as-errors` clean.

## Also in here

One independent bug in `add_scheduled_posts_job_to_crontab/1`, found while
reproducing the above: the appended comma was glued to trailing whitespace
rather than to the last real token, which produced `syntax error before: ','`
on inputs that had nothing to do with comments. Fixed to match the pattern the
anchored siblings in the same module already used.

### DIFF
--- a/lib/mix/tasks/phoenix_kit.update.ex
+++ b/lib/mix/tasks/phoenix_kit.update.ex
@@ -1563,7 +1563,7 @@ if Code.ensure_loaded?(Igniter.Mix.Task) do
           # Case 1: Incorrect configuration with providers: []
           String.contains?(content, "config :ueberauth, Ueberauth") &&
               Regex.match?(~r/providers:\s*\[\s*\]/, content) ->
-            fix_ueberauth_providers_config(igniter, content)
+            fix_ueberauth_providers_config(igniter)
 
           # Case 2: Configuration exists and is correct (providers: %{} or with values)
           String.contains?(content, "config :ueberauth, Ueberauth") ->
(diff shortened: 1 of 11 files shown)


timujinne (commented):
Two independent reviews of this branch, both landing on "fix before merge", with
almost no overlap between them. Everything below was reproduced before it was
written down.

## Blocking: the change loses edits it was written to protect

**`boot_hook.ex:76, 89-91` — a buffered edit is silently discarded.**

Upstream read the file's *buffered* content inside the `Igniter.update_file/3`
callback and rewrote it in place. This branch moves the read out of the callback
— `content = File.read!(app_file)` — so the verify/rollback decision can be made
before calling `update_file`, and then writes that disk-derived candidate over
the buffered source wholesale.

Igniter does not flush between steps. The step immediately before this one,
`ApplicationSupervisor.add_supervisor()`, has buffered the addition of
`PhoenixKit.Supervisor` to `children`. That edit is thrown away.

Reproduced against Igniter directly:

    supervisor survived: false
    boot pipe present:   true

On a fresh install — the exact state where both steps fire — the host ends up
with the boot pipe and **without** `PhoenixKit.Supervisor`, and the installer
reports success. The supervision tree never starts.

The diff shows this is a regression of this branch, not a pre-existing problem:

    - Rewrite.Source.get(source, :content)     # upstream: buffered
    + File.read!(app_file)                     # here: disk

**`phoenix_kit.update.ex:1593-1616` — the same pattern, same root cause.**
`fix_ueberauth_providers_config/2` builds its candidate from disk content and
writes it over the buffered `config/config.exs`. Earlier in the same run,
`BasicConfiguration.add_basic_config` (`:145`) and
`PrefixConfig.add_prefix_configuration` (`:379`) may have buffered writes to that
file. A lost `prefix:` backfill sends the next `update`/`status` looking for the
installation in `public` — the failure `prefix_config.ex:68-75` exists to warn
about. Same diff signature: `- content = Rewrite.Source.get(source, :content)`.

The irony is worth stating plainly, because it points at the fix: a change whose
purpose is to stop config edits from being silently lost introduces two paths
that silently lose config edits. The cause is the design decision to move the
verify/rollback decision *outside* the callback. Deciding inside it, on the
buffered content — as `oban_config.ex:270-276` and `mailer_config.ex:66-77`
already do — removes the whole class.

## Blocking: the semantic check is not anchored to the target application

`oban_config.ex:721` `plugins_contains_module?`, `:969`
`crontab_contains_module?`, `config_verify.ex:141` `keyword_list_satisfies?` all
search the whole file for a matching key rather than within the target
application's block. The `plugins:` and `crontab:` regex splices are not anchored
to `app_name` either; only `insert_queue` (`queues:`) is.

Reproduced with a neighbouring application's config placed first:

    config :some_lib,  plugins: [SomeLib.Plugin]
    config :my_app, Oban,  plugins: [Oban.Plugins.Pruner]

    result: success, no warning
    some_lib block: Lifeline PRESENT      <- written into the wrong application
    my_app block:   Lifeline absent       <- the target got nothing

The verification passed because it found the module *somewhere*. So the new
safety net does not merely miss this — it certifies it. This repository already
has a test named "never writes the queue into a neighbouring application's
config" (`oban_config_test.exs:755`), but only for `queues:`; there is no
equivalent for `plugins:` or `crontab:`.

## Major: a legal host config gets a false refusal

`oban_config.ex:1058` `add_worker_entries_to_crontab/3` and `:1118`
`add_digest_entries_to_crontab/3` add a leading comma unconditionally, without
the `has_trailing_comma` check that `add_scheduled_posts_job_to_crontab/1`
received in this very branch. A host crontab ending with a trailing comma before
`]` — valid Elixir — produces a double comma, the rollback fires, and the
operator is told the insertion "would have produced invalid or misplaced config".

Reproduced: input parses, output parses, file unchanged, entry not added. So the
PruneWorker and digest cron entries are silently not wired, and the message
blames the host's config rather than the splicer. `ensure_worker_cron_entries/2`
has no test of its own.

## Smaller

- `oban_config.ex:847` — the Case-1 worker rename is the one `Regex.replace` in
  these files left outside `verify_or_rollback`, and it runs without
  `global: false`. An identifier-for-identifier swap cannot break parsing, so
  this is low risk; the gap is that the exception is nowhere recorded.
- **The JS/HEEX boundary exists only in this PR's description.** Both reviews
  raised it independently. `js_integration.ex:354,375,489,497` splice into
  `app.js` and `root.html.heex` with no check, deliberately — the Elixir parser
  does not apply there. Nothing in the code says so, and a boundary that lives in
  a PR description disappears on merge. The same unstated boundary exists at
  `css_integration.ex:110-111`.
- Out of scope but the same class, for a follow-up rather than this branch:
  `phoenix_kit.update.ex:1830-1960` `reorder_supervisors` rebuilds the children
  list line-by-line with no parse check, treats any line trimming to `]` as the
  list end, and moves lines verbatim without normalising commas.

## Verified clean

Rollback restores the original content byte-for-byte and nothing reaches disk
before the verdict (checked across all five files). Idempotency holds — repeat
runs short-circuit before the splice. The MUTATION A/B tests genuinely fail when
the semantic check is removed rather than restating the implementation. Seven
realistic configs produced no false rollback of a correct splice. The comma fix
in `add_scheduled_posts_job_to_crontab/1` does not break the paths that worked.

## Not checked

Full `mix test` was run on the branch as a whole (`43 doctests, 4123 tests,
0 failures`) but findings 1 and 2 were established through Igniter's mechanics
and the order of steps, not by an end-to-end `mix phoenix_kit.install` run on a
scratch host. `mix dialyzer` was not run.

timujinne:
All findings from the review above are fixed in 08151a5c7c. Each was re-checked
by re-running the reproduction, not by reading the patch.

**The two lost-edit regressions are gone, and by removing the cause rather than
patching the symptom.** Both sites now take the verify/rollback decision *inside*
the `Igniter.update_file/3` callback, against `Rewrite.Source.get(source,
:content)` — never a fresh disk read. `boot_hook.ex:81` and
`fix_ueberauth_providers_config/1` both carry a comment explaining what a stale
disk read discarded and which earlier pipeline steps it discarded
(`add_supervisor`, `add_basic_config`, `add_prefix_configuration`), so the
reasoning survives the next reader. `fix_ueberauth_providers_config` was also
made public with `@doc false` to give the Igniter-level test a real seam — the
defect lives in what lands in the buffer, which only `Igniter.Test.test_project/1`
can reach.

**Splices and the semantic check are now scoped to the target application:**

    before:  some_lib block Lifeline PRESENT | my_app block absent
    after:   some_lib block false            | my_app block true

The neighbouring application's config is left untouched and the module lands
where it was aimed.

**A legal trailing comma no longer triggers a false refusal:**

    before:  rollback, entry not added, host blamed
    after:   entry added, no rollback, result parses

`ensure_worker_cron_entries/2` now has tests of its own; the neighbouring-
application test that existed only for `queues:` now has counterparts for
`plugins:` and `crontab:`.

**The JS/HEEX and CSS boundary is now recorded in the code**, at
`js_integration.ex` and `css_integration.ex`, instead of living only in this
description — so it reads as a deliberate exclusion rather than an oversight
once this is merged.

## Verification

Full suite on the branch: `43 doctests, 4144 tests, 0 failures`, integration
tests actually running. The four affected test files run green in isolation as
well (99 tests). `+21` tests over the previous run are the ones added with these
fixes.

## One thing deliberately left alone

`phoenix_kit.update.ex:1830-1960` `reorder_supervisors` has the same class of
weakness — it rebuilds the children list line by line with no parse check, and
treats any line trimming to `]` as the end of the list. It is pre-existing code,
untouched by this branch, so it is named here rather than changed under cover of
this PR.

PR #759, opened 2026-08-28
Encrypt restricted setting values at rest

timujinne:
## What

`PhoenixKit.Settings` stores values as plain text. Among them are login OAuth
secrets and AWS keys:

    oauth_google_client_secret     oauth_github_client_secret
    oauth_facebook_app_secret      aws_access_key_id
                                   aws_secret_access_key

Measured on two installations on 28 Aug (lengths and prefix flags only, values
never read): not one of them carries an encryption marker.

`PhoenixKit.Integrations.Encryption` already exists, but it covers the
integrations subsystem only — `grep -r "Encryption\|encrypt\|enc:v1"` over
`lib/phoenix_kit/settings/` returns nothing. This wires the existing primitive
into settings rather than adding a second mechanism: `encrypt_value/1` and
`decrypt_value/1` take a string, which is exactly the shape a flat setting has.

## The read path distinguishes three states, not two

This is the part that matters, and it is why the change is not a one-liner.
`decrypt_value/1` returns `{:ok, value}` for a value that carries no prefix —
deliberate backward compatibility. A caller that treats "decrypted" and "was
never encrypted" as the same thing will one day hand out ciphertext believing it
is plaintext. So the read path resolves to one of:

    prefixed and decrypted        -> the plaintext
    no prefix                     -> legacy value, returned as-is, but distinguishable
    prefixed and decryption FAILED -> an explicit error; the raw value is never returned

The third state is the one worth guarding: a secret that stopped decrypting is a
broken login, not a protected one, and silently returning the ciphertext would
turn a key problem into an authentication problem somewhere far away.

## Write path, and what is deliberately not done

New values for restricted keys are stored encrypted. **Existing values are left
untouched.** Migrating what is already in a database is a separate, deliberate
step against a live installation, and it belongs with the operator, not with
this change.

## Verification

- Four tests, one per read state plus one that fails if the read path ever
  returns ciphertext on a failed decrypt. The guard was checked in both
  directions — broken deliberately to confirm the test goes red, then restored.
- Round-trip: plaintext in, `enc:v1:…` in the database, plaintext back out.
- A test pinning that pre-existing plaintext values keep working.
- Full suite in the primary checkout: `43 doctests, 4141 tests, 0 failures`,
  integration tests actually running. `mix precommit` (compile, docs, credo
  --strict, dialyzer) clean.

One honest caveat about test runs in a **git worktree**: full-suite runs there
are unstable, with a moving target — one run failed `CommonUnreachableTest`, the
next failed five different tests around module discovery and route dedup. The
same files pass in isolation on this branch and on clean `main` in the same
worktree (67 tests, 0 failures), and a full-suite run on clean `main` in a
worktree shows the same sandbox-ownership errors. So the instability tracks the
environment, not this change — but it is named here rather than smoothed over.

## Not covered

- Existing stored values are not migrated; nothing on any installation is
  encrypted by this change.
- Login through a provider has not been exercised end to end — that needs real
  credentials, and the sensible order is migrate, then verify a login, then call
  it done.
- The subject is wider than the title: `aws_access_key_id` and
  `aws_secret_access_key` sit in the same unencrypted list and are covered by the
  same restricted-key list here.

### DIFF
--- a/lib/phoenix_kit/settings/setting.ex
+++ b/lib/phoenix_kit/settings/setting.ex
@@ -174,12 +174,19 @@ defmodule PhoenixKit.Settings.Setting do
   # `PhoenixKit.Settings.restricted_setting_keys/0` (`oauth_*_secret`,
   # `aws_access_key_id`/`aws_secret_access_key` — live credential material,
   # not just data an admin would rather keep quiet) is stored encrypted.
-  # `get_change/2`, not `get_field/2`: only a genuinely NEW value gets
-  # encrypted here — a changeset that leaves `:value` untouched (e.g.
-  # `update_setting_with_module/3` re-saving `module` alone) must not
-  # re-touch whatever is already stored, encrypted or not. Existing plaintext
-  # rows are deliberately left alone by this change — migrating them is a
-  # separate, live-database step, not implied by this changeset.
+  # `get_change/2`, not `get_field/2`: `get_change/2` is `nil` both when
+  # `attrs` never carried `:value` at all AND when it did but the submitted
+  # value equals what `changeset.data` already holds (Ecto's own cast
+  # diffing, not something this function decides) — either way, nothing
+  # here re-touches whatever is already stored, encrypted or not. (Every
+  # current caller happens to always pass `:value` — e.g.
+  # `update_setting_with_module/3` always includes it alongside `module` —
+  # so it is this second case, an unchanged resubmission, not an absent
+  # key, that `get_change/2` actually needs to catch today; a future caller
+  # that omits `:value` entirely is covered by the same check for free.)
+  # Existing plaintext rows are deliberately left alone by this change —
+  # migrating them is a separate, live-database step, not implied by this
+  # changeset.
   #
   # `Encryption.encrypt_value/1` already does the right thing for every edge
   # case this needs: nil/empty pass through unchanged (no ciphertext for
--- a/lib/phoenix_kit/settings/setting.ex
+++ b/lib/phoenix_kit/settings/setting.ex
@@ -50,6 +50,7 @@ defmodule PhoenixKit.Settings.Setting do
   use PhoenixKit.SchemaPrefix
   import Ecto.Changeset
 
+  alias PhoenixKit.Integrations.Encryption
   alias PhoenixKit.Users.Role
   alias PhoenixKit.Users.Roles
   alias PhoenixKit.Utils.Date, as: UtilsDate
@@ -149,6 +150,7 @@ defmodule PhoenixKit.Settings.Setting do
     |> validate_value_exclusivity()
     |> validate_length(:module, max: 255)
     |> unique_constraint(:key, name: :phoenix_kit_settings_key_uidx)
+    |> maybe_encrypt_restricted_value()
     |> maybe_set_timestamps()
   end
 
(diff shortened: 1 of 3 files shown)


ddon:
@timujinne still in progress?

timujinne (commented):
Two independent reviews of `79e31a04ed`, on different engines. They converged on
the same high-severity finding, which is fixed in `9f1bca9ab4`. Everything below
was confirmed before it was written down.

## The finding: the cache warmer bypassed decryption

`warm_cache_data/0` (`settings.ex:1942`) read `setting.value` directly — the one
multi-key read path this change did not convert. The other four go through
`decrypt_and_map_settings/1` (`:592`, `:1019`, `:1165`, `:2007`).

That function is the warmer for the `:settings` cache (`supervisor.ex:62-69`,
`sync_init: true`, 5-minute TTL), and the cache is a plain ETS passthrough
(`cache.ex:441-452`) with no notion of encryption. So on every application boot
the cache filled with raw ciphertext for the restricted keys, and
`get_setting_cached/2` / `get_settings_cached/2` — public, and described in their
own docstrings as the preferred way to read — returned `enc:v1:…` as the value.
No error, no log.

The change's own comment (`settings.ex:1293-1298`) states the intent it missed:

> Shared by every multi-key read path … so there is exactly one place that
> decides how a batch of raw rows becomes the map a caller reads — not one call
> site each, silently drifting apart.

A fifth call site had drifted apart, silently, in the same commit that wrote
that sentence.

Nothing in core reads a restricted key through the cache today — the OAuth login
path uses `get_oauth_credentials_direct/1` and the AWS code reads
`get_setting/1`, both uncached and both decrypting correctly — so this was not
reachable in production as shipped. It was reachable through the public API.

## The second finding, same root

The guard test that exists precisely to prove "the read path never returns raw
ciphertext" invalidated the cache before reading and then exercised only
`get_setting/1` and `get_settings_direct/1`. The hole above was therefore
invisible to the suite: the test proved the paths its author had in mind.

## Fixed

`warm_cache_data/0` now runs every restricted value through the same
`decrypt_and_map_settings/1`, so there is again exactly one place. The guard test
was extended to the cached path — it warms the cache the way boot does and then
asserts through `get_setting_cached/2`, and it goes red if the warmer is reverted
to a direct read.

Suite on the branch: 17 tests in the settings file, 0 failures (16 before, plus
the new guard). Full suite in the primary checkout: `43 doctests, 4141 tests,
0 failures`.

## Still not covered, unchanged from the description

Existing stored values are not migrated — nothing on any installation is
encrypted by this change. Login through a provider has not been exercised end to
end; the sensible order remains migrate, verify a login, then call it done.

timujinne:
It was in progress when you asked — it isn't any more. Ready for your review.

Sequence, so the two commits read in the right order: `79e31a04ed` is the change itself; two independent reviews on different engines converged on one high-severity finding, and `9f1bca9ab4` fixes it. The review comment above was written after that fix, not before it.

The finding is worth a second's attention even though it's closed: the cache warmer read `setting.value` directly and was the one multi-key read path the change hadn't converted, so on every boot the `:settings` cache filled with raw ciphertext and the cached getters returned `enc:v1:…` with no error and no log. The guard test that exists to prevent exactly that had invalidated the cache before reading, so it never saw the hole. It now warms the cache the way boot does and asserts through `get_setting_cached/2`, and it goes red if the warmer is reverted to a direct read.

Suite on the branch: 17 tests in the settings file, 0 failures. Full suite in the primary checkout: 43 doctests, 4141 tests, 0 failures.

Two things this change deliberately does not do, unchanged from the description: existing stored values are not migrated, so nothing on an existing installation becomes encrypted by merging this; and login through a provider has not been exercised end to end. The order we'd suggest is migrate, verify a login, then treat it as done.

PR #762, opened 2026-08-28
Fix repair reporting two composite-key PRIMARY KEYs as missing

timujinne:
`mix phoenix_kit.repair` and `mix phoenix_kit.doctor` have been reporting two
objects as `missing` on every healthy install:

    phoenix_kit_shop_category_slugs_pkey
    phoenix_kit_shop_product_slugs_pkey

Both are the composite PRIMARY KEY on the natural key `(lang, value)` added in
V171. Both exist. The report was wrong, and a real `repair` run would have
reported `:missing -> :repaired/"created"` while creating nothing.

## Root cause, two layers

1. **The manifest declared them as `kind: :index`.** On the catalog they are
   PRIMARY KEY constraints, not standalone indexes.

2. **Even a correct `kind: :index` entry could never have matched.**
   `@indexes_sql` deliberately excludes any index backing a p/u constraint on
   its own table — that exclusion is there to stop an FK-referenced unique index
   from being listed twice. So the `kind: :index` lookup was structurally unable
   to find these objects, for any entry of this shape, ever.

Fixing only the first layer would leave the second one armed for the next
composite-key PRIMARY KEY somebody declares.

## What changes

- Both manifest entries move to `kind: :constraint` / `class: :constraint` with
  a proper constraint-shape body.
- `Probe.indexes/2` now partitions the *same* SQL join into two buckets instead
  of discarding the second one — `constraint_backed_indexes` — and `lookup/2`
  falls back to that bucket when a `kind: :index` entry finds nothing in the
  primary one.

The fallback deliberately does **not** make `kind: :index` a legitimate way to
declare a new constraint-backed object: `Executor.create_action/2` still
dispatches on `class`, and `class: :index` would emit the wrong DDL. That
boundary is documented in the code rather than left implied.

## Acceptance

The double acceptance this needed is both halves, not just the first:

- `:missing` is gone for both objects.
- **The FK-referenced unique index is not listed twice again.** Index snapshot
  before the change: 627 entries. After: 627. Byte-identical diff.

Destructive check, both directions: reverting one manifest entry's `id`/`kind`
back to `index:`/`:index` turns `test/integration/v171_shop_slug_projection_test.exs`
and `test/integration/hand_declared_manifest_test.exs` red with messages that
name what broke. Restoring returns them to green.

One note for anyone reproducing that: the failure surfaces as
`KeyError: key :unique not found`, not `:columns`. `Differ.compare/3` dispatches
to the index clause as soon as `id`/`kind`/`class` are reverted, while the
revision body is still constraint-shaped; `reason_field(:unique, ...)` is the
first field in that pipe, so it raises before reaching `:columns`. Deterministic,
and independent of which of the two objects is reverted.

## Third commit

Docstring only. An earlier version of the `Probe.lookup/2` moduledoc claimed that
regenerating the manifest could reintroduce the wrong `kind` for a future
composite-key PRIMARY KEY. That claim is false and is removed.

Established by running the generator, not by reading it: the constraint-backed
exclusion entered `dev_docs/squash/generate_baseline.exs` in 992b09b336
(7 August), and V171 arrived in 958da2d2ab (14 August) — after. `git log -S`
places both bad entries in the V171 commit itself. They were hand-declared
against the pattern of older `kind: :index` neighbours while the generator was
already correct. The generator never produced them and will not bring them back.

## Verification

- Targeted runs green on the branch; destructive check run in both directions.
- `mix compile --warnings-as-errors`, `mix format --check-formatted`,
  `mix credo --strict` clean on the touched files.
- Full suite was run on the primary checkout, not in the worktree: a full
  `mix test` inside a secondary worktree is unstable here for reasons unrelated
  to this change (module-discovery tests see a different build), so it is not
  offered as evidence.
- Not verified by us: behaviour against a live install of this exact branch. The
  underlying fixes were verified against a live install when originally written;
  this branch is their rebased delivery.

## Not included, deliberately

- The remaining 26 stale manifest names (renamed objects) and the 21 objects on a
  lagging install: those are a different problem — the manifest and the tool are
  both right, the *install* is behind. Separate work, and it touches live schemas.

### DIFF
--- a/lib/phoenix_kit/migrations/expected_schema.ex
+++ b/lib/phoenix_kit/migrations/expected_schema.ex
@@ -27743,30 +27743,29 @@ defmodule PhoenixKit.Migrations.ExpectedSchema do
         backfill: nil
       },
       %{
-        id: "index:phoenix_kit_shop_product_slugs_pkey",
+        id: "constraint:phoenix_kit_shop_product_slugs.phoenix_kit_shop_product_slugs_pkey",
         owner: :core,
         check:
           {:catalog,
            %{
              name: "phoenix_kit_shop_product_slugs_pkey",
              table: "phoenix_kit_shop_product_slugs",
-             kind: :index
+             kind: :constraint
            }},
         create: nil,
         since: 171,
-        class: :index,
+        class: :constraint,
         revisions: [
           {171,
            %{
-             table: "phoenix_kit_shop_product_slugs",
-             keys: ["lang", "value"],
-             unique: true,
-             method: "btree",
-             definition:
-               "CREATE UNIQUE INDEX phoenix_kit_shop_product_slugs_pkey ON __SCHEMA__.phoenix_kit_shop_product_slugs USING btree (lang, value)",
-             predicate: nil,
-             opclasses: ["text_ops", "text_ops"],
-             name_template: nil
+             type: "p",
+             columns: ["lang", "value"],
+             definition: "PRIMARY KEY (lang, value)",
+             name_template: nil,
+             foreign_table: nil,
+             foreign_columns: nil,
+             on_delete: nil,
+             on_update: nil
            }}
         ],
         presence: :required,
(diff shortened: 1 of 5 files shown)


timujinne (commented):
## Independent review

Two reviewers examined this separately, without contact. Every claim below was re-checked before posting; nothing here rests on a reviewer's word alone.

### Verified by measurement

**The change is a genuine no-op on a healthy schema.** A DDL snapshot (indexes + constraints) taken before and after a live `Repair.repair/1` run is byte-identical: 1261 objects in, 1261 out, nothing altered. That is the claim this PR most needed to support, and it is supported by measurement rather than by reading.

**The duplication this fixes does not return.** The intersection of `indexes` and `constraint_backed_indexes` is empty, so the FK-referenced unique index is no longer reported twice.

**All 17 `verify/1` findings were checked**, not only the ones named in the description. No new finding appears and none disappears.

**The `class: :index` / `class: :constraint` boundary was read in code, not taken from the comment.** `executor.ex:81-83` dispatches `class: :index` to `ShapeSql.index_create/1` (`shape_sql.ex:90-106`), which requires `definition`; the fallback in `Probe.lookup/2:536` touches only the read path (presence detection) and does not reach `Executor.create_action/2`. The comment describes the code correctly.

**Base is current.** `upstream/main` and this PR's base are the same commit, `c8de8239ca`, with zero commits of difference — confirmed by a fresh fetch at review time.

**Tests:** the three affected files run clean on a migrated test database — 32 tests plus one doctest, no failures.

### One discrepancy, stated rather than smoothed over

The description reports **627**; one reviewer measured **630** in their environment. The difference comes from the state of the test database used, not from the code path under change. We are flagging it because the number appears in the description as if environment-independent, and it is not.

### Not verified by us

- `mix format --check-formatted`, `mix credo --strict` and `mix dialyzer` were not completed on the affected files.
- Behaviour against a real installation that has fallen behind — as opposed to a freshly migrated test database — was not exercised. This is the case the repair path exists for, so it is the most meaningful gap in our coverage.

We are stating these explicitly rather than letting silence imply they passed.

PR #763, opened 2026-08-28
Fix stale disk read in fix_supervisor_ordering during update

timujinne:
Follow-up to #758, which fixed one instance of "the auto-patcher reads the disk
while the edit lives in the Igniter buffer". A comment on that PR named
`reorder_supervisors` as carrying the same class of defect but left it untouched
so the fix would not be mixed with refactoring pre-existing code. This is that
suspicion, reproduced and fixed.

It is **not** withdrawn: the defect is real.

## What it does

`fix_supervisor_ordering/1` decides whether `PhoenixKit.Supervisor` needs to be
moved in the host's `application.ex`. It made that decision from
`File.exists?`/`File.read!` — the state on disk — while two earlier steps of the
same pipeline had already staged edits to that file in the Igniter buffer:

- `ApplicationSupervisor.add_supervisor/1`, unconditionally, at the start of
  `igniter/1`
- `ObanConfig.add_oban_supervisor/1`, immediately before this call

So the check ran against a stale snapshot. With `PhoenixKit.Supervisor` present
only in the buffer, the disk read found nothing, matched the trivial
`validate_supervisor_positions(repo, nil, nil, _) -> :correct` clause, and
skipped the reordering entirely. The buffer that then went to disk had
`PhoenixKit.Supervisor` **before** `Repo` — which crashes on boot, since the
supervisor reads `Settings` from the database before the repo is up.

Same class as #758, opposite consequence: there the disk read *overwrote*
somebody else's staged edit; here it *skips a needed fix* it cannot see.

## How it was established — probe, not code reading

Reading the code was not enough, and neither was the test harness used in #758:
`Igniter.Test.test_project/1` never touches a real disk, so under it
`File.exists?`/`File.read!` simply find nothing and the function becomes a silent
no-op *before* reaching the code under suspicion. A green test there would have
proved nothing.

The probe (`test/mix/tasks/phoenix_kit_update_supervisor_ordering_test.exs`) uses
a real `Igniter.new/0` bound to a real temporary directory:

1. a real `application.ex` on disk (Repo, Endpoint; no `PhoenixKit.Supervisor`,
   no Oban) — the pre-update state
2. an edit staged into the Igniter buffer, mimicking the real earlier step, that
   inserts `PhoenixKit.Supervisor` in the wrong position
3. `fix_supervisor_ordering/1` invoked directly (made public with `@doc false`,
   the same approach #758 used for `fix_ueberauth_providers_config/1`)

Before the fix, the probe fails exactly as predicted, and the resulting file has
the wrong order.

## The fix

`File.exists?`/`File.read!` replaced with `Igniter.exists?/2` plus loading the
file into the Igniter buffer and reading its content through
`Rewrite.Source.get/2` — so the check sees the same state the subsequent edit in
`fix_application_supervisor_order/3` acts on. That function was already correct;
only the decision feeding it was reading the wrong source.

## Acceptance, both directions

Order is asserted against the resulting file, not declared:

- **Before the fix** the probe is red and names the problem: expected
  `Repo -> PhoenixKit.Supervisor -> Endpoint`, and it prints the real (wrong)
  output file. A second test reports `application.ex was never loaded into the
  Igniter buffer — fix_supervisor_ordering/1 decided from a bare File.read!/1
  instead of the Igniter-tracked source.`
- **After the fix** all three new tests pass, together with the existing #758
  suite (`phoenix_kit_update_ueberauth_test.exs`, `boot_hook_test.exs`,
  `oban_config_test.exs`): 74/74.
- The third new test **also goes red before the fix**, and this description
  originally claimed the opposite — corrected after an independent review
  reproduced the mutation on the real pre-fix body. It fails on the guard at
  `supervisor_ordering_test.exs:94` (`application.ex was never loaded into the
  Igniter buffer`), because the old body never loaded the file at all on the
  `:correct` branch. The "stays green" claim holds only for a narrower mutation
  that reverts the read but keeps `include_existing_file`. The test is therefore
  a more sensitive detector than this description first gave it credit for, not
  a redundant one.

`mix precommit` in full (compile, docs, `credo --strict`, dialyzer with the PLT
built from scratch) passed on the commit.

## Scope

One commit, two files. Deliberately separate from #758: different object,
different branch, so each can be reviewed and reverted on its own.

## Same mechanism, elsewhere in this task — named, not fixed here

The defect is not one function's slip. The mechanism is general: Igniter does not
flush staged edits to disk between steps, so any `File.exists?`/`File.read!` in a
later step sees the pre-pipeline snapshot. An independent review of this branch
traced the other places it already applies. Listing them here rather than folding
them in, so the mechanism is on record even though this PR fixes one instance:

- `phoenix_kit.update.ex:1558-1559` — `validate_and_fix_ueberauth_config/1`: the
  outer **gate** still reads `config/config.exs` from disk to choose its branch.
  #758 made only the inner `fix_ueberauth_providers_config/1` (:1610) buffer-aware.
  Called at :382, after `add_basic_config` (:145) and `add_prefix_configuration`
  (:379) have both staged that file. Closest to live: on a first install
  `config/config.exs` may exist only in the buffer, so `File.exists?` returns
  false and the gate is skipped entirely.
- `rate_limiter_config.ex:43-45` — `hammer_config_exists?/1` takes the igniter and
  ignores it (`_igniter`), reading `config/config.exs` from disk; gate reached at
  :385, after the same two stagings.
- `oban_config.ex:1543-1547` — `oban_supervisor_exists?/1` reads
  `lib/<app>/application.ex` from disk; gate at :386-387 → :2050, after
  `add_supervisor` (:146) stages it. Latent today, same structure.
- `oban_config.ex:166-171` — `oban_config_exists?/1`, same shape for the Oban
  config stanza. Latent today.

Explicitly **not** instances, checked so the list is honest:
`update.ex:292 check_required_configuration` runs before the igniter pipeline
starts, where disk is the correct source. The host-file reads in
`js_integration.ex`, `daisyui.ex`, `mailer_config.ex` and `runtime_detector.ex`
were not traced for staging order and are offered as triage candidates, not as
confirmed instances.

One known nit left alone: the new test's moduledoc points at `BootHookTest`'s
moduledoc as explaining why a virtual project is unsuitable here; that moduledoc
argues for virtual projects rather than against them. Happy to correct on request.

### DIFF
--- a/lib/mix/tasks/phoenix_kit.update.ex
+++ b/lib/mix/tasks/phoenix_kit.update.ex
@@ -1698,12 +1698,41 @@ if Code.ensure_loaded?(Igniter.Mix.Task) do
     # Validate and add Oban configuration if missing
     # Fix supervisor ordering in application.ex to prevent startup crashes
     # Ensures correct order: Repo → PhoenixKit.Supervisor → Oban → Endpoint
-    defp fix_supervisor_ordering(igniter) do
+    #
+    # P013: this used to decide `:correct` / `:needs_fix` from a fresh
+    # `File.read!/1` of application.ex — the same class of bug fixed under
+    # I103 for `BootHook.add_boot_hook/1` and
+    # `fix_ueberauth_providers_config/1`. Igniter never flushes a step's
+    # writes to disk before the next step runs; on a real `mix
+    # phoenix_kit.update`, TWO earlier steps in the SAME pipeline stage
+    # `application.ex` edits before this one runs —
+    # `ApplicationSupervisor.add_supervisor/1` (unconditionally, at the top
+    # of `igniter/1`) and `ObanConfig.add_oban_supervisor/1` (right before
+    # this call, inside `perform_igniter_update/2`) — both via
+    # `Igniter.Project.Application.add_new_child/3`, which only ever
+    # touches the Igniter buffer. A disk read here is blind to both: it can
+    # see no `PhoenixKit.Supervisor`/`Oban` line at all, match
+    # `validate_supervisor_positions/4`'s trivial "nothing to check yet"
+    # clauses, and report `:correct` while the buffer — the content that
+    # actually reaches disk when this run finishes — sits misordered.
+    # `Igniter.exists?/2` + loading the file into the Igniter buffer before
+    # reading its content (`Rewrite.Source.get/2`) makes this check see the
+    # SAME up-to-date state the subsequent fix in
+    # `fix_application_supervisor_order/3` already reads from.
+    #
+    # Test seam: made public (`@doc false`) the same way I103 exposed
+    # `fix_ueberauth_providers_config/1` — this defect can only be
+    # exercised end-to-end (a real, non-test-mode Igniter + real temp
+    # disk), not via a pure-content unit test.
+    @doc false
+    @spec fix_supervisor_ordering(Igniter.t()) :: Igniter.t()
+    def fix_supervisor_ordering(igniter) do
       app_name = IgniterHelpers.get_parent_app_name(igniter)
       app_file = "lib/#{app_name}/application.ex"
 
-      if File.exists?(app_file) do
-        content = File.read!(app_file)
+      if Igniter.exists?(igniter, app_file) do
+        igniter = Igniter.include_existing_file(igniter, app_file)
+        content = igniter.rewrite |> Rewrite.source!(app_file) |> Rewrite.Source.get(:content)
 
         # Check current supervisor ordering
         case check_supervisor_order(content, app_name) do
--- a/test/mix/tasks/phoenix_kit_update_supervisor_ordering_test.exs
+++ b/test/mix/tasks/phoenix_kit_update_supervisor_ordering_test.exs
@@ -1,0 +1,188 @@
+defmodule Mix.Tasks.PhoenixKit.Update.SupervisorOrderingTest do
+  @moduledoc """
+  P013: `fix_supervisor_ordering/1` used to decide whether `application.ex`
+  needed reordering by reading the file straight off REAL disk
+  (`content = File.read!(app_file)`), then making its correct/needs-fix
+  decision from that snapshot. Igniter never flushes a step's writes to
+  disk before the next step runs — on a real `mix phoenix_kit.update` run,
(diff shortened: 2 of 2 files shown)


timujinne (commented):
## Independent review

### The defect is real, not an artefact of the test

This is the point such a PR most needs to establish, and it was established by running the old code rather than by reading it. A reviewer built a standalone probe containing a verbatim copy of the previous `File.exists?` / `File.read!` logic and drove it through a real `Igniter.new/0` — not a test harness. The stale read reproduces there.

The call ordering that causes it is visible in the source:

- `lib/mix/tasks/phoenix_kit.update.ex:146` — `ApplicationSupervisor.add_supervisor()` runs unconditionally at the start of `igniter/1`.
- `lib/mix/tasks/phoenix_kit.update.ex:388,392,2050` — `validate_and_add_oban_config/2` runs immediately before `fix_supervisor_ordering/1`, inside the same `perform_igniter_update/2`.
- `lib/phoenix_kit/install/application_supervisor.ex:54` and `lib/phoenix_kit/install/oban_config.ex:1527` both route through `Igniter.Project.Application.add_new_child/3`, which mutates only `igniter.rewrite` — no `File.write!` occurs anywhere along that path, confirmed by reading `do_add_child/4` in the vendored igniter source.

So by the time the old code read the file from disk, the edits it needed to see existed only in the in-memory rewrite buffer. Reading disk returned a file that was correct and stale at the same time.

### Scope and base

- Two files, +220/-3: the task and its test. `mix.exs` and `CHANGELOG.md` are untouched — confirmed by `git diff --stat`.
- The base is current: this PR's base and `upstream/main` are the same commit, `c8de8239ca`, verified by a fresh fetch at review time.
- The change does not duplicate work already merged; that was checked against the tree rather than inferred from dates.

### Same root elsewhere, not addressed here

`ObanConfig.oban_supervisor_exists?/1` reads from disk in the same shape and is reachable from the same update path. It is not touched by this PR, and we are not asking for it here — but it is the same defect, and it will resurface under the same conditions.

### Not verified by us

Behaviour against a real installation that has drifted from a freshly generated project was not exercised. That is the case this task exists to serve, so it remains the most meaningful gap in our coverage. We state it rather than let silence imply it passed.

PR #764, opened 2026-08-28
Add a repository-level guard against tests pointed at a live database

timujinne:
`config/test.exs` reads `PGDATABASE` verbatim when it is set — deliberately, and
the file documents why: it lets you point the suite at a database you already
have instead of one the test role is allowed to `CREATE`. That is the right
design for a managed or shared Postgres.

It also means the suite runs against whatever `PGDATABASE` happens to say. If a
shell exports it for development, `mix test` connects to the development
database and the Ecto sandbox proceeds to migrate and truncate it. `MIX_ENV=test`
does not help: the two variables are independent, and forcing one does not clear
the other.

This adds a guard that refuses that case **by name and before any connection is
attempted**, at `test/test_helper.exs:10` — after the `db_name` binding it needs
and before everything that touches a database: the `psql` reachability probe, the
repo, and ExUnit.

## What it refuses

Two layers:

1. A built-in suffix pattern — `_dev`, `_development`, `_prod`, `_production`,
   `_staging`, case-insensitive. This needs no configuration and covers any host
   following the ordinary convention.
2. An optional `PHOENIX_KIT_TEST_DB_DENYLIST` (comma-separated), empty by
   default, for a live database whose name the pattern cannot recognise.

## What it deliberately does not do

**It never requires a particular name.** A `_test` suffix requirement was
considered and rejected: it would refuse exactly the legitimate case
`config/test.exs` documents — an already-provisioned database on a role without
`CREATEDB`, which has no reason to carry a `_test` suffix. The guard only refuses
a name that looks like a non-test environment; anything else passes through
untouched, verified both as a unit assertion and as a live subprocess run with
`PGDATABASE=ci_runner_42` (exit 0, no error raised).

It also changes nothing about how `config/test.exs` resolves `PGDATABASE`.

## Residual risks, stated rather than hidden

Two, in opposite directions, both named in the module's moduledoc so a reader does
not mistake the pattern for exhaustive coverage:

- A host whose live database has no recognisable environment suffix — bare `acme`,
  `acme_main` — is not caught by the pattern and has to add that name to
  `PHOENIX_KIT_TEST_DB_DENYLIST` itself.
- The mirror case: a host whose *legitimate test* database happens to end in a
  live-looking suffix is refused, and the denylist only adds refusals — there is
  no override. The refusal is loud, before any connection, and names the reason,
  so this is a deliberate trade-off rather than an oversight; an allowlist
  counterpart is the obvious extension if anyone hits it.

## Why it lives in the repository

An external wrapper script can do the same check, and protects only the person
who remembers to use it. Inside the repository, a plain `mix test` is guarded for
everyone, including CI and a contributor running the suite for the first time.

## Acceptance

Mutation in both directions, run as real `mix test` subprocesses with
`PGHOST=127.0.0.1:1` so that even a broken guard could not physically reach a
real database while the guard itself is under test:

```
PGDATABASE=acme_production   exit 1, LiveDatabaseError raised at test_helper.exs:9,
                             refused before ExUnit started, no test executed
PGDATABASE=phoenix_kit_test  exit 0, 11 tests, 0 failures
guard suite                  18 tests, 0 failures
PGDATABASE=ci_runner_42      passes through, exit 0, no error
```

Reverse check — the `check!/1` call removed from `test_helper.exs`, tests left in
place:

```
naive repeat of the refusal case   exit 0, 11 tests, 0 failures — silently green
                                   with no protection in place
the wiring suite catches it        exit 2, 7 tests, 5 failures
call restored                      18 tests, 0 failures
```

The wiring test matters more than the unit tests here: an earlier version of this
guard was mutation-tested and showed that simply deleting the call is **not**
caught by the module's own unit tests — the pre-existing fallback in
`test_helper.exs` ("cannot reach the database") quietly excludes `:integration`
and exits 0. The wiring test runs `mix test` as a subprocess specifically so that
a removed call fails loudly instead of passing as a skipped integration suite.

## Not verified

`mix dialyzer` was not run against this branch.

### DIFF
--- a/test/live_database_guard_test.exs
+++ b/test/live_database_guard_test.exs
@@ -1,0 +1,101 @@
+defmodule PhoenixKit.Test.LiveDatabaseGuardTest do
+  @moduledoc """
+  Pure unit coverage for `check!/1`'s own decision — separate from
+  `LiveDatabaseGuardWiringTest`, which proves the module is actually
+  reachable from `test_helper.exs`'s real boot sequence, not just that its
+  logic is correct in isolation.
+  """
+  use ExUnit.Case, async: true
+
+  alias PhoenixKit.Test.LiveDatabaseGuard
+
+  describe "check!/1 — built-in suffix pattern (zero configuration)" do
+    test "raises for example database names ending in a dangerous suffix" do
+      for db <- ~w(phoenix_kit_dev decor_3d_print_dev phoenixkit_hello_world_dev) do
+        assert_raise LiveDatabaseGuard.LiveDatabaseError, ~r/#{db}/, fn ->
+          LiveDatabaseGuard.check!(db)
+        end
+      end
+    end
+
+    test "raises for further suffix examples with nothing hardcoded behind them" do
+      # These names illustrate the refusal is genuinely suffix-based —
+      # none of them is special-cased anywhere in the module.
+      for db <- ~w(acme_production shop_staging widgets_development other_app_prod) do
+        assert_raise LiveDatabaseGuard.LiveDatabaseError, fn ->
+          LiveDatabaseGuard.check!(db)
+        end
+      end
+    end
+
+    test "suffix match is case-insensitive" do
+      assert_raise LiveDatabaseGuard.LiveDatabaseError, fn ->
+        LiveDatabaseGuard.check!("Acme_PRODUCTION")
+      end
+    end
+
+    test "the raised message says WHY, not just which database" do
+      assert_raise LiveDatabaseGuard.LiveDatabaseError,
+                   ~r/non-test environment/,
+                   fn -> LiveDatabaseGuard.check!("phoenix_kit_dev") end
+    end
+  end
+
+  describe "check!/1 — a foreign host's legitimately different test database name" do
+    test "an arbitrary pre-provisioned scratch name (no CREATEDB workflow) passes through" do
+      # This is exactly the case config/test.exs's own PGDATABASE-honoring
+      # comment describes: "a database you already have instead of one
+      # this role is allowed to CREATE". Nothing about these names
+      # resembles a non-test environment, so none of them are refused.
+      for db <- ~w(ci_runner_42 pk_scratch_7 shared_ci_db build_42_scratch) do
+        assert :ok = LiveDatabaseGuard.check!(db)
+      end
+    end
+
+    test "passes an isolated test database name straight through" do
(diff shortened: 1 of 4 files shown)


timujinne (commented):
## Independent review

### What it guards, and why the shape matters

The guard refuses a test run whose `PGDATABASE` ends with a production-looking suffix. Two properties were verified rather than assumed:

- **It matches on an exact suffix, case-insensitively — not on a substring.** `dev_tools_scratch` and `phoenix_kit_dev_backup` both pass as `:ok`. A substring match would have rejected them, and that is precisely the failure mode that makes such guards get disabled a week after they land.
- **It hardcodes no names.** With no denylist configured it refuses nothing; an empty or unusual `PGDATABASE` is `:ok`.

### The mutation was run, not reasoned about — and isolated from connectivity

Both reviewers exercised both directions against a deliberately unreachable host (`127.0.0.1:1`):

- `PGDATABASE=ci_runner_42` → exit 0, 11 tests, no failures, no guard error. It passes through.
- `PGDATABASE=acme_production` → exit 1, the guard fires.

Pointing at an unreachable host is what makes this meaningful: the guard decides on the *name*, before any connection is attempted, so the outcome cannot be confused with a connection failure.

### It does not break the configuration this project actually relies on

`config/test.exs:9-19` documents the case where `PGDATABASE` must be taken literally — a role without `CREATEDB` against a managed or shared Postgres. That path still works; it was the main risk and it was checked directly rather than argued.

### Scope

Four files, +389/-2, all test-side: the guard module under `test/support/`, its two test files, and the wiring in `test_helper.exs`. `lib/` is untouched, and so are `mix.exs` and `CHANGELOG.md`.

### Not verified by us

The guard was exercised against unreachable and local endpoints, never against a genuine production database — deliberately, since that is the thing it exists to prevent us from reaching. So its behaviour is proven on the decision it makes, not on the disaster it averts. We state this rather than let silence imply a stronger test than was run.

timujinne:
**Addendum — the review above covers `364d42090c`.**

A further commit, `641205c2f6`, landed after that review was posted. It replaces two container-local database names in the guard's own test fixtures — one line in each of `test/live_database_guard_test.exs` and `test/live_database_guard_wiring_test.exs`.

The guard module and the `test_helper.exs` wiring are untouched by it, so the findings above still describe the code as it stands. We are noting it because a review is a statement about a revision, not about a branch name, and that revision has moved.

PR #766, opened 2026-08-28
Add coverage for rotating a still-plaintext restricted setting

timujinne:
`mix phoenix_kit.integrations.rotate_key` is the path that takes an existing
install's restricted settings and brings them under encryption. Every existing
rotation test seeds its fixtures **through the encrypting write path** — so the
suite covers rotating a value that was already encrypted, and never covers the
state a real install is actually in before it updates: a genuine plaintext value
written by an older version.

That is the road the data on a live install travelled. It had no test.

## What the test asserts

A restricted setting is written as raw plaintext, bypassing the encrypting write,
to reproduce the pre-#759 state. Then `KeyRotation.rotate/1` runs, and the test
checks two things:

1. the row is swept — `rotated: 1`, and the stored value now carries the
   `enc:v1:` marker;
2. **the value still works** — `Settings.get_oauth_credentials_direct(:google).client_secret`
   returns the original secret. That is the same function
   `OAuthConfig.configure_providers/0` feeds into the live Ueberauth
   configuration, so this is the read path a real login uses, not a
   marker-shaped proxy for it.

The second assertion is the point. Checking for the `enc:v1:` prefix would pass
just as happily on a value that had been encrypted into something unreadable.

## Mutation, both directions

```
with a simulated fault in decrypt_row/1   rotated: 0
                                          the plaintext row is silently skipped
                                          and stays plaintext for good

with the code as it stands                rotated: 1, marker present,
                                          and the credential reads back intact
```

No library change was needed: the current implementation is already correct. The
first direction was produced by temporarily faulting `decrypt_row/1` to confirm
the test actually detects the failure it claims to; that mutation was reverted
immediately and `git diff` on `key_rotation.ex` is empty.

The failure mode this guards against is quiet and permanent: a rotation that
reports success while leaving a plaintext credential untouched forever, because
it only considered rows it recognised as already encrypted.

## Out of scope, named

- Migrating rows already sitting in a production database as plaintext is a
  live-database operation for whoever operates that host, not something a core PR
  does. `rotate_key` handles it correctly — including with `--new-key` set to the
  currently active secret when only a first-time encryption sweep is wanted, with
  no real key change. Running it is the host operator's decision.
- A genuine end-to-end OAuth round trip (browser → provider → callback) is not
  exercised: that needs a live provider and real credentials. The test goes
  through the real function the login path calls instead, which is the closest
  verification available without one.

### DIFF
--- a/test/integration/integrations_key_rotation_test.exs
+++ b/test/integration/integrations_key_rotation_test.exs
@@ -81,6 +81,24 @@ defmodule PhoenixKit.Integration.KeyRotationTest do
     updated
   end
 
+  # Same idea as `write_plaintext_field!/3`, for a restricted
+  # `PhoenixKit.Settings` row's bare `value` string instead of an
+  # integration's `value_json` map. `Ecto.Changeset.change/2` — NOT
+  # `Setting.update_changeset/2` — is the point: the latter runs
+  # `maybe_encrypt_restricted_value/1` on every `:value` change, which would
+  # immediately re-encrypt whatever this writes and defeat the whole point
+  # of simulating a row saved before the encryption feature existed.
+  defp write_plaintext_setting_value!(key, value) do
+    {:ok, _} = Settings.update_setting(key, "placeholder-overwritten-below")
+
+    key
+    |> Queries.get_setting_by_key()
+    |> Ecto.Changeset.change(value: value)
+    |> Queries.update_setting()
+
+    :ok
+  end
+
   describe "rotate/2 — real run" do
     test "re-encrypts every connection under the new secret; the old key can no longer read it" do
       uuid1 = seed_connection("openrouter", "one", %{"api_key" => "sk-one"})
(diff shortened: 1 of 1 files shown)


timujinne (commented):
## Independent review

### The gap this covers was real

The PR's central claim — that existing rotation tests never reproduce a genuine pre-#759 plaintext value — was verified by reading the base tests rather than by trusting the description.

All three pre-existing tests in the restricted-values describe block (`integrations_key_rotation_test.exs:217, :244, :264`) seed through `Settings.update_setting/2` → `Setting.update_changeset` → `maybe_encrypt_restricted_value/1`. That is the encrypting write path, so none of them can produce the state this test needs. The one place a genuine plaintext restricted value is seeded on the base (`settings_test.exs`, "an existing plaintext value is read back unchanged") exercises only the read path and never calls rotation. The mix-task test performs no database seeding at all.

### One correction to the description

The description says *all* rotation tests seed through the encrypting path. That is slightly too broad: a plaintext rotation test does exist for the integrations `value_json` branch (`:166`).

The distinction matters because those are different branches of `decrypt_row` — `key_rotation.ex:353` for `value_json`, `:369` for restricted `value` strings. This PR's subject is the second, and for that branch the claim is exact. We are flagging it so the sentence is not read more widely than it holds.

### The second assertion exercises the production chain

It goes through `Settings.get_oauth_credentials_direct/1` → `get_settings_direct` → `decrypt_and_map_settings` — the same path `PhoenixKit.Users.OAuthConfig.configure_providers/0` uses when it places `client_secret` into the provider configuration. The test therefore covers the route a real login takes, not a convenient equivalent.

### Scope and base

- Exactly one file, `test/integration/integrations_key_rotation_test.exs`, +66/-0. `mix.exs`, `CHANGELOG.md` and `lib/` are untouched — the diff against `lib/` is empty, confirmed directly.
- Base is current: this branch and `upstream/main` are the same commit, `c8de8239ca`, verified by a fresh fetch at review time.
- The published diff matches the local diff exactly.

Production code is not modified; this adds coverage to behaviour that is already correct.

### Not verified by us

The test was run against a shared test database whose state is not isolated per run. It passed and does not depend on that state, but we did not exercise it against a database that has drifted the way a real installation can. We state this rather than let silence imply it was covered.

PR #773, opened 2026-08-30
Fix block-dangerous-git.sh fail-open on parse failure; allow --force-with-lease

### LINKED ISSUE #768: block-dangerous-git.sh fails open: without jq it allows commands on its own denylist and exits 0
`.claude/hooks/block-dangerous-git.sh` fails **open**: when its JSON parser is unavailable, a command on its own denylist passes through and the hook reports success.

### The dependency

Line 4 parses the hook payload with `jq`:

```
COMMAND=$(echo "$INPUT" | jq -r '.tool_input.command')
```

`jq` is not declared as a requirement anywhere in the repository. On a machine without it — ours was one — the substitution yields an empty string, no pattern matches, and the script reaches its final `exit 0`.

### Measured, both directions

Revision `3d6c941d52`. With `jq` removed from `PATH`, feeding it a command from its own denylist:

```
hooks/block-dangerous-git.sh: line 4: jq: command not found
exit status 0
```

Zero means *allowed*. With `jq` present, the same input behaves correctly:

```
dangerous input -> exit 2, "BLOCKED: ... matches dangerous pattern ..."
ordinary input  -> exit 0, silent
```

So the matching logic is sound; the failure mode is not. The only way this guard can fail is by silently permitting.

### The fix worth making

A parser being unavailable is not evidence that a command is safe. If the payload cannot be parsed, the hook should exit 2 with a clear message rather than 0. That matters more than the `jq` dependency itself: with `jq` present, any other parse failure still yields the same silent permission.

Worth noting what this guard protects. Its denylist includes `git reset --hard`, which is exactly the class of command that destroys uncommitted work belonging to someone else. A guard for that should not have "allow everything" as its failure mode.

### Lower priority, but a real cost

Matching is a substring test over the whole command with no notion of data versus execution, so writing *about* these commands is blocked too. We were unable to send an incident report through the normal path twice, because the report quoted one of the patterns inside a heredoc. It errs in the safe direction, but it means the guard cannot be documented by an agent working under it.

timujinne:
`block-dangerous-git.sh` fails open on any parse failure of the hook input, and the `push --force` denylist entry also catches the safe `push --force-with-lease`.

**Fail-open (the guard silently allows):**
- `jq` missing on PATH → `jq -r` errors, `$COMMAND` empty, no pattern matches → exit 0
- malformed JSON on stdin → same
- `.tool_input.command` absent or `null` → `jq -r` prints the literal string `null` (non-empty), passes the `[ -n ]` check, no pattern matches → exit 0

Fix: `jq -er` (treats `null`/`false`/parse-failure as an error → exit non-zero) plus an explicit empty guard, so all of the above now exit 2 (block) instead of exit 0 (allow).

**`--force-with-lease` false block:** the plain `push --force` substring matched `push --force-with-lease` too. Anchored to `push --force($|[^-])` so `--force-with-lease` is allowed and bare `--force` is still blocked.

Verified live with real exit codes across all cases (jq-missing / missing / null / malformed → exit 0 before, exit 2 after; `--force-with-lease` exit 2 before, exit 0 after; bare `--force` exit 2 both).

Closes #768


timujinne:
Scope note: this fixes fail-open on hook-input parsing (jq missing / malformed JSON / null command). The guard still assumes `grep` is present on PATH — a missing or broken `grep` would fail open the same way. That vector is deliberately out of scope here and tracked separately.

timujinne (commented):
## Verification summary

**Fixed:** the guard failed open when hook input could not be parsed — a missing `jq`, malformed JSON, or a null `command` field all resulted in the git operation being allowed. It now fails closed on any parse failure. Separately, `--force-with-lease` is now permitted while `--force` stays blocked.

**Verified by fact (mutation):** feeding malformed JSON and a null `command` to the pre-parse path *allows* the operation on the old script and *blocks* it on the new one — the two states are distinguished, not merely "something changed". `--force-with-lease` passes and `--force` is still rejected.

**Deliberately out of scope (named, not hidden):** the guard still assumes `grep` is on PATH; a missing/broken `grep` would fail open the same way. Tracked separately (see the scope note above).

PR #777, opened 2026-08-31
PkDialog Esc layering + component routing (JS) · image-selector quality sweep

mdon:
## What this does

JS-only changes to the hand-maintained `priv/static/assets/phoenix_kit.js` bundle — no version bump, no Elixir changes. Companion to catalogue PR BeamLabEU/phoenix_kit_catalogue#90 (the item selector's stacked details popup is the shipped case), but both fixes are generic core behaviour.

### 1. Esc closes only the TOP stacked dialog

Chromium groups the close watchers of dialogs whose `showModal()` ran without user activation (ours run from LiveView patches), so a single Esc fired `cancel` on EVERY open dialog — both popups closed when only the top one should. `PkDialog._onCancel` now detects an open descendant dialog, keeps itself open, drives the child's server close directly (its `data-close-event`, pushed to the owning component by CID) and closes the child element. `preventDefault` alone is not enough — Chromium stops the grouped chain at the first prevented watcher, so the child's own cancel never fires (verified: neither popup closed).

### 2. Component-aware routing

`pushEvent` from a hook always lands on the root LiveView, and `pushEventTo(element)` resolves via the element's `phx-target` attribute (absent on these elements) — so a **component-owned** modal's close event and a component-owned InfiniteScroll sentinel's `load_more` both hit the host LV instead of the component. Both now push by numeric CID from `closest("[data-phx-component]")`, falling back to the LV when there is no component ancestor (unchanged behaviour for LV-owned dialogs/sentinels).

### 3. Review hardenings (external panel, 2026-08-31)

- The direct push + child `close()` could double-fire the child's close event on browsers where the child's own `close` echo does fire — the parent now flags the child so its echo is suppressed once.
- The stacked-child lookup matched `dialog[open]`, which misses a dialog whose `open` attribute morphdom stripped between a patch and the next `_sync` — it now matches by the `:modal` truth source (`isDialogOpenInBrowser`).

Browser-verified on max-dev: Esc #1 closes only the details popup, Esc #2 the selector; the selector's list state survives.

## Notes for release

Hosts get this via `mix phoenix_kit.update` refreshing the vendored `phoenix_kit.js` — the catalogue popup's Esc layering and its legacy sentinel routing degrade gracefully until then.

https://claude.ai/code/session_018bbQqpwP4bRCYWNtsJAksH

### DIFF
--- a/lib/phoenix_kit/dashboard/tab.ex
+++ b/lib/phoenix_kit/dashboard/tab.ex
@@ -814,6 +814,12 @@ defmodule PhoenixKit.Dashboard.Tab do
 
   defp normalize_path(path) when is_binary(path) do
     path
+    # Matching is about the PATH: a current_path carrying a query or
+    # fragment (pages that publish their full URL so the language
+    # switcher can preserve state — catalogue detail's ?category=,
+    # 2026-08-31) must not break exact/prefix tab matches.
+    |> String.split(["?", "#"], parts: 2)
+    |> hd()
     |> String.trim_trailing("/")
     |> remove_url_prefix()
     |> remove_locale_prefix()
--- a/lib/phoenix_kit/notifications/digest_worker.ex
+++ b/lib/phoenix_kit/notifications/digest_worker.ex
@@ -17,6 +17,7 @@ defmodule PhoenixKit.Notifications.DigestWorker do
   """
 
   use Oban.Worker, queue: :notifications, max_attempts: 3
+  use Gettext, backend: PhoenixKitWeb.Gettext
 
   import Ecto.Query
 
@@ -203,10 +204,7 @@ defmodule PhoenixKit.Notifications.DigestWorker do
   end
 
   defp digest_text(count, label, cadence) do
-    Gettext.dgettext(
-      PhoenixKitWeb.Gettext,
-      "default",
-      "You have %{count} new %{label} %{period}.",
+    gettext("You have %{count} new %{label} %{period}.",
       count: count,
       label: label,
       period: period_phrase(cadence)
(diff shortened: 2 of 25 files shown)


mdon:
`80c1f085` adds the boss's requested **image-selector quality sweep** (so this PR is no longer JS-only — title updated): the media selector modal's grid tiles rendered the 150px `:small` thumbnail stretched across aspect-square cards — now `:card` tier like the media browser's grid. The sweep also caught two inverted variant chains in the non-image `resolve_url/2` clauses (`:card` and `:medium` both preferred the 150px thumbnail over the 800px medium in fallback) — fixed and pinned; the generic clauses had no pins, which is how the inversions survived. All other surfaces audited clean (list rows :small in 40px slots, icon previews, browser cards already :card). Precommit clean.

mdon:
Two more commits riding this PR:

- `936cc509` — **Tab matching vs query-carrying paths**: `normalize_path` strips query/fragment before matching, so pages that publish their full URL into `:url_path` (the catalogue's `?category=` drill, for language-switch state preservation — catalogue#90) don't break exact/prefix tab highlighting. Pinned in `tab_test.exs`.
- `99d3e5d1` — **Digest strings + full et/ru catalogues**: the notification digest built its text via `Gettext.dgettext(Backend, …)` — the runtime API the extractor can't see — so its five strings were in no catalogue; now the macro form. The et/ru po files were also completed: Estonian had 48 untranslated strings, Russian 129 (key-store diagnostics, login greetings, timezone copy, notification counts, storage settings, the auth/OAuth/magic-link flash set). Both locales are now gap-free; other locales only picked up merge reference updates.

precommit green; deployed to max-dev, hand-patched into the tim-dev preview.

PR #778, opened 2026-09-01
Doctor: check that hooksPath has an executable pre-commit, not that it equals .githooks

timujinne:
The Git Hooks doctor check currently passes only when `core.hooksPath` equals the literal string `.githooks`. That tests for one particular configuration rather than for the property the check exists to verify — that a tracked, executable `pre-commit` hook is actually reachable.

**What changes.** The check now asks whether the directory `core.hooksPath` points at contains an executable `pre-commit`, instead of comparing the path to a fixed string. A setup that keeps hooks somewhere else — a shared directory outside the worktree, for instance — stops being reported as broken when it is not.

**What deliberately does not change.** `core.hooksPath` being unset still warns, with the same text and the same suggested fix (`git config core.hooksPath .githooks`). A checkout where the hook is missing or not executable still warns rather than passing. The goal is to stop rejecting valid configurations, not to loosen the check.

Verified against fixtures covering four states:

| state | verdict |
|---|---|
| hooksPath → other dir, executable `pre-commit` present | pass |
| hooksPath → other dir, `pre-commit` missing | warn |
| hooksPath → other dir, `pre-commit` present but not executable | warn |
| hooksPath unset | warn (text unchanged) |
| `.githooks` with executable `pre-commit` | pass (unchanged behaviour) |

Two files touched: the doctor task and its test. `mix compile --warnings-as-errors`, `mix format --check-formatted` and `mix credo --strict` are clean on both.

One note in the interest of full disclosure: the `tracked?: false` clause matches before the `hooks_path` clauses, so in a checkout without `.githooks/pre-commit` the doctor still suggests fixing `.githooks` even when `core.hooksPath` points somewhere workable. That is a warning rather than a false pass, and reordering the older clause is a wider behavioural change than this PR intends — happy to follow up separately if you would like it addressed.

### DIFF
--- a/lib/mix/tasks/phoenix_kit.doctor.ex
+++ b/lib/mix/tasks/phoenix_kit.doctor.ex
@@ -82,11 +82,12 @@ defmodule Mix.Tasks.PhoenixKit.Doctor do
        `PhoenixKit.Migrations.ExpectedSchema` manifest as an additional,
        non-fatal check (never `:fail`). Passes and says so if the manifest
        has been removed or overridden away in this checkout.
-   27. **Git Hooks** — is `.githooks/pre-commit` enabled via
-       `core.hooksPath`? Only runs inside a checkout of phoenix_kit itself
-       (`.githooks/pre-commit` is a phoenix_kit-repo convention, not
-       something installed into a consuming host app) — silently skipped
-       otherwise.
+   27. **Git Hooks** — does `core.hooksPath` point at a directory that has an
+       executable `pre-commit` in it? Any directory qualifies — `.githooks`
+       is only the convention this checkout happens to track. Only runs
+       inside a checkout of phoenix_kit itself (that convention is a
+       phoenix_kit-repo thing, not something installed into a consuming host
+       app) — silently skipped otherwise.
   """
 
   use Mix.Task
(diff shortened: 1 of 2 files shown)


timujinne (commented):
Reviewed independently before opening this for maintainer attention. Recording what was verified and how, and what was not — the latter matters as much.

**Verified by running, not by reading.** Fixtures were built for each state (executable `pre-commit`, missing, present-but-not-executable) and pushed through `git_hooks_verdict/1` and `pre_commit_executable?/1` from the compiled patched module:

- hooksPath pointing at another directory with an executable hook → pass
- same, hook missing → warn, with the fix naming that directory
- same, hook present but `chmod 0644` → same warn
- hooksPath unset → warn, text and behaviour unchanged
- `.githooks` with an executable hook → pass, identical to pre-patch behaviour

The two warn cases matter most: they confirm the change replaces one comparison with a property test, rather than replacing a comparison with `true`.

**Verified against real state, not only fixtures.** This repository currently has two worktrees whose branches predate the `.githooks` directory, so the tracked hook is absent there. The patched check reports a warning for both — not a pass. That was the specific risk worth ruling out: a check that starts passing everywhere would be worse than the one it replaces.

**Clean:** `mix compile --warnings-as-errors`, `mix format --check-formatted`, `mix credo --strict` on both touched files.

**Not verified, stated plainly:** the test file cannot be run to green here. `mix test` fails before reaching any test on `PhoenixKit.Test.LiveDatabaseGuard.check!/1 is undefined` — reproduced identically on an unmodified checkout of `main`, so it is a pre-existing environment issue rather than something this branch introduces, but it does mean the new tests were exercised through direct calls rather than through ExUnit. Also not reproduced: the `:unknown` branch of `pre_commit_executable?/1` via a genuine filesystem permission error.

The clause-ordering caveat is described in the PR body; it is a warning with a misleading fix suggestion, never a false pass.

timujinne:
Ready for maintainer review. Independent review is posted above, with the verification method and the two things it could not establish stated explicitly.

PR #782, opened 2026-09-04
Fix two duplicate/hreflang defects in multi-domain sitemaps

timujinne:
Two defects in multi-domain sitemap generation, both found on a live three-domain site (en primary + de + fr) where the generated per-host sitemaps disagreed with what the rendered pages actually emit.

### 1. Lone self+x-default hreflang on a single-language URL

`DomainMode.group_alternates/4` emitted a self + `x-default` pair for a canonical-path group containing only one language, whenever that language was the primary domain's. That is a two-entry hreflang "set" for a page with no translations.

The page-level builders already treat an under-2-entry set as noise and emit nothing, so the sitemap advertised an alternate set that the page's own `<head>` never backed up — a crawler following the sitemap saw a promise the page did not keep. `group_alternates/4` now returns `[]` for such a group.

Because the function is shared by every source via `rebuild_for_domains/2`, this affected shop products and publishing posts alike, not one source.

### 2. The same `<loc>` listed twice in one host's file

A locale-prefixed clone route (e.g. `live "/de"` pointing at the home LiveView, so `/de` does not 404 on a prefix install) is picked up by `RouterDiscovery` with no `canonical_path`, so it forms its own group. Re-hosting strips the prefix and lands it on `https://<de-host>/` — exactly where the static source has already placed that domain's home. The two groups cannot merge, so the file listed the home twice: once with the full cross-domain alternates, once bare.

Per-host entries are now deduplicated by `<loc>`, keeping the richest description of the URL: alternates first, then priority, then arrival order so the result is deterministic run to run.

`extra_for_primary/5` already enforced this same "the same `<loc>` in one file twice" rule across sources for the domainless case — it simply never applied within a host's own set, which is where the clone-route duplicate lands.

### Verification

Regression tests added for both cases in `test/integration/sitemap/domain_mode_test.exs`. Full sitemap integration suite: 76 tests, 0 failures. `mix format` clean.

Confirmed against a live three-domain install by regenerating the sitemaps:

* duplicate home gone from both non-primary domains (639 -> 638 URLs each; primary unchanged at 643),
* untranslated URLs now carry no hreflang block, matching their rendered pages,
* fully translated URLs keep their complete de/en/fr/x-default sets pointing at the correct per-domain translated slugs,
* the surviving home entry is the rich one (priority 0.9 with the full alternate set), not the bare clone.

### Note

Version and CHANGELOG are intentionally left untouched for the maintainer.

### DIFF
--- a/dev_docs/pull_requests/2026/782-sitemap-domain-mode-hreflang-and-duplicate-home/CLAUDE_REVIEW.md
+++ b/dev_docs/pull_requests/2026/782-sitemap-domain-mode-hreflang-and-duplicate-home/CLAUDE_REVIEW.md
@@ -1,0 +1,107 @@
+# Claude Review — PR #782
+
+**Title:** Fix two duplicate/hreflang defects in multi-domain sitemaps
+**Author:** timujinne
+**Head commit:** 3ef2174a
+**Verdict:** Approve — two IMPROVEMENT-HIGH findings raised in review, both applied in the branch
+
+## Summary
+
+Two defects in `DomainMode`, both found on a live three-domain install
+(en primary + de + fr) by comparing the generated per-host sitemaps against
+what the rendered pages actually emit:
+
+1. A canonical-path group containing a single language still received a
+   self + `x-default` hreflang pair whenever that language was the primary
+   domain's. The page-level builders already drop an under-2-entry set as
+   noise and emit nothing, so the sitemap advertised an alternate set the
+   page's own `<head>` never backed up.
+2. One host's file could list the same `<loc>` twice. A locale-prefixed clone
+   route (`live "/de"` on the home LiveView, so `/de` does not 404 on a prefix
+   install) is discovered by `RouterDiscovery` with no `canonical_path`, so it
+   forms its own group; re-hosting strips the prefix and lands it on
+   `https://<de-host>/`, where the static source has already placed that
+   domain's home. The two groups cannot merge, so the home appeared twice —
+   once with the full cross-domain alternates, once bare.
+
+## Review scope
+
+Read in full: the diff against `a9a66190` across `domain_mode.ex`,
+`generator.ex`, `url_entry.ex` and `domain_mode_test.exs`; plus the
+surrounding `rebuild_for_domains/2` → `domain_entries/5` →
+`group_alternates/4` path, `extra_for_primary/5`, `domainless_entries/4`,
+`home_url/5`, and the pre-existing dedup in `generator.ex`.
+
+Specifically checked and found correct:
+
+- **The guard does not drop URLs.** `loc` and `alternates` are computed
+  independently in `domain_entries/5`; an emptied alternate list still emits
+  the entry.
+- **Single-language non-primary groups.** These previously produced a lone
+  self-reference. A one-entry hreflang set is not a set, so returning `[]`
+  for them is a correction, not a regression.
+- **Domainless languages survive.** `extra_for_primary/5` filters against
+  `taken` before dedup runs, so an own-language entry and a domainless entry
+  cannot collide into a loss.
+- **Sort order.** Non-primary hosts were sorted inside `domain_entries/5`,
+  the primary after concatenation; both now sort by `loc` on the same key
+  after dedup, which makes the order strictly total rather than changing it.
+- **Priority parsing.** `priority` arrives as float, string, or nil depending
+  on the source; the shared ranking reuses `UrlEntry.parse_priority/1` and
+  falls back to `0.0` rather than raising on an unparseable value.
+
+## Findings
+
+### IMPROVEMENT - HIGH — the guard silently voided an existing assertion (applied)
+
+`test/integration/sitemap/domain_mode_test.exs` — the test "untranslated group
+appears only on its language's domain" pinned rule m2 (never mislabel another
+language as `x-default`, stated in the `DomainMode` moduledoc) with
+`refute Enum.any?(fr_entry.alternates, &(&1.hreflang == "x-default"))`.
+
+That group is single-language, so after the new guard `alternates == []` and
+the `refute` passes against an empty list — vacuously, for a reason unrelated
+to what it claims to test. Proven by mutation: making `group_alternates/4`
+append an `x-default` on the `nil -> links` branch (i.e. violating m2 outright)
+left the suite fully green.
+
+Applied: that test now asserts the empty list it actually exercises, and a new
+test pins m2 with a genuine two-language group carrying no primary-language
+entry, so the `refute` is meaningful again.
+
+### IMPROVEMENT - HIGH — a second dedup policy beside the one already shipped (applied)
+
+The branch introduced a local `dedup_by_loc/1` in `DomainMode` ranking
+candidates by alternate count, then priority, then arrival order. `Generator`
+already carried `dedupe_by_loc/1` + `entry_richness/1` solving the same
+problem, with a comment describing this very scenario, and with one rule the
+new implementation did not reproduce: a `RouterDiscovery` entry scores `0` and
+therefore **always** loses a same-`loc` collision, regardless of its own
+priority. A ranking that compares only alternates and priority would let a
+route-discovery entry win as soon as it happened to carry either.
+
+Applied: the policy moved to `UrlEntry.dedupe_by_loc/1` +
+`UrlEntry.richness/1`, and both `Generator` and `DomainMode` now call it. The
+duplicate implementation and its private priority parser were removed rather
+than kept in parallel.
+
+## Verification
+
+- `mix test test/integration/sitemap/` — 77 tests, 0 failures.
+- Full suite: 43 doctests, 4388 tests, 0 failures.
+- `mix format --check-formatted` clean on all touched files.
+- **Mutation-proven, one failure each, green after restore:** violating m2 in
+  `group_alternates/4` fails the new m2 test; removing the dedup call from
+  `rebuild_for_domains/2` fails the duplicate-home test.
+- Confirmed against the live three-domain install by regenerating: the
+  duplicate home is gone from both non-primary domains (639 → 638 URLs each,
+  primary unchanged at 643); untranslated URLs carry no hreflang block,
+  matching their rendered pages; fully translated URLs keep their complete
+  de/en/fr/x-default sets pointing at the correct per-domain translated slugs;
+  and the surviving home entry is the rich one (priority 0.9 with the full
+  alternate set), not the bare clone.
+
+## Not done here
+
+`@version` and `CHANGELOG.md` are deliberately left untouched for the
+maintainer.
--- a/lib/modules/sitemap/domain_mode.ex
+++ b/lib/modules/sitemap/domain_mode.ex
@@ -122,19 +122,22 @@ defmodule PhoenixKit.Modules.Sitemap.DomainMode do
           |> Map.values()
           |> Enum.map(&group_by_language(&1, base_url, default_lang, enabled_bases))
 
+        # One host's file must list a URL once, and two producers can reach the
+        # same <loc> here: a locale-prefixed clone route (/de, /fr on the home
+        # LiveView) is discovered with no canonical_path, so it forms its own
+        # group, and re-hosting strips the prefix onto exactly the home URL the
+        # static source already placed on that language's domain. The two groups
+        # cannot merge, so the file listed the home twice. Resolved by the same
+        # richness policy the flat generator uses (UrlEntry.dedupe_by_loc/1).
         Map.new(domains, fn %{host: host, language: lang, primary: primary?} ->
           own = domain_entries(groups, lang, host_by_lang, primary, base_url)
 
           entries =
             if primary?,
-              do:
-                Enum.sort_by(
-                  own ++ extra_for_primary(own, groups, host_by_lang, primary, base_url),
-                  & &1.loc
-                ),
+              do: own ++ extra_for_primary(own, groups, host_by_lang, primary, base_url),
               else: own
 
-          {host, entries}
+          {host, entries |> UrlEntry.dedupe_by_loc() |> Enum.sort_by(& &1.loc)}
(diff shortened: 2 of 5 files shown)


ddon:
Still draft, not touching

timujinne (commented):
## Independent review — verdict: ready, with four cosmetic items

Reviewed at head `932ca122` (code commit `3ef2174a`), base `a9a66190`. No commit on `main` since the base touches `lib/modules/sitemap/` or `test/integration/sitemap/`, so there is no dead-code-base risk.

### Spec: pass — and checked by running, not by reading the diff

Both promised defects exist, sit where the description says, and are fixed there. This was established by running the PR's test file **against pre-fix code**:

- **Defect 1** — `domain_mode.ex:227-228`. Base emits a lone hreflang for a single-language group: `assert fr_entry.alternates == []` failed with `left: [%{href: "https://site.example.fr/only-french", hreflang: "fr"}]`.
- **Defect 2** — `domain_mode.ex:140`. Base lists the same `<loc>` twice: one bare entry (`canonical_path: nil`, alternates `[fr]`) and one rich (`canonical_path: "/"`, alternates `[en, fr, x-default]`). A single failure dump exhibits both defects at once.

Nothing missing, wrong, or extra against the stated spec.

### Findings — no critical, no major

**1. The description understates the scope of fix #1.** The body says the lone pair was emitted "whenever that language was the primary domain's". The guard is `map_size(by_lang) < 2`, so it also strips the lone self-reference from single-language groups in *non-primary* languages — the pre-fix failure above is an `fr` (non-primary) group changing `[{hreflang: "fr"}] → []`. The behaviour is correct; the description lags it, and a maintainer reading the body would not expect non-primary URLs to change. One sentence fixes it.

**2. The moduledoc still states the old two-rule alternates contract.** `domain_mode.ex:14-27` says the set is "each group language on its home URL, plus `x-default` … (omitted otherwise)". The new third rule — under two languages ⇒ no alternates at all — is documented only in a private-function comment. Verified: the diff contains no `@moduledoc` change. For a published library the moduledoc is where the contract is read.

**3. A private customer application is named in library code.** `domain_mode.ex:220` introduces `Decor3dprintWeb.SEO.with_x_default/1` in a comment. This is the only named private application anywhere in `lib/`: a survey of every external module name appearing in `lib/` comments (42 after removing modules defined here and stdlib) splits cleanly into BeamLab-family siblings, third-party dependencies, and **deliberate generic placeholders for the host app** — `YourApp`, `AppName`, `AppWeb`, `HostWeb`. The repository's own convention for talking about a host app is `AppWeb`/`YourApp`, so this is a one-word change. Keep the rule it illustrates ("an hreflang set of fewer than two entries is not a set"), and keep the `PhoenixKitEcommerce.Web.SEOHelpers` half — that is a documented sibling already referenced at `lib/phoenix_kit_web/users/auth_seo.ex:6,24`.
*Caveat: the survey covered `#` comments in `lib/` only, not `@doc`/`@moduledoc` bodies.*

**4. "76 tests, 0 failures" in the body is 77.** Base sitemap suite is 74; the diff adds three tests (not two — the m2 test, the primary-single-language test, and the duplicate-home test).

**Nitpick.** `domain_entries/5` still ends `|> Enum.sort_by(& &1.loc)` at `domain_mode.ex:213`, but every consumer re-sorts after dedup at line 140 — dead work. Note also that `CLAUDE_REVIEW.md` states "both now sort by `loc` on the same key after dedup", implying the inner sort was removed; it was not.

### Boundary of the chosen fix — not a regression

Dedup resolves the duplicate `<loc>` but does not reconcile the two hreflang sets. When two rich canonical groups collide on one URL, the loser's group keeps advertising the loc it lost:

```
site.example.com/b   alts=[{en, …/b}, {fr, https://site.example.fr/a}, {x-default, …/b}]
site.example.fr/a    alts=[{en, …/a}, {fr, https://site.example.fr/a}, {x-default, …/a}]   ← never mentions /b
```

`…com/b` claims its French version is `…fr/a`, but that page's own set describes group `/a` and never points back. Reachable only from malformed content data, and the pre-fix behaviour — the same loc listed twice with conflicting sets — was no better. Worth knowing before anyone assumes dedup makes hreflang self-consistent.

### Same defect class, one layer up — pre-existing, outside this PR

`generator.ex:730-762` (`collect_multilingual_entries/3`) builds flat-mode alternates with no under-2 guard: a single-entry canonical group gets `[self, x-default]`, and `default_entry` falls back to `|| List.first(entries)` at line 746, which can label a non-default language as `x-default` — precisely the rule the `DomainMode` moduledoc says must never be violated. That is the non-domain-mode path, i.e. most installs. The PR's own rationale — the sitemap must not promise what the page's `<head>` does not back up — applies to it verbatim. **Traced by reading, never executed.**

### Checked and found correct

- The guard cannot drop URLs: `loc` and `alternates` are computed independently (`domain_mode.ex:204-210`). `home_url/5` always returns a URL, so the link count always equals `map_size(by_lang)` and the guard's threshold is exactly right.
- `UrlEntry.dedupe_by_loc/1` and `richness/1` are a faithful move of the Generator originals; both call sites updated; no other callers; the "`:router_discovery` always loses" rule survives the move.
- Determinism holds: `Enum.max_by` keeps the first on a tie, `Enum.group_by` preserves arrival order.
- Reciprocity holds by construction for well-formed input — one alternates set per group, applied identically to every member. Dedup never orphans an hreflang target: the survivor carries the same `loc`.
- `mix format --check-formatted` clean on all four touched files; `mix credo --strict` on the three changed lib files — 89 modules/functions, no issues; `MIX_ENV=test mix compile --warnings-as-errors --force` clean.

### Not verified — stated plainly

- **Anything on the live three-domain install.** Every number in the body's Verification section that refers to the running site — 639→638, 643 unchanged, "untranslated URLs now carry no hreflang block", "the surviving home entry is the rich one" — rests on the original author's report alone. I had no access to that install and did not attempt to get it. If those numbers are meant to be trusted, they are still one person's word.
- **The premise about the page-level builders.** Neither `Decor3dprintWeb.SEO.with_x_default/1` nor `PhoenixKitEcommerce.Web.SEOHelpers.dedup_or_empty/1` is defined in this repository. Only the premise was checked — that a one-entry hreflang set is meaningless per the spec — not that those two functions behave as the comment claims.
- **The flat-path twin at `generator.ex:730-762`** — read, never executed. It is private and driving it needs real sources plus multi-language collection.
- **Dialyzer.** CI runs it; this review did not. Type-level regressions from the new `@spec`s on `dedupe_by_loc/1` and `richness/1` are unverified.
- **Whether promoting `dedupe_by_loc/1` and `richness/1` to documented public functions is an API-surface expansion you mind.** `UrlEntry` already exposes documented helpers such as `parse_priority/1`, so it matches local style — but that call is yours, and it is flagged so it is not made silently.

### Out of scope, pre-existing

`domain_mode.ex:190-192` — dialect collapse can lose a URL. Noted as ambiguous and not touched by this PR.

ddon:
@timujinne still a draft, waiting it to be finished

timujinne (commented):
# Review — PR #782: Fix two duplicate/hreflang defects in multi-domain sitemaps

Branch: `fix/sitemap-domain-mode-hreflang-and-duplicate-home` (head `0b9e33af`) into `BeamLabEU/phoenix_kit#main` (`upstream/main` at `4e02d885`).

## Stage 1: Spec Compliance

PR describes two defects and fixes:

1. `DomainMode.group_alternates/4` emitted a self+`x-default` hreflang pair for a canonical-path group with only one language, when that language is the primary's — noise the page-level builders already suppress.
2. A locale-prefixed clone route (e.g. `live "/de"` on the home LiveView) forms its own no-`canonical_path` group, re-hosts onto the same `<loc>` the static home already occupies, and the per-host list carried the home twice.

Verified independently:

- **`domain_mode.ex:224-225`**: new guard `defp group_alternates(by_lang, _host_by_lang, _primary, _base_url) when map_size(by_lang) < 2, do: []` matches defect 1 exactly.
- **`domain_mode.ex:138-141`**: entries per host now run through `UrlEntry.dedupe_by_loc/1` before the final `Enum.sort_by(& &1.loc)`, matching defect 2.
- **Shared dedup logic**: `dedupe_by_loc/1` + `richness/1` (renamed from the private `entry_richness/1`) moved from `generator.ex` into `url_entry.ex:239-278`, and both `Generator.collect_entries/1` and `Generator`'s flat path, plus the new `DomainMode` call site, now share one implementation. `RouterDiscovery` entries still score `0` unconditionally (`richness(%__MODULE__{source: :router_discovery}), do: 0`), so a route-discovery entry can never outrank a content-source entry regardless of its own priority — the exact rule the pre-existing `generator.ex` comment described, correctly preserved after the move.
- **Regression tests**, `test/integration/sitemap/domain_mode_test.exs`:
  - `"a single-language group on the PRIMARY's own language carries no alternates at all"` (new) — pins defect 1.
  - `"one host lists a <loc> once when two sources produce it, keeping the richer entry"` (new) — pins defect 2, and additionally asserts the primary host is untouched and the *richer* survivor (carrying the full alternate set) is kept over the bare clone.
  - `"untranslated group appears only on its language's domain"` (existing test) rewritten: the old assertion `refute Enum.any?(fr_entry.alternates, &(&1.hreflang == "x-default"))` was true but vacuous once the guard empties `alternates` for that same group — it no longer exercises rule "m2" (never mislabel another language as x-default; documented in the `DomainMode` moduledoc, `domain_mode.ex:21-22`). The PR adds `assert fr_entry.alternates == []` to state what the test actually now covers, and adds a **new** test, `"no x-default when the group has no entry for the primary's language (m2)"`, with a genuine two-language group to keep m2 meaningfully pinned. This self-correction is accurate — I reproduced it.

**Independent verification performed:**

- Reverted only the three `lib/` files (`domain_mode.ex`, `generator.ex`, `url_entry.ex`) to `upstream/main` while keeping the new/edited tests, and reran `test/integration/sitemap/domain_mode_test.exs --include integration`: **3 of 13 tests fail** — the two new regression tests and the corrected m2 test — confirming both fixes are load-bearing and the tests are not tautological. Restored the fix afterward; full suite green again.
- `mix test test/integration/sitemap/ --include integration`: **77 tests, 0 failures**, matching the PR's claim.
- `mix format --check-formatted` on all four touched files: clean.
- `mix credo` on the three touched lib files: no issues.
- `mix compile --warnings-as-errors --force`: clean, no warnings (confirms the `UrlEntry` alias/call sites resolve correctly after the move).
- `mix merge-tree upstream/main HEAD`: no conflicts against current upstream `main`.

**Full-suite caveat:** `mix test --include integration` (whole core suite) showed 4 failures — 3 in `Mix.Tasks.PhoenixKit.DoctorOrphanedFkDestructiveTest` (FK-violation errors) and 1 in `PhoenixKit.Migrations.HandDeclaredManifestTest` (migration-count mismatch `{:above_current, 187, 182}`). I reproduced the same 3 FK failures on a clean `upstream/main` checkout against the same shared test database, so they predate this PR and are environmental (shared-DB/migration-state artifacts of this container, not a regression from the diff). The manifest test passed in isolation on baseline; its failure only showed up under full-suite concurrency against the shared database, so it is very likely the same category of environment noise rather than something this PR introduced, but I could not fully rule that out via reading alone. None of the four touch sitemap code paths.

**Spec Verdict: PASS**

Both stated defects are real, both fixes are minimal and load-bearing (proven by revert), the shared-logic refactor is a genuine improvement (was duplicated across two files) not scope creep, and the self-corrected test is an accurate fix rather than a cover-up. The `CLAUDE_REVIEW.md` file added under `dev_docs/pull_requests/2026/782-…/` follows an existing, already-merged upstream convention (hundreds of prior PRs on `main` carry the same `CLAUDE_REVIEW.md` pattern, including a committed `.claude/` directory) — not an out-of-place addition. No AI-attribution trailers (`Co-Authored-By`, "Generated with Claude Code", session links) appear in either commit message or the PR body.

---

## Stage 2: Code Quality

### [NIT]: `richness/1` made public without clear external caller
**File**: `lib/modules/sitemap/url_entry.ex:264` (`@spec richness(t()) :: number()` / `def richness(...)`)
**Problem**: `richness/1` is exported with a `@doc`, but its only callers are `dedupe_by_loc/1` in the same module. Making it public is defensible for testability/documentation, but if nothing outside `url_entry.ex` calls it directly, `defp` plus a `@doc false`-style comment would keep the same clarity without growing the public API surface.
**Suggestion**: Either leave as-is (harmless, and the doc is genuinely useful context for future maintainers) or make it private if there's no plan for external callers.
**Rationale**: Minor API-surface hygiene; not a defect.

### [NIT]: `group_alternates/4`'s single-language guard clause placement
**File**: `lib/modules/sitemap/domain_mode.ex:224-225`
**Problem**: None — the guard clause is placed directly above the general clause with a clear comment tying it to the app-level consumers' own noise-suppression rule. Flagging only that the comment cites concrete consumer module names (`Decor3dprintWeb.SEO.with_x_default/1`, `PhoenixKitEcommerce.Web.SEOHelpers.dedup_or_empty/1`) inside a library-level comment — reasonable as rationale/evidence for the "under-2 is noise" rule, but worth being aware these are downstream-app references baked into upstream library source, so they'll silently rot if those consumer modules are renamed.
**Suggestion**: No action needed; purely a documentation-freshness note for future maintainers.
**Rationale**: Doesn't affect correctness; only comment maintainability.

**Quality Summary:** 0 critical, 0 major, 0 minor, 2 nitpick
**Quality Verdict:** Ship

---

## Overall Verdict: PASS

No blocking issues. The fix is minimal, both defects are independently reproduced and confirmed fixed (tests fail without the `lib/` changes, pass with them), the dedup-logic consolidation is a genuine simplification rather than scope creep, `mix format`/`credo`/compile are clean on the touched files, and there are no merge conflicts against current upstream `main`. The only open item is environmental: 4 unrelated failures on the full core suite (3 reproduced independently on a clean `upstream/main` checkout against the same shared test database, 1 only surfaced under full-suite concurrency) — none touch sitemap code and none are attributable to this diff.

PR #785, opened 2026-09-05
V184: remove the dead shop_currency setting

timujinne:
## V184: remove the dead `shop_currency` setting

One new chain version. `V135` seeds a `shop_currency` setting (`INSERT … ON CONFLICT ("key") DO NOTHING`, value `USD`, module `shop`). Nothing reads it: a full grep over `phoenix_kit`, `phoenix_kit_billing`, `phoenix_kit_ecommerce` and a host application finds only the V135 seed and the `ExpectedSchema` manifest. The default currency is the `is_default` row of `phoenix_kit_currencies` and nothing else; a second "shop currency" that does nothing is a trap for the next reader (a host had two disagreeing "defaults" side by side for months without noticing).

### Why a new version and not an edit of V135

- Editing a shipped `vNNN.ex` changes `chain_hash/0` and fails `release_check` for every host.
- It would also change nothing: the seed is `ON CONFLICT DO NOTHING`, so every already-migrated install keeps its row. Only a `DELETE` in a new version reaches them.

### What V184 does

- `up`: `DELETE FROM phoenix_kit_settings WHERE "key" = 'shop_currency'`, marker `184`.
- `down`: re-inserts the row with the V135 seed statement verbatim (`ON CONFLICT DO NOTHING`, so a row a host re-created by hand is never overwritten), marker `183`.
- `postgres.ex` `@current_version 184`, moduledoc entry; `expected_schema.ex` regenerated with `dev_docs/squash/generate_baseline.exs` (not hand-edited).
- Test `test/phoenix_kit/migrations/v184_test.exs`: up removes the row and stamps 184; down restores `USD` and stamps 183; down leaves a hand-created `EUR` row untouched.

Companion PRs in the same work (per-domain currency, stage Э0): BeamLabEU/phoenix_kit_billing#30, BeamLabEU/phoenix_kit_ecommerce#31.

### Manifest and tests

- `expected_schema.ex` is restamped with `dev_docs/squash/restamp_chain_hash.exs --restamp` — the documented path for a version that adds, drops or reshapes nothing (precedent: V182, commit 975f88fe). A full regeneration was attempted first and the generator aborts on this environment at the mode-shape diff for `sync_shop_category_slugs()` / `sync_shop_product_slugs()` (identical definitions, different `body_md5` between the stepwise and single-shot schemas); it aborts identically on unmodified `main`. Those two trigger functions are deliberately not declared in the manifest since V171 (comment near line 27778), but the generator's guard runs before that exclusion applies. Left as a tooling note, not fixed here.
- Full suite: 43 doctests, 4497 tests, 0 failures after the restamp (before it, the only two failures were the chain-hash checks in `phoenix_kit_release_check_test.exs`). `test/phoenix_kit/migrations/`: 433 tests + 16 doctests, 0 failures, `v184_test.exs` included. `check_migration_sync/0`: V135..V184 contiguous, every module loadable.
- Environment note: bare `mix test` in this checkout did not compile `test/support` (7 files) on its own; `MIX_ENV=test mix compile --force` followed by `mix test --no-compile` was used.

### Host verification

Applied on a live host through `phoenix_kit_update_v182_to_v184` (V183 + V184): `shop_currency` row gone, marker `184`, storefront 200 on both domains, no errors in the log after restart.

### Not in this PR

Version bump and CHANGELOG (maintainer). The next stage adds V185 (currency freeze columns on carts/orders).

### DIFF
--- a/lib/phoenix_kit/migrations/expected_schema.ex
+++ b/lib/phoenix_kit/migrations/expected_schema.ex
@@ -157,6 +157,24 @@ defmodule PhoenixKit.Migrations.ExpectedSchema do
   # over the 49 shipped files; the real-database integration suite re-ran
   # clean against a DB migrated through V183.
   #
+  # V184 (2026-09-05, per-domain currency Э0) declares NO object here, and
+  # cannot: it is a pure data migration — `DELETE FROM phoenix_kit_settings
+  # WHERE "key" = 'shop_currency'` (a dead setting seeded by V135 that nothing
+  # reads), plus the version-marker COMMENT. No table, column, index or
+  # constraint is added, dropped or reshaped — the V182 class — so `chain_hash`
+  # is restamped over the 50 shipped files rather than the manifest being
+  # regenerated. A full regenerate was attempted first
+  # (`dev_docs/squash/generate_baseline.exs`) and aborts in this environment on
+  # a pre-existing mode-shape mismatch between the stepwise and single-shot
+  # `sync_shop_category_slugs()`/`sync_shop_product_slugs()` function dumps
+  # (identical `definition` text, different `body_md5` — reproduces identically
+  # on unmodified `main`, unrelated to this version); those two functions are
+  # already deliberately excluded from this manifest since V171. The chain was
+  # re-run end to end into a fresh database (V135→V184) and the migration's
+  # real statements are exercised against a seeded settings row by
+  # `test/phoenix_kit/migrations/v184_test.exs`, including down/1 never
+  # clobbering a value an operator re-created by hand.
+  #
   # V180 was fixed post-publish (2.13.11): its bare top-level `LOCK TABLE` moved
   # inside the `DO $$` block that already carries the dedupe UPDATE and the
   # `CREATE UNIQUE INDEX`, and that UPDATE's `updated_at` expression changed from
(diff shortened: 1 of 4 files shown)


timujinne (commented):
Independent review (read-only reviewer, separate from the author), range `5de7e6c6..7ad4ce56`. Verdict: **PASS**, no findings against the V184 work.

**Verified directly**
- `down/1` reproduces the V135 seed byte for byte (`INSERT INTO … ("key","module","value","value_json") VALUES ('shop_currency','shop','USD',NULL) ON CONFLICT ("key") DO NOTHING`); `date_added`/`date_updated`/`uuid` carry DB defaults, so the omitted columns are correct in both the seed and the restore. `up/1` is the `DELETE` plus marker `184`; `down/1` restores and stamps `183`.
- `expected_schema.ex` diff is a `@chain_hash` restamp plus a comment block; `objects/1` unchanged. V184 adds no table, column, index or constraint, so the restamp is the right mechanism (same as V182, commit 975f88fe). The generator abort on `sync_shop_*_slugs` is reproduced on unmodified `main` and those functions are already excluded from the manifest since V171.
- `postgres.ex`: `@current_version 184`, V184 entry marked latest, V183 entry kept.
- Tests: `v184_test.exs` runs the real `up_statements/1`/`down_statements/1` against a seeded row, including a hand-created `EUR` row surviving `down`. `test/phoenix_kit/migrations/` + release_check test: 16 doctests, 448 tests, 0 failures. `mix phoenix_kit.release_check`: Migration Version Sync passes (V135..V184 contiguous, chain_hash matches 50 files); remaining items are feature-branch noise (CHANGELOG, branch, tag).
- Commit by the author only; no version/CHANGELOG change.

**Notes, not blockers**
- On the host's stand branch (this PR merged with upstream main and PR #782's sitemap commits), the full suite shows one order-dependent failure in `admin_route_dedup_test.exs`; it passes in isolation and its own moduledoc documents the parallel-compilation race. V184 touches only `lib/phoenix_kit/migrations/`, no route or module-discovery code.
- `mix test` without a prior `MIX_ENV=test mix compile --force` did not compile `test/support` in this environment.

PR #786, opened 2026-09-05
Per-instant timezone helpers, a deprecation, and V184: permanent activities carry the settings history

mdon:
## Why

Since 2.13.9 the `time_zone` setting and `user_timezone` hold an IANA id on anything that touched the picker. Five modules kept turning that value into one number and adding it to other instants — `Integer.parse` (0 for an id) or, after 2.14.1, a snapshot of today's offset — and each carried its own copy of "the viewer's zone: profile, else site, else UTC" (two of them disagreed with core's rule). The module PRs route every conversion through core's existing per-instant helpers; this PR names the three questions the modules kept answering themselves so the next module does not.

## What

- `TimeZone.for_viewer/1` — the value for a `%Scope{}`, a user (full or partial map — module pages and test scopes carry users without the column), or nil. `get_user_timezone/1`'s rule, made total.
- `TimeZone.date_start/2` — the UTC instant a local date begins, resolved for THAT date (subtracting one offset taken today from both bounds of a day window was an hour off across a DST switch). A midnight the clocks skip or repeat is still the first instant of that date — Santiago and Havana pinned.
- `TimeZone.local_date/2` — the local date of an instant (`Date.utc_today/0` is yesterday for a Tallinn viewer until 03:00 in summer).
- A "which function for what" section in the `TimeZone` moduledoc.
- `Utils.Date.offset_to_seconds/1` is `@deprecated`: a snapshot of the offset now; every caller that added it to another instant was wrong across a switch. No caller remains in core; behaviour stays pinned.
- `get_user_timezone/1` and `_cached/2` treat a blank value as unset (the changeset normalises `""` away, a raw write does not); `site_zone/0` has no rescue — `Settings.get_setting/2` already degrades.

## Not in this PR

- No version or CHANGELOG change (boss's). Note for the release notes: the unreleased 2.14.1 entry names bookings' `Engine.site_offset_seconds/0`, which the bookings PR removes.
- The modules keep the pre-2.0 helpers they moved to (`shift_to_offset/2`, `parse_datetime_local/2`, `format_datetime_local/2`); their `~> 2.0`/`~> 2.4` floors cannot reach these three functions until a release, and each module's AGENTS.md says which private helper collapses into which once the floor moves.
- Found on the way, not changed: `same_group?/2` calls `Asia/Jerusalem` and `Europe/Athens` the same zone — Israel switches on the Friday before the EU's last-Sunday-of-March, two days apart, below the group derivation's sampling step.

## Tests

`for_viewer/1` (scope, full user, partial map, nil, blank), `date_start/2` (per-date DST, legacy and fractional offsets, blank/nil/junk, the three midnight-switch cases, agreement with `day_start/2`), `local_date/2`, the blank-value resolver. Full suite 4517/0 on the tree merged with 2.15.0; `mix precommit` green.

Related module PRs: [calendar #6](https://github.com/BeamLabEU/phoenix_kit_calendar/pull/6), [bookings #4](https://github.com/BeamLabEU/phoenix_kit_bookings/pull/4), [crm #33](https://github.com/BeamLabEU/phoenix_kit_crm/pull/33), [publishing #45](https://github.com/BeamLabEU/phoenix_kit_publishing/pull/45), [posts #18](https://github.com/BeamLabEU/phoenix_kit_posts/pull/18), [newsletters #33](https://github.com/BeamLabEU/phoenix_kit_newsletters/pull/33) — same branch name `pr/timezone-per-instant` on each fork; none depends on this PR (they use pre-2.0 helpers).

## Part two: V185 — the settings history lives in the activity feed, permanently

The question the first part could not answer: a stored instant does not say which regime wrote it, because nothing recorded when `time_zone` changed or what it was before (`date_updated` holds the last change only; no settings writer logged to the activity feed, which prunes after 90 days anyway).

The feed is the right home for that answer — it already holds who did what to which resource, with a before/after convention in its metadata and an admin page — once two things hold, and V185 makes them hold:

- **`phoenix_kit_activities.permanent`** (boolean, default false): an entry the pruner never deletes, for records rather than news. Any module may keep an entry this way through `Activity.log/1`; settings changes are the first.
- **`inserted_at` to the microsecond** (`timestamp(0)` → `timestamp`, no rewrite): "what was X at that instant" walks a resource's entries by time, and two changes inside one second must not read as simultaneous.
- `phoenix_kit_posts.time_zone` (varchar(64), nullable) — the one core-owned table with a typed wall clock; the posts PR writes it.

`PhoenixKit.Settings.History` records a permanent `setting.changed` entry for every settings write that changes a value — the key, the value before and after, the source in `metadata`, the actor and mode on the entry, the setting row as the resource — written at the query layer inside the same transaction as the write, so every public writer records (single, batch, module and boolean forms, JSON, the admin pages) and an unchanged save records nothing. The value before is read under a row lock inside the transaction (two racing writers cannot both record the same old value). A restricted setting records that a change happened with both values withheld, and `value_at/2` answers nil for it at every instant. `Settings.history/2` and `Settings.value_at/2` ("what was this setting at that instant"; a `DateTime` in any zone is the instant it names; a JSON setting is its document). The Activity page shows settings changes beside everything else, no new UI.

Manifest: two objects declared and one revision appended by hand from a chain-built database, `chain_hash` restamped; the four-file manifest gate and the full suite with integration are green. **Renumbered V184 → V185 after #785 merged with its own V184** (merge commit `44d0f01e`: the chain re-run into a fresh database V135→V185, the manifest and chain hash follow, the full suite green).

Not here: the user's own timezone changes — the feed already records `user_timezone_from`/`_to` on a profile change, and the module rows now carry their zone. Reviewed by codex, kimi and Gemini 3.1 Pro on the way (the locked old value, the UTC-normalised instant, the withheld secret on the no-history path, microseconds, the error and batch result shapes).


ddon:
@mdon conflict here, please resolve

PR #789, opened 2026-09-07
Core: recognize catalogue-owned tables in storage/doctor, document module table ownership

timujinne:
## Summary

- Storage's orphaned-file check now recognizes the catalogue module's JSONB-embedded image references on all three catalogue tables that go through `PhoenixKitCatalogue.Attachments`: `phoenix_kit_cat_items.data->>'featured_image_uuid'` / `data->'media_order'` / `data->'ecommerce'->>'file_uuid'`, `phoenix_kit_cat_categories.data->>'featured_image_uuid'` / `data->'media_order'` / `data->'ecommerce'->>'image_uuid'`, and `phoenix_kit_cat_catalogues.data->>'featured_image_uuid'` / `data->'media_order'` — gated on the tables actually existing. Without this, a live catalogue item, category, or catalogue-record image would be classified as an orphan and queued for deletion.
- Documented (comment only, no behavior change) why doctor's V56 NULL-uuid check deliberately does not include the catalogue tables: their uuid column has carried `DEFAULT uuid_generate_v7() NOT NULL` since creation, so they never went through the backfill that check exists to catch.
- Added an inventory section to `dev_docs/guides/2026-09-05-module-table-extraction-guide.md` listing which core-baseline tables are already owned by `phoenix_kit_ecommerce` (10 tables), `phoenix_kit_catalogue` (18 tables), and `phoenix_kit_entities` (2 tables) — informational only, ahead of a future baseline squash. No drop, no migration change.
- Checked `lib/modules/sitemap/generator.ex`'s source registry — it already picks up module-registered sources generically, no change needed there.

## Test plan

- [x] `test/integration/storage_catalogue_orphan_test.exs` (11 tests) — each JSONB reference shape (featured image, media order, and the ecommerce-extension namespace on items/categories) protects a real file from `Storage.file_orphaned?/1`; negative cases confirm an unrelated file still reads as orphaned.
- [x] `mix format` — clean
- [x] `mix compile --warnings-as-errors` — clean
- [x] `mix credo --strict` on changed files — no issues
- [x] `mix test` — 42 doctests, 4754 tests, 0 failures (fresh scratch DB; the previously reported "3 pre-existing failures" in `phoenix_kit_doctor_orphaned_fk_destructive_test.exs` did not reproduce — that file alone also runs clean, 4 tests, 0 failures)

### DIFF
--- a/lib/modules/storage/storage.ex
+++ b/lib/modules/storage/storage.ex
@@ -2158,6 +2158,29 @@ defmodule PhoenixKit.Modules.Storage do
            f.uuid
          )
        )},
+      # The catalogue module (phoenix_kit_catalogue) stores its image
+      # references inside the JSONB `data` column rather than dedicated
+      # columns — a plain join would miss them entirely, and a live
+      # catalogue item/category image would look orphaned and get queued
+      # for deletion by DeleteOrphanedFileJob.
+      {"phoenix_kit_cat_items",
+       dynamic(
+         [f],
+         fragment(
+           "NOT EXISTS (SELECT 1 FROM phoenix_kit_cat_items ci WHERE ci.data->>'featured_image_uuid' = ?::text) AND NOT EXISTS (SELECT 1 FROM phoenix_kit_cat_items ci WHERE ci.data->'media_order' @> to_jsonb(ARRAY[?::text])) AND NOT EXISTS (SELECT 1 FROM phoenix_kit_cat_items ci WHERE ci.data->'ecommerce'->>'file_uuid' = ?::text)",
+           f.uuid,
+           f.uuid,
+           f.uuid
+         )
+       )},
+      {"phoenix_kit_cat_categories",
+       dynamic(
+         [f],
+         fragment(
+           "NOT EXISTS (SELECT 1 FROM phoenix_kit_cat_categories cc WHERE cc.data->'ecommerce'->>'image_uuid' = ?::text)",
+           f.uuid
+         )
+       )},
       {"phoenix_kit_publishing_contents",
        dynamic(
          [f],
--- a/lib/modules/storage/storage.ex
+++ b/lib/modules/storage/storage.ex
@@ -2160,9 +2160,9 @@ defmodule PhoenixKit.Modules.Storage do
        )},
       # The catalogue module (phoenix_kit_catalogue) stores its image
       # references inside the JSONB `data` column rather than dedicated
-      # columns — a plain join would miss them entirely, and a live
-      # catalogue item/category image would look orphaned and get queued
-      # for deletion by DeleteOrphanedFileJob.
+      # columns — a plain join would miss them entirely, and a
+      # live catalogue item, category, or catalogue-record image would
+      # look orphaned and get queued for deletion by DeleteOrphanedFileJob.
       {"phoenix_kit_cat_items",
        dynamic(
          [f],
@@ -2177,7 +2177,18 @@ defmodule PhoenixKit.Modules.Storage do
        dynamic(
          [f],
          fragment(
-           "NOT EXISTS (SELECT 1 FROM phoenix_kit_cat_categories cc WHERE cc.data->'ecommerce'->>'image_uuid' = ?::text)",
+           "NOT EXISTS (SELECT 1 FROM phoenix_kit_cat_categories cc WHERE cc.data->>'featured_image_uuid' = ?::text) AND NOT EXISTS (SELECT 1 FROM phoenix_kit_cat_categories cc WHERE cc.data->'media_order' @> to_jsonb(ARRAY[?::text])) AND NOT EXISTS (SELECT 1 FROM phoenix_kit_cat_categories cc WHERE cc.data->'ecommerce'->>'image_uuid' = ?::text)",
+           f.uuid,
+           f.uuid,
+           f.uuid
+         )
+       )},
+      {"phoenix_kit_cat_catalogues",
+       dynamic(
+         [f],
+         fragment(
+           "NOT EXISTS (SELECT 1 FROM phoenix_kit_cat_catalogues ct WHERE ct.data->>'featured_image_uuid' = ?::text) AND NOT EXISTS (SELECT 1 FROM phoenix_kit_cat_catalogues ct WHERE ct.data->'media_order' @> to_jsonb(ARRAY[?::text]))",
+           f.uuid,
            f.uuid
          )
        )},
--- a/dev_docs/guides/2026-09-05-module-table-extraction-guide.md
+++ b/dev_docs/guides/2026-09-05-module-table-extraction-guide.md
@@ -39,3 +39,64 @@ example `phoenix_kit_shop_products` once the shop reads products from
 `phoenix_kit_catalogue`) are marked with a `COMMENT ON TABLE … 'deprecated
 <date>: …'` by the host app, never dropped by anyone in this cycle; the
 squash of Phase 2 decides their fate.
+
+## Inventory: which core-baseline tables are already module-owned
+
+Each module below is at Phase 0 (adopted, `down/1` only unstamps) or
+already past it. Core's baseline still creates every table listed — this
+inventory is what a future baseline squash removes from core once all
+hosts have adopted the owning module's V1, not something this PR drops or
+migrates. Only the module's designated version table carries the
+`COMMENT ON TABLE … '<ns>_schema:<N>'` marker; the rest of a module's
+tables are recognized by ownership (this list), not by a per-table
+comment.
+
+**`phoenix_kit_ecommerce`** — marker `pke_schema:<N>` on
+`phoenix_kit_shop_config` (`PhoenixKitEcommerce.Migrations.version_table/0`).
+Ten tables, all `phoenix_kit_shop_*`:
+
+* `phoenix_kit_shop_config`
+* `phoenix_kit_shop_shipping_methods`
+* `phoenix_kit_shop_categories`
+* `phoenix_kit_shop_products`
+* `phoenix_kit_shop_product_slugs`
+* `phoenix_kit_shop_category_slugs`
+* `phoenix_kit_shop_carts`
+* `phoenix_kit_shop_cart_items`
+* `phoenix_kit_shop_import_configs`
+* `phoenix_kit_shop_import_logs`
+
+**`phoenix_kit_catalogue`** — marker `pkc_schema:<N>` on
+`phoenix_kit_cat_catalogues` (`PhoenixKitCatalogue.Migrations.version_table/0`).
+Eighteen tables, all `phoenix_kit_cat_*`:
+
+* `phoenix_kit_cat_catalogues`
+* `phoenix_kit_cat_folders`
+* `phoenix_kit_cat_categories`
+* `phoenix_kit_cat_manufacturers`
+* `phoenix_kit_cat_suppliers`
+* `phoenix_kit_cat_manufacturer_suppliers`
+* `phoenix_kit_cat_items`
+* `phoenix_kit_cat_item_catalogue_rules`
+* `phoenix_kit_cat_item_supplier_info`
+* `phoenix_kit_cat_pdfs`
+* `phoenix_kit_cat_pdf_pages`
+* `phoenix_kit_cat_pdf_page_contents`
+* `phoenix_kit_cat_pdf_extractions`
+* `phoenix_kit_cat_attribute_groups`
+* `phoenix_kit_cat_attributes`
+* `phoenix_kit_cat_attribute_values`
+* `phoenix_kit_cat_item_attribute_groups`
+* `phoenix_kit_cat_item_attribute_sets`
+
+**`phoenix_kit_entities`** — marker `pkn_schema:<N>` on
+`phoenix_kit_entities` itself (`PhoenixKitEntities.Migrations.version_table/0`).
+Two tables:
+
+* `phoenix_kit_entities`
+* `phoenix_kit_entity_data`
+
+Storage's orphaned-file check and doctor's NULL-uuid check are the two
+places that have historically hardcoded a table list next to this kind of
+inventory; see their own comments for why each one currently does or does
+not need the catalogue tables added — the answer is not the same for both.
--- a/lib/mix/tasks/phoenix_kit.doctor.ex
+++ b/lib/mix/tasks/phoenix_kit.doctor.ex
@@ -944,7 +944,13 @@ defmodule Mix.Tasks.PhoenixKit.Doctor do
     repo = get_repo!()
     escaped_prefix = String.replace(prefix, "'", "\\'")
 
-    # Key FK source tables whose uuid column must not be NULL
+    # Key FK source tables whose uuid column must not be NULL.
+    #
+    # `phoenix_kit_cat_items`/`phoenix_kit_cat_categories` (catalogue module)
+    # are deliberately absent: their uuid column has carried `DEFAULT
+    # uuid_generate_v7() NOT NULL` since their V135 baseline creation, so
+    # they never went through the V56 backfill this check exists to catch
+    # and cannot hold a NULL uuid.
     source_tables = [
       "phoenix_kit_users",
       "phoenix_kit_user_roles",
--- a/test/integration/storage_catalogue_orphan_test.exs
+++ b/test/integration/storage_catalogue_orphan_test.exs
@@ -3,11 +3,12 @@ defmodule PhoenixKit.Integration.StorageCatalogueOrphanTest do
   The orphan-file check must see catalogue references.
 
   The catalogue module (`phoenix_kit_catalogue`) stores its image references
-  inside the JSONB `data` column of `phoenix_kit_cat_items` /
-  `phoenix_kit_cat_categories` rather than dedicated FK columns, so a plain
-  join misses them — without the JSONB-aware checks added alongside this
-  test, a live catalogue item/category image would be classified as an
-  orphan and queued for deletion by `DeleteOrphanedFileJob`.
+  inside the JSONB `data` column of `phoenix_kit_cat_items`,
+  `phoenix_kit_cat_categories`, and `phoenix_kit_cat_catalogues` rather than
+  dedicated FK columns, so a plain join misses them — without the
+  JSONB-aware checks added alongside this test, a live catalogue item,
+  category, or catalogue-record image would be classified as an orphan and
+  queued for deletion by `DeleteOrphanedFileJob`.
   """
   use PhoenixKit.DataCase, async: false
 
@@ -35,13 +36,16 @@ defmodule PhoenixKit.Integration.StorageCatalogueOrphanTest do
     file
   end
 
-  defp insert_catalogue! do
+  defp insert_catalogue!(data \\ %{}) do
     %{rows: [[uuid]]} =
-      Repo.query!("""
-      INSERT INTO phoenix_kit_cat_catalogues (uuid, name, inserted_at, updated_at)
-      VALUES (gen_random_uuid(), 'Test catalogue', now(), now())
-      RETURNING uuid
-      """)
+      Repo.query!(
+        """
+        INSERT INTO phoenix_kit_cat_catalogues (uuid, name, data, inserted_at, updated_at)
+        VALUES (gen_random_uuid(), 'Test catalogue', $1, now(), now())
+        RETURNING uuid
+        """,
+        [data]
+      )
 
     uuid
   end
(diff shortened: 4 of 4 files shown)


lib/modules/storage/storage.ex
```diff
@@ -2158,6 +2158,29 @@ defmodule PhoenixKit.Modules.Storage do
...
+           f.uuid,
+           f.uuid,
+           f.uuid
+         )
+       )},
+      {"phoenix_kit_cat_categories",
+       dynamic(
+         [f],
+         fragment(
+           "NOT EXISTS (SELECT 1 FROM phoenix_kit_cat_categories cc WHERE cc.data->'ecommerce'->>'image_uuid' = ?::text)",
```
timujinne: This only covers the ecommerce extension's nested `data->'ecommerce'->>'image_uuid'`. Categories also carry a plain `data->>'featured_image_uuid'` and a `data->'media_order'` gallery via the same generic `PhoenixKitCatalogue.Attachments` flow items use (`category_form_live.ex`), so a category's own featured image is still classified as orphaned. Mirror the item fragment above: check both keys here as well.

lib/modules/storage/storage.ex
```diff
@@ -2158,6 +2158,29 @@ defmodule PhoenixKit.Modules.Storage do
...
+           f.uuid
+         )
+       )},
```
timujinne, line 2194: Missing a sibling entry for `phoenix_kit_cat_catalogues`: it has the identical featured-image/gallery mechanism (`catalogue_form_live.ex`), and this PR's own doc lists it as the module's version-marker table, so it is the first table a maintainer will expect to see here.

timujinne (commented):
## Review

## Findings

### 1. CRITICAL — `lib/modules/storage/storage.ex:2176-2183` — the new `phoenix_kit_cat_categories` check only covers the ecommerce-extension image, not the resource's own plain featured image / gallery

```elixir
{"phoenix_kit_cat_categories",
 dynamic(
   [f],
   fragment(
     "NOT EXISTS (SELECT 1 FROM phoenix_kit_cat_categories cc WHERE cc.data->'ecommerce'->>'image_uuid' = ?::text)",
     f.uuid
   )
 )},
```

`PhoenixKitCatalogue.Attachments` (the module's generic file-attachment layer) reads/writes `data["featured_image_uuid"]` and `data["media_order"]` for *any* resource passed to `mount_attachments/2` — and `category_form_live.ex` wires `open_featured_image_picker`, `clear_featured_image`, and `handle_reorder_files` exactly like `item_form_live.ex` does. So a category's own featured image or gallery file (as opposed to the ecommerce extension's nested `data["ecommerce"]["image_uuid"]`) is still invisible to this check and will be classified as orphaned and queued for `DeleteOrphanedFileJob` — the exact failure mode the PR says it fixes for "a live catalogue item/category image."

Fix: add `data->>'featured_image_uuid'` and `data->'media_order'` checks to the categories fragment, mirroring what was done for `phoenix_kit_cat_items`.

### 2. CRITICAL — `lib/modules/storage/storage.ex` — `phoenix_kit_cat_catalogues` has no orphan check at all

`catalogue_form_live.ex` wires the identical `Attachments.mount_attachments/open_featured_image_picker/clear_featured_image/handle_reorder_files` flow as items and categories, against the same `data` JSONB column (`phoenix_kit_cat_catalogues.data jsonb DEFAULT '{}'::jsonb`, confirmed in `v135.ex` and `PhoenixKitCatalogue.Schemas.Catalogue`). `phoenix_kit_cat_catalogues` is even the table this PR's own new doc inventory names as the catalogue module's version-marker table — but no `{"phoenix_kit_cat_catalogues", ...}` entry was added to `optional_checks`. A catalogue's featured image/gallery file has exactly the same orphan-deletion exposure as items and categories.

Fix: add an entry for `phoenix_kit_cat_catalogues` checking `data->>'featured_image_uuid'` / `data->'media_order'`.

### 3. MINOR — Test Plan's "3 pre-existing failures" claim did not reproduce

The PR body states: *"`mix test` — 42 doctests, 4749 tests, 3 pre-existing failures (verified via `git stash` unrelated to this change, in `phoenix_kit_doctor_orphaned_fk_destructive_test.exs`)."* I ran the full suite from a clean scratch DB (same counts: 42 doctests, 4749 tests) and got **0 failures**; that file alone also ran clean (4 tests, 0 failures). Either the 3 failures are non-deterministic/environment-dependent, or the claim is stale. Doesn't block this PR, but as written the Test Plan overstates a pre-existing problem that isn't reproducible as described.

### 4. NITPICK — commit headline doesn't follow the stated convention

AGENTS.md: *"Commit messages: start with `Add`, `Update`, `Fix`, `Remove`, `Merge`."* This PR's headline is "Recognize catalogue-owned JSONB image references as module table sources." In practice this convention is loosely enforced — several recent commits on `main` (e.g. "Extend tabs to Organization settings", "Move personal integrations off Settings...") also don't follow it — so treat as low severity, but noting it since the doc is explicit.

### 5. NITPICK — `lib/mix/tasks/phoenix_kit.doctor.ex:949-953` comment's reasoning also applies to a third table it doesn't mention

The new comment explains why `phoenix_kit_cat_items`/`phoenix_kit_cat_categories` are absent from `source_tables`. `phoenix_kit_cat_catalogues` has the identical `uuid DEFAULT uuid_generate_v7() NOT NULL` since V135 and is also an FK target (categories/items reference it), so the same reasoning covers it too — it's just not mentioned. Not a defect (the list was already curated/non-exhaustive before this PR), just an easy one-word addition for completeness since the comment is being written now anyway.

## Checked and found clean

- **Table inventory** (`dev_docs/guides/2026-09-05-module-table-extraction-guide.md`): all 10 `phoenix_kit_ecommerce`, 18 `phoenix_kit_catalogue`, and 2 `phoenix_kit_entities` table names verified against each module's Ecto schemas (`schema "phoenix_kit_..."` grep on `upstream/main`) and cross-checked against core's own `lib/phoenix_kit/migrations/postgres/v135.ex` + `expected_schema.ex` — exact match, nothing missing or extra.
- **Version-marker claims** (`pke_schema`/`pkc_schema`/`pkn_schema`, `version_table/0`, marker tables `phoenix_kit_shop_config` / `phoenix_kit_cat_catalogues` / `phoenix_kit_entities`) verified against each module's `lib/.../migrations.ex` on a **freshly fetched** `upstream/main` — exact match (`@marker_prefix`, `@version_table`, `def version_table`). Note for future audits: this container's cached local `main` branches for these three module repos were weeks stale (e.g. `phoenix_kit_ecommerce`'s local `main` was 3 weeks behind and missing this exact mechanism) — always `git fetch upstream main` fresh before trusting a local checkout here.
- **doctor.ex NULL-uuid comment**: `DEFAULT uuid_generate_v7() NOT NULL` on both `phoenix_kit_cat_items` and `phoenix_kit_cat_categories` confirmed in `v135.ex`. Comment-only change, no logic touched.
- **storage.ex mechanism**: new entries reuse the existing generic `existing_optional_tables/0` (`information_schema` lookup on `LIKE 'phoenix_kit_%'`) — no hardcoded existence list, no new compile-time or runtime dependency on `phoenix_kit_catalogue` (`mix.exs` has zero references to it, confirmed by grep). Matches the file's established `{table, dynamic}` pattern; SQL is parameterized (`fragment(..., f.uuid, ...)`), no injection risk.
- **Item coverage** (`featured_image_uuid`, `media_order`, `ecommerce.file_uuid`) matches `PhoenixKitCatalogue.Attachments`' actual write paths on `upstream/main`.
- **`ecommerce.file_uuid`/`image_uuid` namespace**: confirmed written by `phoenix_kit_ecommerce`'s catalogue extension (`catalogue/shop_sections.ex`, `product_source/catalogue/view.ex`, `web/category_form.ex`) — currently only on unmerged feature branches, not yet in `phoenix_kit_ecommerce`'s own `main` (see Not verified).
- `mix compile --warnings-as-errors` — clean.
- `mix credo --strict` on the three changed lib/test files — no issues.
- `mix test test/integration/storage_catalogue_orphan_test.exs` — 6/6 pass; assertions match the actual write shapes.
- Full suite (`mix test`, fresh scratch DB, `MIX_ENV=test`) — **42 doctests, 4749 tests, 0 failures**.
- No unrelated files touched. No CHANGELOG/version bump — consistent with repo convention (individual PRs don't bump these; only periodic "Release X.Y.Z" commits do — confirmed in `git log upstream/main`).
- No `Co-Authored-By` or tool-attribution trailers in the commit.
- Core (`mix.exs`) has no dependency on `phoenix_kit_catalogue`/`ecommerce`/`entities` — the change stays runtime-only, keeping core decoupled from optional modules.

## Not verified

- Whether `phoenix_kit_ecommerce`'s catalogue-extension branches (which define the `ecommerce.file_uuid`/`image_uuid` namespace this PR protects) are actually deployed on any live host today — confirmed they exist on feature branches (`feature/catalogue-extension`, `feature/catalogue-product-source`, `feature/catalogue-filters`), not yet on `phoenix_kit_ecommerce`'s `main`.
- Dialyzer — not run (small, type-shape-unremarkable diff; low risk, but not empirically checked here).
- Root cause of the PR body's claimed "3 pre-existing failures" — couldn't reproduce them to confirm or refute.

## Out of scope (pre-existing, not introduced by this PR)

- `phoenix_kit_cat_pdfs.file_uuid` and `phoenix_kit_cat_pdf_pages.file_uuid` are real FK columns (not JSONB) pointing at Storage files, but neither table appears in `orphaned_files_query`'s `optional_checks`. A PDF attached to a catalogue item looks like it can already be queued for deletion as an orphan today — pre-existing, unrelated to this diff, flagging for awareness only.

## Verdict: CHANGES REQUESTED

The fix doesn't fully deliver on its own stated goal: a category's plain featured image/gallery file, and any catalogue's featured image/gallery file, remain classified as orphaned and eligible for `DeleteOrphanedFileJob` deletion after this PR merges. Before this leaves draft:

1. Add `data->>'featured_image_uuid'` + `data->'media_order'` checks to the `phoenix_kit_cat_categories` fragment.
2. Add a `phoenix_kit_cat_catalogues` entry with the same two checks.
3. Add test cases for both (mirroring the existing item tests).
4. Optionally reconcile or drop the "3 pre-existing failures" claim in the Test Plan.

timujinne (commented):
## Re-check after `9693db0` — findings 1–3 resolved ✅

Verified against the PR head and the catalogue module's actual writer (`PhoenixKitCatalogue.Attachments`), not against the commit message.

- **Categories** — the `phoenix_kit_cat_categories` fragment now checks `data->>'featured_image_uuid'`, `data->'media_order' @> to_jsonb(ARRAY[?::text])` and `data->'ecommerce'->>'image_uuid'`, in the same `NOT EXISTS … AND` shape, operators and `?::text` casts as the items fragment above it. Resolved.
- **Catalogues** — new `phoenix_kit_cat_catalogues` entry with the featured-image and `media_order` checks. Catalogues have no ecommerce image key (confirmed against the ecommerce module's catalogue-extension branches, which only reference `item[ecommerce][file_uuid]` and `category[ecommerce][image_uuid]`), so the two-key entry is correct. Resolved.
- **`media_order` semantics** — `Attachments.inject_attachment_data/2` persists `media_order` as a JSONB array of uuid strings for item, category and catalogue forms alike; `@> to_jsonb(ARRAY[?::text])` is true iff the uuid is an element, and absent key / JSON `null` / `[]` simply do not match. Same idiom the file already uses for `phoenix_kit_project_portal_submissions`.
- **Tests** — `storage_catalogue_orphan_test.exs` grows from 6 to 11: positive cases for both keys on categories and catalogues (media tests include a second unrelated uuid, exercising subset containment) plus a catalogue negative. At the parent commit the categories fragment checked only the ecommerce key and catalogues had no entry, so all four positives fail without the fix. Standalone: 11 tests, 0 failures; full suite on a fresh DB: 42 doctests, 4754 tests, 0 failures.
- **Test Plan** — the "3 pre-existing failures" claim is corrected in the PR body to the reproducible numbers.
- Table and key names are compile-time literals; the only bound value is `f.uuid`. No injection surface.

`mix compile --warnings-as-errors`, `mix format --check-formatted`, `mix credo --strict` clean; the repo's pre-commit hook passed on the commit. One commit, subject per `AGENTS.md`, no tool-attribution trailers.

### Follow-ups (not blocking)

- No expression index supports `(data->>'featured_image_uuid')` or `(data->'media_order')` on the cat tables; one orphan query now does up to 8 sequential scans of them and `DeleteOrphanedFileJob` runs it per file. Fine as a backstop on modest tables; worth a ticket for large catalogues.
- Protection is pointer-based: records created before the catalogue module started writing `media_order` hold gallery files only via folder membership, which this query does not treat as a reference (pre-existing). A host running `mix phoenix_kit.cleanup_orphaned_files --delete` right after upgrading could sweep legacy gallery files; one warning line in the PR body would help.

LGTM — ready to leave draft.

timujinne (commented):
# Review: phoenix_kit#789 — "Recognize catalogue-owned JSONB image references as module table sources"

Branch `feature/module-table-sources` (4fbb34a2) on top of `main` (3ac3a406, release 2.20.0). Diff: 4 files, +220/-1.

## Stage 1: Spec Compliance

The spec asked for exactly three things. All three are present and none of the diff exceeds them (4 files touched, no sitemap changes).

1. **Storage orphan-file check** (`lib/modules/storage/storage.ex:2161-2183`) — two new entries added to `optional_checks`, gated by the existing `existing_optional_tables/0` presence check, following the identical pattern used for `phoenix_kit_project_portal_submissions`/`phoenix_kit_shop_products`/etc. immediately above them. **PASS** — verified against the live dev database (`dev_decor3dprint_db`):
   - `phoenix_kit_cat_items.data->>'featured_image_uuid'`, `data->'media_order'` (jsonb array of uuid strings, confirmed via `jsonb_typeof` and sample rows — 640/665 items carry it), and `data->'ecommerce'->>'file_uuid'` are all real, populated keys.
   - `phoenix_kit_cat_categories.data->'ecommerce'->>'image_uuid'` is confirmed to be the correct path — categories have no top-level `image_uuid` key; `ecommerce.image_uuid` exists (7/20 categories carry the key, currently all `null` in this dataset, but the path is correct and schema-verified).
   - The implementer's claim that there's no top-level `image_uuid` on categories is correct.

2. **`phoenix_kit.doctor`** (`lib/mix/tasks/phoenix_kit.doctor.ex:943-955`) — deliberately left untouched except for an explanatory comment. **PASS**, and the reasoning is verified, not just asserted:
   - `phoenix_kit_cat_items`/`phoenix_kit_cat_categories` are created in `lib/phoenix_kit/migrations/postgres/v135.ex:1672` and `:1687` with `"uuid" uuid DEFAULT #{p}uuid_generate_v7() NOT NULL` inline in the `CREATE TABLE`.
   - `V56` is confirmed (via `lib/phoenix_kit/migrations/uuid_fk_columns.ex`) to be the batched-backfill migration `check_null_uuids/1`'s moduledoc refers to, and it predates V135 by construction — cat tables never went through it.
   - Because the column carries a DB-level `NOT NULL` constraint from creation, no NULL value can ever exist in it regardless of default expression — adding these tables to `source_tables` would be permanently-dead code that can never fire. Omitting them is correct, not just harmless.

3. **Guide inventory** (`dev_docs/guides/2026-09-05-module-table-extraction-guide.md`) — cross-checked line-by-line against the actual `CREATE TABLE IF NOT EXISTS` statements in core (`v135.ex`, `v149.ex`, `v173.ex`, `v177.ex`, `shop_slug_projection.ex`/V171) and against each module's own `migrations.ex` (`phoenix_kit_ecommerce`, `phoenix_kit_catalogue`, `phoenix_kit_entities`). **PASS, exact match**:
   - Ecommerce: 10 tables (8 from v135 baseline + `phoenix_kit_shop_product_slugs`/`phoenix_kit_shop_category_slugs` from V171's `shop_slug_projection.ex`), marker `pke_schema:` on `phoenix_kit_shop_config` — matches `@marker_prefix "pke_schema:"` / `@version_table "phoenix_kit_shop_config"` in the fork.
   - Catalogue: 18 tables, set-equal to the fork's `migrations.ex` CREATE TABLE list, marker `pkc_schema:` on `phoenix_kit_cat_catalogues` — matches.
   - Entities: 2 tables, marker `pkn_schema:` on `phoenix_kit_entities` — matches.
   - No table is missing or extra in any of the three lists.

No `EXTRA`/`WRONG`/`AMBIGUOUS` issues found. One `EDGE_CASE` worth flagging as a should-fix, not a spec gap:

- [MINOR / should-fix] `lib/mix/tasks/phoenix_kit.doctor.ex:959-962` — the pre-existing (not touched by this PR) `source_tables` list still includes `phoenix_kit_shop_carts`, `phoenix_kit_shop_products`, `phoenix_kit_shop_categories`, `phoenix_kit_shop_shipping_methods`. Those are also created in the same V135 baseline with the identical `uuid DEFAULT uuid_generate_v7() NOT NULL` inline definition (verified: `v135.ex:1169,1194,1221,1239,1277` all carry the same clause), so by the PR's own stated logic these four entries are equally dead code — the check can never fire for them either. This is pre-existing, not introduced by this PR, and the spec explicitly scoped this task to catalogue tables only, so it is out of scope to fix here — flagging only so a follow-up cleanup isn't lost.

**Spec Verdict:** PASS

---

## Stage 2: Code Quality

### Storage check correctness (data-risk review)

Ran targeted edge cases against the live dev database and the new integration test:

- Item with only `featured_image_uuid` set, empty/absent `media_order` → protected. Verified live: 640/665 items carry `featured_image_uuid`; matches product code path.
- Item with a file only inside `media_order` (array containment via `@> to_jsonb(ARRAY[?::text])`) → protected. Verified: `media_order` is confirmed to be a plain JSON array of UUID strings live (sampled 3 rows with 9-10 elements each), same containment idiom already used for `phoenix_kit_project_portal_submissions.file_uuids` two entries above — reuses a proven pattern rather than inventing a new one.
- Category with `ecommerce.image_uuid` set → protected; verified live that this is the real (nested, not top-level) path.
- Soft-deleted item/category (catalogue uses `status = "deleted"`, not row deletion — confirmed in `phoenix_kit_catalogue/catalogue.ex`) still keeps its row and JSONB data, so its image stays protected even after a soft-delete. This matches the existing, deliberately lenient behavviour of every other check in this list (e.g. `phoenix_kit_shop_products`, `phoenix_kit_posts` — none of them filter by status either): correct by consistency, and safe (errs toward not deleting a file, never toward deleting a live one).
- All four new checks are correctly gated by `existing_optional_tables/0` (table-existence check via `information_schema.tables`), so a host without the catalogue module installed sees no behavior change and no query failure.
- Confirmed these are wired into the same `Enum.reduce(optional_checks, base, ...)` fold as the pre-existing checks — not a dead branch.

No CRITICAL or MAJOR issues. No SQL injection risk: `f.uuid` is always parameterized through Ecto's `dynamic/2` + `fragment/…` placeholders (`?`), never string-interpolated.

### Test coverage

`test/integration/storage_catalogue_orphan_test.exs` (129 lines, 6 tests) directly exercises:
- `featured_image_uuid` protects a file.
- `media_order` protects a file.
- `ecommerce.file_uuid` protects a file.
- A different item's `featured_image_uuid` does NOT protect an unrelated file (negative case).
- Category `ecommerce.image_uuid` protects a file.
- A different category's `ecommerce.image_uuid` does NOT protect an unrelated file (negative case).

Ran in isolation: **6 tests, 0 failures** (verified live — see Verification section).

Missing test case (should-fix, not blocking): no test for a category/item whose `data` JSONB is `{}` (the column default) or absent-key case — the fragments use `->>`/`->` on possibly-missing keys, which is safe in Postgres (returns `NULL`, never errors), but an explicit test would document that the check degrades gracefully rather than relying on inference. Minor, not required for correctness.

### Style / idiom

- Fragment SQL matches the existing house style exactly (raw `NOT EXISTS (SELECT 1 FROM … WHERE …)` correlated subqueries, `?::text` casts, comments explaining the JSONB rationale) — no new pattern introduced, which is the right call for a one-off addition to an already-established list.
- Doctor comment is placed directly above the list it explains, consistent with how the file documents other design decisions inline.

**Quality Summary:** 0 critical, 0 major, 1 minor (pre-existing dead-code entries, out of scope), 1 nitpick (missing empty-JSONB test case)
**Quality Verdict:** Ship

---

## Verification

- `mix deps.get` in the PR worktree: clean, `mix.lock` untouched (`git status --short` empty afterward).
- `MIX_ENV=test mix test test/integration/storage_catalogue_orphan_test.exs`: **6 tests, 0 failures.**
- `MIX_ENV=test mix test` (full suite) against the shared container-wide `phoenix_kit_test` database initially reported 757 failures, all `column "permanent" of relation "phoenix_kit_activities" does not exist` — unrelated to this diff (that column belongs to V185, nothing this PR touches) and consistent with another concurrent process on this heavily multi-tenant container migrating/resetting that same shared database mid-run. Re-ran against a private, isolated database (`phoenix_kit_test_review789`, dropped afterward) to get a trustworthy signal:
  - Full suite, isolated DB, run 1: 42 doctests, 4749 tests, 1 failure (trace lost to a `tail` truncation in that specific invocation).
  - Full suite, isolated DB, run 2 (full output captured, nothing truncated): **42 doctests, 4749 tests, 0 failures.**
  - `test/mix/tasks/phoenix_kit_doctor_orphaned_fk_destructive_test.exs` alone (the file the implementer named as the site of their 3 failures): **4 tests, 0 failures.**
  - Conclusion: the implementer's exact failure count (3, all in the destructive doctor test) did not reproduce, but neither did anything the PR itself could plausibly cause — that test file is untouched by this diff (confirmed via `git blame`, last touched by an unrelated commit predating this branch) and passes cleanly alone and as part of a clean full run. The full suite has some cross-run flakiness independent of this PR (1 failure appeared once, 0 twice, in different files depending on shared-vs-isolated DB) — worth a separate look at suite hygiene, but it is not something this PR introduces or is responsible for fixing.
  - Did not run the `upstream/main` baseline comparison the task asked for — a temporary worktree's `mix compile` hung for 8+ minutes with no CPU-time progress and was killed; since the full suite already runs clean on the PR branch itself in isolation, and the flagged pre-existing-failure file passes standalone, the baseline comparison is not needed to reach a verdict here. Flagging this gap explicitly rather than fabricating a baseline result.
- Live database checks performed with a read-only `psql` connection against `dev_decor3dprint_db` (JSONB key enumeration and sample rows only — no writes).
- Cleanup done: private review database dropped, temporary `upstream/main` worktree removed, no stray files left in the PR worktree (`git status --short` clean).

---

## Overall Verdict: PASS

PR #790, opened 2026-09-07
Add non-destructive avatar cropping

alexdont:
Apple-Photos-style avatar cropping: the crop is **data, never pixels**. `custom_fields["avatar_crop"]` stores a focal point, zoom, and the image's aspect ratio; the uploaded file and all its variants stay untouched. Open the editor again and the crop is sitting where you left it — loosen it, recenter it, zoom back out, or reset to the full image, with nothing ever re-encoded and no quality lost. No migration: the key rides the users table's existing JSONB and is registered in `CustomFields.@internal_keys`.

**Why it's fast with many avatars on a page:** the storage module's variants are aspect-preserving scales of the original, so the normalized crop numbers hold against any variant. List pages keep loading the same `small` variants they load today — the crop is a few floats already inside the user record, applied as an inline style by the shared `user_avatar` component. No extra requests, no imaging jobs, no per-crop files. When a crop is zoomed, the component buys sharpness from the variant ladder instead (a 3× crop shows ⅓ of the source, so the same box needs a bigger variant), falling back to the original past `large`.

**The editor** lives in a modal on `/profile/settings`: drag to position, slider or wheel to zoom. The `AvatarCrop` hook's `avatarCropLayout` is a mirror of `PhoenixKit.Users.AvatarCrop.layout/1`, and both test suites pin the same numeric cases — the preview must render exactly what the avatar later renders. Gesture ends push the numbers to the LiveComponent, so Save persists what's on screen. Picking a different media file drops the old file's crop; OAuth/Gravatar avatars (no variants) are never cropped.

**Also fixes the blurry settings preview**: it was loading the 150px `thumbnail` variant into a 160px circle (320px on 2× displays). The preview now goes through the shared `user_avatar` component at a new `"xl"` size, which loads `medium` — same as the admin user form always did. The admin form's preview also applies the stored crop to the current avatar (a pending selection previews uncropped; its crop doesn't exist yet).

**Verification**
- `mix precommit` green: format, credo --strict, dialyzer, `mix test.js` (87 passing, 6 new for the crop geometry).
- Full suite against real PostgreSQL: **42 doctests, 4763 tests, 0 failures**.
- 20 new Elixir tests (geometry mirror cases, normalize clamping, variant ladder, rendered-HTML component tests), mutation-checked: crop-never-applied, zoom-never-sharpens, and xl-blurry-again all fail the suite.
- End-to-end in headless Chromium against a real stored image (1600×900 through `Auth.update_user_avatar`): logged in, opened the editor, zoomed to 2.5× and dragged; the saved preview's inline geometry matched the editor's to the rounding digit and the persisted map read `%{"ar" => 1.777…, "x" => 0.57, "y" => 0.56, "zoom" => 2.5}`.
- New UI strings extracted and translated in all seven locales (the two fuzzy guesses from the merge corrected).

No version bump — that call is yours.

### DIFF
--- a/lib/phoenix_kit/users/auth.ex
+++ b/lib/phoenix_kit/users/auth.ex
@@ -2103,7 +2103,30 @@ defmodule PhoenixKit.Users.Auth do
     existing_custom_fields = user.custom_fields || %{}
     merged_custom_fields = Map.merge(existing_custom_fields, attrs)
 
-    update_user_custom_fields(user, merged_custom_fields)
+    update_user_custom_fields(
+      user,
+      drop_stale_avatar_crop(merged_custom_fields, existing_custom_fields, attrs)
+    )
+  end
+
+  # A stored crop describes the file it was made against — its aspect ratio
+  # is baked into the map — so a crop must never outlive its file: applied
+  # to a different image it renders stretched and mis-framed everywhere.
+  # Enforced HERE, at the one merge every custom-fields write goes through,
+  # because every other enforcement point proved forgettable: the settings
+  # page cleared the crop on a new pick, while the admin form and
+  # update_user_avatar/4 did not. A caller that sets the crop alongside the
+  # file (the settings crop editor) is explicit and wins.
+  defp drop_stale_avatar_crop(merged, existing, attrs) do
+    changed_file? =
+      Map.has_key?(attrs, "avatar_file_uuid") and
+        Map.get(attrs, "avatar_file_uuid") != Map.get(existing, "avatar_file_uuid")
+
+    if changed_file? and not Map.has_key?(attrs, "avatar_crop") do
+      Map.put(merged, "avatar_crop", nil)
+    else
+      merged
+    end
   end
 
   @doc """
--- a/lib/phoenix_kit/users/avatar_crop.ex
+++ b/lib/phoenix_kit/users/avatar_crop.ex
@@ -1,0 +1,166 @@
+defmodule PhoenixKit.Users.AvatarCrop do
+  @moduledoc """
+  Non-destructive avatar cropping, Apple Photos style.
+
+  A crop is data, never pixels: `custom_fields["avatar_crop"]` holds a focal
+  point, a zoom, and the image's aspect ratio, while the original file — and
+  every generated variant of it — stays exactly as uploaded. Adjusting the
+  crop rewrites four numbers; resetting it drops the key; nothing is ever
+  re-encoded, so there is no quality loss and no way to crop yourself into a
+  corner.
+
+  Rendering applies the stored geometry as an inline style on the `<img>`
+  inside the avatar's (square, overflow-hidden) frame. Because the storage
+  module's image variants are aspect-preserving scales of the original, the
+  same normalized numbers are valid against any of them — a page full of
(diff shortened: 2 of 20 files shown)


alexdont:
Pushed a second commit (`be53e28b`) refining the flow per review: **picking a file now opens the crop editor before anything persists** — press Save untouched to accept the default fit, adjust first to land the file and its framing together, or Cancel to discard the pick with the old avatar unchanged. An untouched editor stores no crop at all (centered-at-cover-fit is what no crop renders anyway), and Reset hides on a fresh pick since the default already is the uncropped image.

Verified end-to-end in headless Chromium: pick → editor auto-opens → save untouched → `object-cover` markup with no stored crop; re-pick → **wheel-zoom to 2.16×** (the wheel zooms as well as the slider) → save → inline crop geometry on the preview and `%{"zoom" => 2.158…, "x" => 0.5, "y" => 0.5, "ar" => 1.777…}` persisted. Focused DB-backed suites (1439 tests) and the full precommit gate stay green.

alexdont:
Ran a multi-angle code review over this branch and pushed the fixes as `fe6c93e2`. Five real defects found and fixed, worst first:

1. **Stale crop outliving its file** — the admin user form and `Auth.update_user_avatar/4` replaced the avatar without clearing `avatar_crop`, so the old image's geometry (old aspect ratio included) stretched and mis-framed the new image on every surface. The invariant now lives at the one custom-fields merge every write goes through, DB-tested; explicit crop-with-file writes still win.
2. **Landscape avatars rendered soft** — `variant_for` compared needed pixels to variant *width*, but cover-fit sharpness is bounded by the image's *short side*. It now takes the aspect ratio into account, and is capped at `large` instead of falling through to serving the unbounded original into an avatar circle.
3. **Wheel-zoom / Save race** — the debounced wheel push could lose the final zoom if Save landed inside the 250ms window (worst case: a fresh pick saved uncropped after the preview showed 3×). The hook now flushes on the Save button's capture phase, so the final crop hits the wire before the phx-click. Verified by deliberately racing the window: editor 5.44, persisted 5.4365.
4. **Unguarded `open_avatar_crop`** could crash the LiveView via `signed_url(nil, …)` on a stale DOM click.
5. **Identity crops persisted on the adjust path** — zooming back to the default fit and saving now counts as a reset (`AvatarCrop.drop_identity/1`, public and tested), instead of storing a no-op crop that pinned the avatar to a recorded aspect ratio.

Plus cleanups from the same review: dead `data-image-url` attribute removed, duplicated `` branches collapsed, `getBoundingClientRect` once per drag, mount push skipped when the aspect ratio already matches, DOM ids on the crop buttons.

Known and deliberately not addressed here: the variant ladder mirrors the seeded dimension rows (deriving it from the operator-editable dimension config is the upgrade path, noted in a comment), and comment @-mention avatar chips still render uncropped (`comment_resources.avatar_thumb/1` feeds a background-image; happy to follow up if wanted).

Gates after the fixes: precommit clean (credo, dialyzer, 87 JS tests), 29 crop tests + 4 new DB-backed invariant tests green, mutation-checked (invariant-unenforced, ar-ignored, identity-never-dropped all fail the suite), and the full pick→crop→save E2E re-verified in headless Chromium.

PR #792, opened 2026-09-08
Public-page Edit link: documented helper API, component, per-module permission gate

timujinne:
## Summary

When an admin views a public page (a host LiveView, or a module's own public controller/LiveView such as publishing's), the host can show an "Edit" link into the matching admin page with one line in its layout.

`PhoenixKitWeb.AdminEditHelper.assign_admin_edit/3` already existed, but only the admin dashboard layout ever rendered the assigns it sets; on public pages every host re-implemented "is this an admin, build the admin URL" by hand.

- **Producer API** — `assign_admin_edit/3` is now documented as the public API. The third argument still accepts a string label (all existing callers unchanged) and also `[label: ..., permission: "publishing"]`. With `permission:` the scope must hold `Scope.has_module_access?/2` for that module key in addition to `Scope.can_access_admin_area?/1`, so an admin scoped to Publishing never gets an Edit link into Ecommerce. Non-admins get no assigns at all (absent, not nil).
- **Renderer** — new `PhoenixKitWeb.Components.Core.AdminEditLink.admin_edit_link/1`, imported alongside the other core components. `variant={:button}` (default, outlined button with the pencil icon) or `:menu_item` (daisyUI menu `<li>` for dropdowns). Renders nothing when `url` is nil, so a host writes `<.admin_edit_link url={assigns[:admin_edit_url]} label={assigns[:admin_edit_label]} />` unconditionally.
- `UserDashboardNav.user_dropdown/1` now renders its Edit item through the component.
- Guide: "Edit link on public pages" in `guides/integration.md`; CHANGELOG "Unreleased" entry.

Companion module PRs already use the helper with a string label (no dependency on this PR): BeamLabEU/phoenix_kit_publishing#46, BeamLabEU/phoenix_kit_ecommerce#41.

## Tests

- New `test/phoenix_kit_web/helpers/admin_edit_helper_test.exs` (conn and socket × no scope / non-admin / admin without the module permission / admin with it / string label / keyword opts) and `test/phoenix_kit_web/components/core/admin_edit_link_test.exs` (nil url, both variants, label fallback).
- `MIX_ENV=test mix test test/phoenix_kit_web/helpers test/phoenix_kit_web/components`: 695 tests, 0 failures. `mix format --check-formatted`, `mix credo --strict` clean. Full-suite and dialyzer results follow in a comment.

## Follow-ups noticed, not changed here

- `mix.exs` sets `plt_file: "priv/plts/dialyzer.plt"` under `def cli`, not `def project`, so dialyxir ignores it and rebuilds the default PLT.
- With `PGDATABASE=phoenix_kit_dev` exported, `LiveDatabaseGuard` correctly refuses to run tests; `unset PGDATABASE` first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,20 @@
+## Unreleased
+
+### Added
+
+- **Public "Edit link" API** — `PhoenixKitWeb.AdminEditHelper.assign_admin_edit/3`
+  is now a documented public API: a host LiveView or a module's own public
+  controller/LiveView calls it to declare a public page's matching admin
+  edit target. The label argument now also accepts a keyword list
+  (`label:`, `permission:`), gating the link on a specific module's
+  `Scope.has_module_access?/2` in addition to the existing admin-area check,
+  while every existing string-label call site keeps working unchanged. A new
+  `PhoenixKitWeb.Components.Core.AdminEditLink.admin_edit_link/1` component
+  (`<.admin_edit_link .../>`) renders the link — as a standalone button or a
+  `:menu_item` for dropdowns — and renders nothing when there is no URL, so
+  hosts can drop one line into their public layout unconditionally. See the
+  "Edit link on public pages" section in `guides/integration.md`.
+
 ## 2.22.0 - 2026-09-08
 
 ### Added
@@ -231,6 +248,7 @@
   "Province"; the real value from `beamlab_countries` is "Provinces and
   territories".
 
+||||||| parent of 8fbfa12992 (Make the public-page Edit link a documented API with a component and a permission gate)
 ## 2.19.0 - 2026-09-07
 
 ### Changed
--- a/guides/integration.md
+++ b/guides/integration.md
@@ -270,6 +270,57 @@ Scope.accessible_modules(scope)  # MapSet of granted permission keys
 
 **Route enforcement**: PhoenixKit's `phoenix_kit_ensure_admin` and `phoenix_kit_ensure_module_access` on_mount hooks automatically enforce permissions on admin routes. Sidebar navigation is gated per-user.
 
+### Edit link on public pages
+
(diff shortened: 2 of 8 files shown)


timujinne (commented):
Independent review of the branch (now rebased onto main 2.22.0 as 8b125e50d4; the rebase touched only the CHANGELOG position of the "Unreleased" block).

**Spec: PASS**, item by item:
- `AdminEditHelper.assign_admin_edit/3` keeps the string-label form; keyword form adds `label:`/`permission:`; with `permission:` the gate is `Scope.can_access_admin_area?/1` AND `Scope.has_module_access?/2` (the pure permission-set lookup, the same predicate `Dashboard.Tab.permission_granted?/2` uses). Non-admins get no assigns.
- `Components.Core.AdminEditLink.admin_edit_link/1`: nil url renders nothing, nil label falls back to gettext "Edit", `:button` and `:menu_item` variants, imported with the other core components in `PhoenixKitWeb`.
- `UserDashboardNav.user_dropdown/1` renders its Edit item through the component; `<.link navigate>` is consistent with every other link in that dropdown.
- Guide section and CHANGELOG entry match the API; no version bump.

**Verification**
- Helper + component suites: 18 tests, 0 failures (after the rebase). `mix format --check-formatted`, `mix credo --strict` clean. `mix dialyzer` on the pre-rebase tree: passed.
- Full suite on the pre-rebase tree: 4738 tests, 19 failures on the first pass, 5 on `--failed`, all in `test/integration/integrations_key_rotation_test.exs` (row-count assertions off by one: a second worktree was running its own suite against the same `phoenix_kit_test` database at the time). Not related to this change.

**Nits (non-blocking)**
- No dedicated test for the `user_dashboard_nav.ex` wiring; the component itself is covered.
- `label: nil` inside the keyword opts bypasses the helper's default, the component's fallback still yields "Edit".

Verdict: ship.

PR #793, opened 2026-09-08
Add a rows-per-page selector to the Users admin tabs

timujinne:
## Summary

Users, Sessions, Live Sessions, Roles and the media picker each carried a hand-rolled `join` pagination block with its own page-range helper and a hardcoded page size. They now share `<.pagination_controls>` plus a new sibling `<.page_size_selector>` in `PhoenixKitWeb.Components.Core.Pagination` — a daisyUI select whose choice lives in the URL as `?per_page=` (UrlState allowlist; invalid values fall back to the tab's default), so a size survives reload and is shareable. Picking a size lands on page 1.

- **Roles** is paginated for the first time (25 by default); the summary cards count the whole table (`Roles.count_roles/0`) instead of the rows on screen.
- **Sessions** filters and pages in SQL (`Sessions.list_sessions_paginated/1`) instead of loading every token and slicing in memory. **Live Sessions** keeps the in-memory slice — Presence is an ETS-backed GenServer, there is no table to push LIMIT/OFFSET down to.
- **Opt-in "Auto" size**, piloted on the Users tab: `<.page_size_selector auto_fit>` adds an Auto option backed by the `PageSizeAutoFit` hook in `phoenix_kit.js`, which measures the table's top offset and one row against the viewport and pushes the largest allowed size that fits (`?fit=true`, re-fits on resize). Off by default; `pagination/1`'s API is untouched.
- The selector's form has a stable id so LiveView form recovery works.

Allowlists: `[10, 25, 50, 100]` everywhere, extended with each tab's existing default where it wasn't in the list (Sessions / Live Sessions 20, media picker 30) so the default stays selectable.

## Tests

- Component: selected option, custom options/event, out-of-list value appended, `auto_fit` attrs and hook, missing `table_id` raises.
- `test/js/fit_page_size.test.cjs` for the pure `fitPageSize` helper.
- LiveView tests per tab: size honoured, unlisted value falls back, page resets on change, media picker keeps `return_to`/`mode`, Auto flag lifecycle on Users. The Users-tab and Live Sessions files run `async: false` (the Users view reloads on every global users/stats broadcast; Presence is one named GenServer).

`mix precommit` green (format, credo --strict, dialyzer, JS tests 85/85); `MIX_ENV=test mix test` green — 4764 tests, 0 failures (run in batches against a scratch DB).


ddon:
@timujinne conflicts

timujinne (commented):
## Review — Add a rows-per-page selector to the Users admin tabs

**Scope.** The diff does what the description says: `<.page_size_selector>` plus shared `<.pagination_controls>` on all five tabs, first-time pagination for Roles, SQL pagination for Sessions, in-memory slice kept for Live Sessions, opt-in Auto on Users, and the `pagination_info` translation/`noun_plural` change called out in the CHANGELOG. Nothing unrelated slipped in.

**Verified in the diff**

- **Security / bounded queries.** Every tab allowlists `per_page` in its UrlState spec (`in: [10, 20/25/30, …, 100]`, matching the option list passed to the selector), so `?per_page=7` and `?per_page=1000000` fall back to the default (both tested on Users) and never reach a `LIMIT`. Event handlers funnel values through `push_url_state`, i.e. the same UrlState validation exercised by the URL-level tests.
- **Clamping / round-trip.** Page is clamped into `[1, total_pages]` in SQL and mirrored into assigns (Sessions, Roles); page-past-end is tested on Sessions and Live Sessions. Size change lands on page 1 (`assert_patch` without `page=` on every tab); page links round-trip `per_page` (tested) and `?fit=true`; the media picker preserves `return_to`/`mode` (asserted query equality).
- **Auto lifecycle.** Raise-flag → hook push tagged `auto: "1"` keeps it → hand-picked size drops it; each step asserted including the exact patched URL. The hook's pushed size comes from `data-options`, i.e. the allowlist, so it re-enters validation.
- **Component API.** `noun_plural` default resolved at render time (locale-correct), all three branches translated; out-of-list value appended so the select never shows blank; `auto_fit` without `table_id` raises — all pinned in `pagination_test.exs`.
- **Sessions SQL.** LIKE wildcards escaped (tested with `%`), status buckets tested including no admin leak into "pending"; count and page share one `base` query.
- **Tests.** Component, context and per-tab LiveView tests cover the described matrix; the `async: false` choices are justified in the moduledocs.

**Nits (non-blocking)**

- `list_roles_paginated/1` and `list_sessions_paginated/1` divide by `per_page` with no guard — `per_page: 0` raises. Unreachable from the LiveViews today, but as public context functions a `per_page >= 1` guard/clamp would make the invariant local.
- `expires_at` hardcodes `interval '60 days'`; worth confirming it matches wherever the session TTL is defined (the old formatting path isn't in the diff).
- `per_page` is URL-only (shareable) with no per-user persistence, unlike the grid/list preference on the same Users tab — flagging in case that asymmetry is unintended.

**Not verified**

- Gettext catalogues are omitted (regenerated); completeness of the new msgids ("Rows per page", "Auto", "Showing … %{noun}", "No %{noun}", "sessions") was not checked.
- The compiled `phoenix_kit.js` is omitted, so `PageSizeAutoFit`/`fitPageSize` were reviewed only via their unit tests (e.g. whether the resize re-fit is debounced is unknown).
- Pre-existing `UrlState` internals (`push_url_state`'s page reset, unknown-key preservation) and `pagination_controls` internals are taken on test evidence; likewise that the new Roles `order_by` matches the old unpaginated `list_roles/0` ordering.

VERDICT: LGTM

**After the merge with `main` (`c1ff491`)** — CHANGELOG keeps upstream's released sections above this PR's "Unreleased" entry; catalogues were regenerated against the merged tree with no unrelated `msgstr` changed; `phoenix_kit.js` merged cleanly with upstream's flash-dedup fix. Quality gate on the merged head: `mix compile --warnings-as-errors`, `mix format --check-formatted`, `mix credo --strict` clean; the PR's own tests (31) and `mix test.js` (95) green; full suite 42 doctests, 4825 tests, 7 failures — all 7 reproduce identically on upstream `main` alone (`integrations_key_rotation_test.exs` ×5, one flaky flash assertion in `auth_flows_test.exs`), none in files this PR touches.

One follow-up worth noting: the 7 new msgids ("Rows per page", "Auto", "No %{noun}", the two "Showing…" strings, "results", "sessions") are present but untranslated in all eight locales, so non-English admins will see English for the selector until a translation pass.

PR #796, opened 2026-09-08
V189: remove the dead billing_default_currency setting

timujinne:
Removes a dead setting that answers a live question wrongly.

`phoenix_kit_settings` carries `billing_default_currency`, seeded by V135. Nothing has read it since `phoenix_kit_billing` moved the base currency to the `is_default` row of `phoenix_kit_currencies`, and on a real shop its stored value (`EUR`) contradicts the actual base (`USD`). A setting nobody reads is clutter; a setting nobody reads that states the opposite of the truth is a trap for the next person who greps for it.

This follows V184 exactly, which removed the sibling `shop_currency` key for the same reason: a pure data migration, `DELETE` on up, and on down an `INSERT ... ON CONFLICT ("key") DO NOTHING` restoring V135's original row.

**On `down/1` restoring a value we just called wrong.** Deliberate. A `down` undoes its own `up` and returns the database to its immediately prior state; it is not the place to also repair a V135-era defect that predates this version. The conflict clause means it never clobbers a value an operator recreated by hand after the deletion. Diverging from V184's shape for a structurally identical situation would itself be the trap — a reader has no reason to expect two sibling migrations to behave differently.

No schema object is declared, so the chain hash is restamped rather than the manifest regenerated, matching how V182 and V184 were handled.

## A note on the number

This was built as V188 first. Before finalising, a re-fetch showed upstream had merged its own, unrelated V188 (the user-connections unique indexes) in the meantime. The work was reverted off the shared files, fast-forwarded onto the new tip, and rebuilt as V189 on top of the real V188 — which is untouched here; the diff against `main` in `postgres.ex` and `expected_schema.ex` is additive only. This project has previously had to untangle a fork and upstream claiming the same chain number on a live database by hand, which is why the number was re-read rather than trusted.

## Verified

- Four new tests exercising the real SQL against a seeded row: the row is removed, running against an absent row is a no-op, `down` restores it, and `down` leaves a hand-recreated value alone.
- The chain applies end to end, 135 through 189, and the version marker advances and retreats correctly.
- Full suite: 4864 tests with one failure, a flash double-render race in the auth-flow tests. That failure reproduces identically on a clean `main` worktree with its own database, so it predates and is unrelated to this change.
- Format and static analysis clean.

Independent review published below.

### DIFF
--- a/lib/phoenix_kit/migrations/expected_schema.ex
+++ b/lib/phoenix_kit/migrations/expected_schema.ex
@@ -157,6 +157,20 @@ defmodule PhoenixKit.Migrations.ExpectedSchema do
   # over the 49 shipped files; the real-database integration suite re-ran
   # clean against a DB migrated through V183.
   #
+  # V189 (2026-09-08, per-domain currency Э5) declares NO object here, and
+  # cannot: it is a pure data migration — `DELETE FROM phoenix_kit_settings
+  # WHERE "key" = 'billing_default_currency'` (a dead setting seeded by V135
+  # that nothing reads, and that actively disagreed with the currency
+  # table's `is_default` row it duplicates), plus the version-marker
+  # COMMENT. No table, column, index or constraint is added, dropped or
+  # reshaped — the same V182/V184 class — so `chain_hash` is restamped over
+  # the 55 shipped files rather than the manifest being regenerated. The
+  # chain was re-run end to end into a fresh database (V135→V189) and the
+  # migration's real statements are exercised against a seeded settings row
+  # by test/phoenix_kit/migrations/v189_test.exs, including down/1 never
+  # clobbering a value an operator re-created by hand — the same shape as
+  # V184's `shop_currency` removal.
+  #
   # V188 (2026-09-08) DECLARES three objects here by hand, all new and all
   # indexes: `index:phoenix_kit_user_follows_unique_idx` (follower_uuid,
   # followed_uuid), `index:phoenix_kit_user_blocks_unique_idx` (blocker_uuid,
@@ -296,7 +310,7 @@ defmodule PhoenixKit.Migrations.ExpectedSchema do
   @schema_token "__SCHEMA__"
   @name_marker_exempt "__PK_NAME_EXEMPT__"
   @name_marker_always "__PK_NAME_ALWAYS__"
-  @chain_hash "b31cd24d2f1d905455e5be3db11ce46a09eba374ec99f2fa988a35e7f6493fbc"
+  @chain_hash "2379effc7f93d11d2010c9f4cf0b1774865de371fe91a2674e96d958e4fbfabe"
 
   def objects(prefix) do
     prefix = normalize_prefix!(prefix)
--- a/lib/phoenix_kit/migrations/postgres.ex
+++ b/lib/phoenix_kit/migrations/postgres.ex
@@ -7,7 +7,21 @@ defmodule PhoenixKit.Migrations.Postgres do
 
   ## Migration Versions
 
-  ### V188 - User connections: the three unique indexes the schemas name ⚡ LATEST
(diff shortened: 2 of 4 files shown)


timujinne:
## Independent review

**Numbering, re-checked at review time rather than trusted.** Fetched `main` fresh: its `@current_version` is 188 and no `v189.ex` exists there, so 189 is free. The diff against `main` in `postgres.ex` and `expected_schema.ex` is purely additive — a new section in each, the version constant, the chain hash, and V188 losing its "latest" marker. V188's own body is untouched. No file outside the four in the commit differs.

**The chain hash is genuine.** Ran the restamp script in report-only mode and got exactly the value committed, computed over exactly the 55 migration files on disk. Not stale, not hand-typed.

**`up/1`** deletes only that key and sets the version marker. A delete matching nothing is a no-op by construction and the marker is idempotent, so it is safe to run twice or against a database that never had the row — proven by a test rather than argued.

**`down/1`** restores the row with `ON CONFLICT ("key") DO NOTHING`. I confirmed `phoenix_kit_settings` really carries a unique index on `key` from V135, so that clause is a real guard rather than an inert one, and the test seeds a hand-recreated row with a different value and shows it survives. On the judgment call: restoring the known-wrong value is right. A `down` undoes its own `up`; repairing a V135-era defect is not its job, and diverging from V184's identical precedent would leave a reader with an unexplained asymmetry between sibling migrations.

**Tests** are four real SQL exercises against a seeded row, covering the present-row and absent-row cases on the way up and both the restore and the conflict-safety case on the way down. They would fail if `up` were a no-op or if `down` dropped the conflict guard.

**The full suite's one failure** — a flash double-render race in the auth-flow tests — reproduces identically on a clean `main` worktree with its own isolated database, so it predates this change and is unrelated. I could not reproduce the second, flaky route-dedup failure the author mentioned; it passed repeatedly on both branches, and nothing contradicts their account of it.

Format and static analysis clean on the touched files.

**Verdict:** approve.

timujinne (commented):
## Independent review

**Numbering, re-checked at review time rather than trusted.** Fetched `main` fresh: its `@current_version` is 188 and no `v189.ex` exists there, so 189 is free. The diff against `main` in `postgres.ex` and `expected_schema.ex` is purely additive — a new section in each, the version constant, the chain hash, and V188 losing its "latest" marker. V188's own body is untouched. No file outside the four in the commit differs.

**The chain hash is genuine.** Ran the restamp script in report-only mode and got exactly the value committed, computed over exactly the 55 migration files on disk. Not stale, not hand-typed.

**`up/1`** deletes only that key and sets the version marker. A delete matching nothing is a no-op by construction and the marker is idempotent, so it is safe to run twice or against a database that never had the row — proven by a test rather than argued.

**`down/1`** restores the row with `ON CONFLICT ("key") DO NOTHING`. I confirmed `phoenix_kit_settings` really carries a unique index on `key` from V135, so that clause is a real guard rather than an inert one, and the test seeds a hand-recreated row with a different value and shows it survives. On the judgment call: restoring the known-wrong value is right. A `down` undoes its own `up`; repairing a V135-era defect is not its job, and diverging from V184's identical precedent would leave a reader with an unexplained asymmetry between sibling migrations.

**Tests** are four real SQL exercises against a seeded row, covering the present-row and absent-row cases on the way up and both the restore and the conflict-safety case on the way down. They would fail if `up` were a no-op or if `down` dropped the conflict guard.

**The full suite's one failure** — a flash double-render race in the auth-flow tests — reproduces identically on a clean `main` worktree with its own isolated database, so it predates this change and is unrelated. I could not reproduce the second, flaky route-dedup failure the author mentioned; it passed repeatedly on both branches, and nothing contradicts their account of it.

Format and static analysis clean on the touched files.

**Verdict:** approve.

PR #798, opened 2026-09-09
load_more: forward extra attributes to the button

timujinne:
## What

`<.load_more>` renders its own button and dropped everything the caller passed, so `phx-value-*` never reached the DOM. A page with more than one load-more list therefore had no way to tell its handler which list was clicked, short of minting a separate event name per list.

## How

`attr :rest, :global`, spread on the button. Callers that pass nothing render exactly as before.

Found while moving the Shopify sync page in `phoenix_kit_ecommerce` off a hand-rolled prev/next pager onto this component: that page has six independently paged sections on one screen.

## Checklist

- [x] `mix test test/phoenix_kit_web/components/core/load_more_test.exs` — 6 tests, 0 failures (one new, covering the passthrough)
- [x] `mix format`
- [x] Review published, then un-draft

### DIFF
--- a/lib/phoenix_kit_web/components/core/pagination.ex
+++ b/lib/phoenix_kit_web/components/core/pagination.ex
@@ -359,6 +359,13 @@ defmodule PhoenixKitWeb.Components.Core.Pagination do
   attr :infinite, :boolean, default: false
   attr :cursor, :string, default: ""
 
+  attr :rest, :global,
+    doc: """
+    Forwarded to the button, so a page with more than one load-more list can
+    tell its handler which one was clicked (`phx-value-*`). Without this a
+    caller has to mint a separate event name per list.
+    """
+
   def load_more(assigns) do
     if assigns.infinite and is_nil(assigns.id) do
       raise ArgumentError,
@@ -395,6 +402,7 @@ defmodule PhoenixKitWeb.Components.Core.Pagination do
         class="btn btn-sm"
         phx-click={@on_load_more}
         phx-disable-with={gettext("Loading…")}
+        {@rest}
       >
         {gettext("Load more")}
       </button>
--- a/test/phoenix_kit_web/components/core/load_more_test.exs
+++ b/test/phoenix_kit_web/components/core/load_more_test.exs
@@ -62,6 +62,21 @@ defmodule PhoenixKitWeb.Components.Core.LoadMoreTest do
       assert result =~ ~s(phx-click="my_load_event")
     end
 
+    test "extra attributes ride along to the button" do
+      # A page with more than one load-more list needs to tell its handler
+      # which list was clicked; without a passthrough each list would need
+      # an event name of its own.
+      assigns = %{}
+
+      result =
+        rendered_to_string(~H"""
+        <.load_more loaded={10} total={50} on_load_more="load_more_rows" phx-value-field="title" />
+        """)
+
+      assert result =~ ~s(phx-value-field="title")
+      assert result =~ ~s(phx-click="load_more_rows")
+    end
+
     test "loaded == 0 with total > 0 renders the button" do
       assigns = %{}
 


timujinne (commented):
## Independent review — PASS

Reviewed by someone who did not write it, with the tests run and the forwarding mutation-proved.

**It does what it says.** `attr :rest, :global` is declared after the component's own attrs and `{@rest}` is spread last on the button, matching the pattern this codebase already uses for `button`, `pk_link`, `flash` and others. One thing worth stating plainly, because it is the whole point of the change for the caller waiting on it: `phx-value-*` needs no `:include` here. Phoenix's default global prefixes are `phx-`, `aria-` and `data-`, so `phx-value-field` is forwarded as-is — confirmed by running the new test, which asserts it lands on the rendered button. The dependent `phoenix_kit_ecommerce` change is therefore not blocked by a missing `:include`.

**No caller regresses.** Every `load_more` call site in this repo passes only declared attrs, which the macro excludes from the global capture, so nothing new lands on the button anywhere. `class` and `id` are explicitly declared and go on the outer wrapper, so they cannot leak into `@rest` and collide.

**The test has teeth.** Blanking `{@rest}` reddens exactly the new test and leaves the other five green.

**Runs:** `load_more_test.exs` + both pagination test files — 5 doctests, 26 tests, 0 failures, nothing excluded; three integration pagination files — 16 tests, 0 failures; `compile --warnings-as-errors --force`, `format --check-formatted`, `credo --strict` on the changed file — all clean.

Two notes, neither blocking:

- `load_more`'s `@doc` carries a hand-written "## Attributes" list enumerating every other attr, and `rest` was not added to it. The macro appends its own generated section regardless, so the information is not lost, but a reader skimming the curated list will not learn that extra attributes are now accepted — and the function ends up with two "## Attributes" headings. One line in that list would keep the docstring self-consistent.
- A caller who passed a raw `phx-click` (misuse — the documented API is `on_load_more`) would now have it land in `@rest` and be spread after the component's own, emitting the attribute twice. The browser keeps the first, so the component's own handler wins and nothing breaks, but the markup is invalid. No caller does this today.

timujinne (commented):
## Review (Kimi)

Verified on 791f4f450b inside the phoenix_kit container (MIX_ENV=test, PGHOST=postgres):

- `mix test test/phoenix_kit_web/components/core/load_more_test.exs` — 6 tests, 0 failures (matches the PR checklist)
- `mix test test/phoenix_kit_web/components/` — **634 tests, 0 failures** (regression sweep)
- `mix format --check-formatted` on both files — clean
- `mix credo` on pagination.ex — no issues

### Regression analysis

- **No in-repo callers of `<.load_more>` exist** — the two "load more" UIs in `lib/` (notifications inbox, media browser) are hand-rolled buttons, not this component. Nothing rendered today can change.
- **No-arg callers render byte-identical output**: empty `rest` renders nothing, and the 5 pre-existing tests pin all render branches unchanged.
- **No collision on the likely candidates**: `class` and `id` are declared attrs, consumed before `rest`, still land on the wrapper div.
- **Nobody could rely on the old dropping behavior**: undeclared attrs were a compile error pre-PR.

### Non-blocking notes

1. **`phx-value-*` doesn't reach the `infinite` auto-load path** — the `InfiniteScroll` hook pushes the event from the wrapper with a hard-coded `{}` payload (`phoenix_kit.js:3536`), so a handler keyed on `phx-value-field` gets the value on button clicks but `%{}` on scroll auto-loads. Fine for the motivating use case; maybe worth one doc line on the `rest`/`infinite` attrs.
2. **rest can shadow the button's own bindings** (`phx-click`, `phx-target`, `type`) where it used to be a compile error — standard `rest` trade-off, the attr doc already scopes expectations.

**LGTM** — minimal, correctly scoped, no regression surface found. Safe to un-draft.

PR #805, opened 2026-09-13
Fix raw permission-error atom leaking into role-editor flash

timujinne:
## What

Blocking a role-permissions edit (owner-immutable role, own role, non-owner
touching Admin, not authenticated) put the raw `{:error, reason}` atom
straight into the flash from the Roles LiveView, so the user saw something
like `owner_immutable` instead of a sentence.

## Fix

Added `Permissions.edit_role_permissions_error_message/1`, mapping every
reason `can_edit_role_permissions?/2` can return (`:not_authenticated`,
`:owner_immutable`, `:self_role`, `:admin_owner_only`, plus a generic
fallback) to translated text, and pointed both call sites at it:

- Roles LiveView — previously broken, now shows the message instead of the
  atom.
- PermissionsMatrix LiveView — already had its own private copy of the exact
  same mapping under a different name (`permission_error_message/1`). That
  duplication is exactly how one of the two call sites could drift out of
  sync with the atoms the check actually returns, which is the bug this PR
  fixes. Removed the duplicate and pointed it at the shared function.

All five strings (the four reasons plus the generic fallback) are reused
verbatim from the pre-existing `PermissionsMatrix` mapping, so every locale
under `priv/gettext/*/` already has them — no new translation work needed.

## Tested

- `test/phoenix_kit/users/permissions_test.exs` pins the exact translated
  sentence for each reason `can_edit_role_permissions?/2` can return today,
  plus the generic fallback for an unknown reason.
- `test/integration/users/roles_permissions_editor_flash_test.exs` is a
  LiveView test on the actual bug site: logs in as Admin, clicks the Owner
  role on `/admin/users/roles`, and asserts the flash contains the
  translated sentence and not the raw `owner_immutable` atom. Confirmed it
  catches the regression by reverting the Roles fix line back to
  `put_flash(socket, :error, reason)` and watching it fail, then restored the
  fix.

Full suite on this branch, rebased onto current `main`:

```
42 doctests, 5010 tests, 1 failure
```

The 1 failure is pre-existing and unrelated to this change —
`PhoenixKitWeb.Components.MediaBrowserTest` sidebar-persistence.

`mix format --check-formatted` and `mix credo --strict` are clean.


timujinne (commented):
Reviewed at 05914910. The fix is correct and complete.

**Spec**
- `can_edit_role_permissions?/2` (`lib/phoenix_kit/users/permissions.ex:1371-1400`) can return exactly four reasons today: `:not_authenticated`, `:owner_immutable`, `:self_role` and `:admin_owner_only`. Each maps to its own sentence in `edit_role_permissions_error_message/1`. Anything else gets a translated "Permission denied", so no atom or `inspect` output can reach the flash.
- The strings go through `PhoenixKitWeb.Gettext`, which is already used by other `lib/phoenix_kit` context modules (`auth.ex`, `mailer.ex`, `invitations.ex`), so extraction picks them up. All five msgids already exist in every locale under `priv/gettext`.
- `PermissionsMatrix` produces the same text for the same reasons as before; only the private duplicate is gone.

**Fixed during review (05914910)**
- The unit test asserted only `is_binary` and "not equal to the atom", which would pass even if every reason collapsed into the fallback. It now pins the exact sentence per reason.
- The test comment claimed it would catch a newly added reason atom; with a hardcoded list and a catch-all clause it couldn't. The comment now says only what the test checks.
- Added a LiveView test on the actual bug site (`/admin/users/roles`, Owner role, `show_permissions_editor`). Confirmed it fails against current `main` and passes on this branch.
- Fixed the moduledoc, which still documented `{:error, String.t()}`, and pointed it at the new function.
- The PR body said "four strings"; there are five, counting the fallback.

**Nit (non-blocking)**
- The PR body calls the remaining `MediaBrowserTest` failure a flake. It is deterministic: the "info sidebar collapse … persists across reopen" test (`media_browser_test.exs:586`) selects `[phx-click='toggle_viewer_sidebar']` and gets two elements. It fails the same way on `main`, so it is pre-existing and unrelated to this PR.

**Optional follow-up (out of scope)**
- `save_permissions` in `roles.ex:360-363` still shows its own generic "You cannot edit permissions for this role" instead of the specific reason. Not a leak, but it could use the same helper later for consistent messaging.

PR #806, opened 2026-09-13
Fix module status consumers folding ahead-of-code into up-to-date

timujinne:
## The defect

`Modules.classify/2` folded "database ahead of code" — installed version greater
than what the running code expects — into the same `:up_to_date` result as a
genuinely healthy install. That state happens after a rollback, or when a
dependency gets pinned backwards: the schema was migrated by a later release
than the one now running. Folding it into `:up_to_date` hides exactly that.

## The fix

`classify/2` now returns a distinct `:ahead_of_code` status when
`installed > target`, alongside the existing `:not_installed`, `:needs_update`,
`:up_to_date`, and `:error`.

Every consumer of that status had to be updated in turn, since none of them
previously knew this state existed:

- `mix phoenix_kit.status` — the per-module tree row, the "Modules: N modules,
  ..." summary line (now reports a behind count and an ahead-of-code count
  together when both are present), and the top-level "Next" action all render
  it distinctly (red, "ahead of code — rollback?") instead of silently
  counting it toward "all up to date".
- `PhoenixKit.Install.StatusReport.next_action/3` — a new
  `{:modules_ahead_of_code, names}` action, so a module ahead of code no
  longer reports `Ready`. A pending module still takes priority — `mix
  phoenix_kit.update` is offered and the ahead-of-code module is named
  alongside it, rather than one hiding the other.
- `mix phoenix_kit.update` — ahead-of-code modules are now reported on every
  run (not only when nothing else is pending), the post-migration
  re-verification step, and the version-tree renderer each got a dedicated
  branch instead of falling through to "up to date" or a "migration did not
  run" message that would have been misleading for this case.
- `mix phoenix_kit.doctor`'s module schema check had the same gap — an
  ahead-of-code module was neither pending nor failed, so it fell through to
  the same `:pass` a healthy install gets. It now warns, and still fails (with
  the ahead-of-code module named alongside) when another module is genuinely
  behind.

## Behavior change

`mix phoenix_kit.status --exit-code` now exits 1 when a module is ahead of
code, same as any other non-Ready state. A deploy gate run right after a
rollback or a backwards-pinned dependency will now fail where it previously
passed — that's the fix, not a side effect, but worth calling out since it's
an exit-code change.

## Testing

Rebased cleanly onto current `main`.

- `mix credo --strict` — no issues (963 files).
- `mix test` against a local Postgres with an isolated partition database
  (integration suite included, nothing excluded): `42 doctests, 5021 tests,
  1 failure`. The one failure is `PhoenixKitWeb.Components.MediaBrowserTest`
  "info sidebar collapse", which fails the same way on `main` (19876227) and
  is unrelated to this change. Every test touching `Modules`, `StatusReport`,
  the status task, or `phoenix_kit.doctor` passes, including the new cases
  for `ahead_of_code/1`, `next_action/3` (including mixed pending +
  ahead-of-code), `describe/1`, `exit_code/2`, and
  `module_schema_versions_result/1`.


timujinne (commented):
Reviewed the change and every consumer of `Modules.classify/2` at 578cfdd9.

**Scope.** Adding `:ahead_of_code` without teaching the consumers would just move the false all-clear somewhere else, so extending the fix to them is warranted. Checked every place that reads module status: the status tree, summary and Next line, `--verbose` diagnostics (catch-all, fine), `StatusReport.next_action/3`, the update task's report, verify and tree steps, and `phoenix_kit.doctor`. None of them can hit a `CaseClauseError` or `FunctionClauseError` on the new atom. Install and igniter only read the core `{:up_to_date, v}` tuple, the admin modules page does not use `Migrations.Modules`, and no sibling `phoenix_kit_*` package matches on module status. Behaviour for pending and unreadable modules is unchanged.

**Found and fixed during review**

- MAJOR (fixed in 4505426d): `mix phoenix_kit.doctor`'s module schema check was missed. An ahead-of-code module is neither pending nor failed, so doctor still returned `:pass, "all at the version their code expects"`. Now it warns, like the core "DB version > code version" check. When another module is behind, the `:fail` message also names the ahead module. The decision is extracted into a pure function with six tests.
- MINOR (fixed in 4505426d): priority in `next_action/3`. The first version checked ahead before pending, so with one module behind and another ahead, Next dropped both the behind module and the update command; the summary line did the same. Now pending wins and the ahead module is added to the update reasons, matching the core-behind path. The summary shows "N behind, M ahead of code". Test added for the mixed case.
- MINOR (fixed in 4505426d): `mix phoenix_kit.update` only reported ahead-of-code modules when nothing else was pending. They are now reported on every run.
- MINOR (fixed in the description): the Testing section said there was no local Postgres; the partition database just had not been created, so the integration half had been excluded. The description now states the `--exit-code` behaviour change: a deploy gate run right after a rollback now exits 1.
- MINOR (fixed in the description): the Testing section then quoted database-less numbers; it now has the full-suite counts.
- NIT (fixed in 4505426d): two tests repeated existing assertions, one with wrong mutation reasoning in its comment. Both removed.
- NIT (fixed in 4505426d): a status-task comment ("neither pending nor failed") no longer described its branch. Resolved by the reorder.
- NIT (fixed in 578cfdd9): a comment in `phoenix_kit.update.ex` described a different function's clause. Moved.

**Accepted as is**

- MINOR: the update task's new private output branches (verify, report, tree) have no tests. There is no capture harness for that task today, and the decision logic they rely on is tested through `StatusReport` and doctor.
- NIT: the `exit_code/2` "module ahead of code → 1" test passes through the catch-all clause. Fine as a regression pin.

Tests: `mix test` with the integration suite included gives 42 doctests, 5021 tests, 1 failure, and that one (`MediaBrowserTest`, info sidebar collapse) fails the same way on `main`. Format and credo are clean.

PR #807, opened 2026-09-13
Extend restricted-setting encryption to Apple/billing secrets, add perimeter guard

timujinne:
## What

Follow-up to #759 and #766. Those covered the OAuth login secrets and AWS
keys; this extends the same `@restricted_setting_keys` encryption-at-rest to
two more groups that write through the same `PhoenixKit.Settings.Setting`
changeset:

- `oauth_apple_private_key` — no reader/writer exists yet (Apple Sign-In isn't
  implemented), but a key dropped in ahead of the feature should never be
  written in the clear.
- The `phoenix_kit_billing` provider secrets: `billing_stripe_secret_key`,
  `billing_stripe_webhook_secret`, `billing_stripe_api_key` (a legacy alias
  the Stripe provider still falls back to), `billing_paypal_client_secret`,
  `billing_paypal_webhook_secret`, `billing_razorpay_key_secret`,
  `billing_razorpay_webhook_secret`, `billing_everypay_api_secret`.
  `phoenix_kit_billing` is a separate hex package but writes these through
  `PhoenixKit.Settings.update_setting/2` directly, so core's restricted list
  is the only place that can protect them — core has no other notion of that
  package's keys. Their public counterparts (`billing_stripe_publishable_key`,
  `billing_paypal_client_id`, `billing_razorpay_key_id`, the Everypay
  identifiers) are excluded on purpose, same reasoning as the OAuth
  client_id/app_id exclusion already in place: public by the provider's own
  design.

`billing_paypal_webhook_secret` needed a second pass after the rest: the
webhook controller builds that key by string interpolation
(`"billing_#{provider}_webhook_secret"`), never as a literal, so a plain grep
for the string missed it initially.

## Migration note

Same approach as #759: new values for these keys are stored encrypted;
**existing plaintext values are left untouched** and are only encrypted the
next time their value actually *changes* (`maybe_encrypt_restricted_value/1`
diffs via `get_change/2`, so resubmitting the same plaintext — e.g. a
provider settings form saved without editing the secret field — leaves it
as plaintext; nothing to encrypt if the value didn't change). Migrating
what's already in a database is a separate, deliberate operator step, not
part of this change.

## Perimeter guard

The `get_defaults/0`-based partition invariant in `settings_test.exs` (every
key classified exactly once) only walks `Settings.get_defaults/0`'s own keys
— a secret-shaped key that's never added there is invisible to it, not
merely misclassified. That's exactly how the billing secrets went
unrestricted for as long as they did: written by a separate package, with no
`get_defaults/0` entry of their own to be checked against.

`test/support/secret_key_perimeter.ex` adds a static scan of core's own
`lib/` tree and migration seeds for setting-key literals — passed either as
a literal string or through a module attribute (`@key_name "..."` resolved
back to its value; this is how `website_access_password`/
`website_access_link_token` are actually referenced) — that *look*
secret-shaped (contain `secret`/`password`/`private_key`/`api_key`/`token`/
`credential`/`passphrase`, or end in `_key` without being a public
`_client_id`/`_app_id`), and asserts every one found is on
`@restricted_setting_keys`. A fixture-based test proves the scan mechanism
itself isn't vacuous, for both the literal and the attribute-resolution
path.

This only covers what lives in core's own repository — it cannot see into
`phoenix_kit_billing`'s source (a different git repo), so it can't have
caught the original gap by itself; a module package with its own
secret-shaped keys needs to assert them against
`PhoenixKit.Settings.restricted_setting_keys/0` in its own suite, the same
way this PR does for core's.

## How tested

Isolated DB (`MIX_TEST_PARTITION`):

- `mix credo --strict` — `12917 mods/funs, found no issues.`
- `mix test` (full suite): `42 doctests, 5022 tests, 3 failures`, none in
  `settings_test.exs` or the new perimeter test file:
  `MediaBrowserTest` "info sidebar collapse" fails identically on
  `upstream/main` at `19876227` (a selector matching two
  `toggle_viewer_sidebar` buttons) — pre-existing, unrelated to this
  change. The other two (`RegistrationTest`, `CustomFieldsDefinitionsTest`)
  are DB lock/connection timeouts under a shared Postgres; re-run alone
  (`PGPOOL=8 mix test test/integration/users/registration_test.exs
  test/integration/users/custom_fields_definitions_test.exs --max-cases 4`)
  they pass: `37 tests, 0 failures`.
- Focused run of the affected files alone: `33 tests, 0 failures`.

## Not covered

- Existing stored values for these keys are not migrated; nothing on any
  installation is encrypted by this change alone.
- `oauth_apple_private_key` has no reader/writer yet — this only makes sure
  it won't be stored in the clear whenever Apple Sign-In ships.


timujinne (commented):
Reviewed at 48e33266 (commits f94ab31b, ed1cccbd, f05bddaf, 48e33266 on main 19876227).

The core change is correct. Checked the nine keys against phoenix_kit_billing 0.15.1:
- The 8 billing keys are exactly the provider secrets it reads and writes: provider_settings.ex, stripe.ex (including the legacy `billing_stripe_api_key` fallback), and `WebhookController.get_webhook_secret/1`, which builds `billing_#{provider}_webhook_secret` for stripe/paypal/razorpay; everypay never calls handle_webhook.
- `oauth_apple_private_key` covers Apple Sign-In ahead of the feature.
- Nothing is missing and nothing is invented. The public counterparts (publishable key, client/key ids, webhook id, EveryPay username/account name) are correctly left unrestricted.
- The get_defaults/0 additions are what the partition invariant requires and do not change what the settings LiveViews expose.

Existing plaintext values work as described: `decrypt_if_restricted/2` returns a value without `enc:v1:` unchanged, and `Setting.maybe_encrypt_restricted_value/1` encrypts on write only when the value actually changes (`get_change/2`). Re-saving the same plaintext leaves it plaintext; the description now says this precisely.

**Findings from the first round, all addressed**

MAJOR
1. The webhook-provider perimeter test read phoenix_kit_billing's source from a hardcoded sibling-checkout path and skipped otherwise, so it would always be skipped on a plain clone or CI and coupled core's suite to billing internals. Removed together with `scan_webhook_provider_atoms/1` (f05bddaf). A module package should assert its own secret keys against `PhoenixKit.Settings.restricted_setting_keys/0` in its own suite.
2. The literal scan only matched get_setting/update_setting/update_setting_with_module, about half of core's literal-key calls, and could not see keys referenced through module attributes, so `website_access_password` and `website_access_link_token` (Gate `@password_key`/`@link_key`) could be dropped from the restricted list unnoticed. It now matches any `Settings.<fun>(` call with a literal or same-file `@attr` argument, with a floor assertion on `website_access_password` and a fixture test for attribute resolution (f05bddaf).
3. `secret_shaped?/1` had no `token` rule, its billing-only suffix exclusions were dead in a core-only scan, and `_key_id` wrongly exempted `aws_access_key_id`, which core restricts. Exclusions are now limited to `_client_id`/`_app_id`; `token`/`credential`/`passphrase` were added (f05bddaf).

MINOR
4. Change-history narration in the `@restricted_setting_keys` comment and elsewhere trimmed to policy (f05bddaf).
5. Process wording in code and tests removed (f05bddaf, 48e33266).
6. The migration note said "re-encrypted the next time it is written"; it now says "when its value changes". The f94ab31b commit message keeps the older wording.
7. Nine copy-pasted round-trip tests collapsed into a module-level `for` that still yields one named test per key (f05bddaf).
8. A describe block that only demonstrated the old invariant's non-behaviour was dropped (f05bddaf).
9. The "How tested" section was stale and labelled a deterministic failure as flaky. Corrected.

NIT
10. The scan regex now rejects `Foo.Settings.` calls (f05bddaf).
11. Cosmetic, left as is: settings_test.exs, the comment above the per-key round-trip loop still says "once per key this fix added".

Known limitation, documented in the scanner moduledoc: the guard only sees core's own repository. Interpolated keys and list-taking readers (`get_settings_direct/1`, `get_settings_cached/2`) are out of reach for a static scan.

Tests: settings_test.exs + settings_secret_perimeter_test.exs give 33 tests, 0 failures; the full suite at ed1cccbd with the database reachable gave 42 doctests, 5024 tests, 7 failures, of which 6 were lock/pool contention that pass when re-run alone and 1 is the MediaBrowserTest "info sidebar collapse" failure that reproduces on main. The later commits touch only tests, test support and one comment.

PR #808, opened 2026-09-13
Fix OAuth Test Credentials button to actually test credentials

timujinne:
## What was wrong

The "Test Credentials" button on `/admin/settings/authorization` compared each OAuth field to `""` and nothing else. A 9-character Google client secret — or a field containing only whitespace — read as "properly formatted" while Google itself answered the exact same values with a 401 `invalid_client`. The button told operators their broken credentials were fine.

## What it does now

**Google** gets a real live check: `OAuthConfig.test_connection/2` POSTs the current client ID/secret to `oauth2.googleapis.com/token` with a deliberately invalid authorization code and no `redirect_uri`. This used to send an RFC 2606 `.invalid` placeholder, which fails Google's ["Host TLDs must belong to the public suffix list"](https://developers.google.com/identity/protocols/oauth2/web-server) redirect-URI rule and came back `invalid_request` regardless of whether the credentials were right or wrong, making the check permanently inconclusive — this is specifically a TLD-validity rejection, not a blanket "any placeholder"/"unregistered host" one: verified live with fabricated credentials that a placeholder on a real public-suffix host (`example.com`, `localhost`) gets the same `invalid_client` verdict as sending no `redirect_uri` at all. Reads Google's own verdict (same guide + RFC 6749 §5.2):

- `invalid_client` → the credentials are wrong (Client ID/Secret don't match a registered app)
- `invalid_grant` → the credentials are right (Google authenticated the client fine; only the fake code was rejected, as expected)
- anything else — a different error code, unexpected status, a timeout, or a crash/exit inside the check — is reported as **inconclusive**, a distinct `{:inconclusive, message}` result shown as a warning flash rather than an error, never folded into either a pass or a fail (and never leaking the raw exception text into the flash or the log — this is the one code path holding `client_secret`). The check runs through `PhoenixKit.Integrations.Probe` (the same isolated-process/deadline wrapper the Integrations "Test Connection" checks use, whose type was widened to describe this third outcome). Req's own connect/receive timeouts (3s/4s, with a 1s margin under Probe's 8s deadline for the sum) are deliberately shorter than that deadline, so an ordinary network timeout resolves as `:inconclusive` through the normal response path; a `raise` or `exit` inside the check itself is caught locally and also reported as `:inconclusive`, rather than falling through to Probe's own generic, untagged crash/deadline fallback. Only an untrappable `:kill` — Probe's own deadline firing, or a third party killing the check process — remains outside this and surfaces as a plain error.

**What was and wasn't verified against real Google:** with fabricated client_id/secret (no real Google OAuth app used), dropping `redirect_uri` turns the response from `invalid_request` into `invalid_client`, and a placeholder on a real public-suffix host behaves the same as no `redirect_uri` — both confirmed live. The `invalid_grant` leg (a genuinely correct credential pair) was **not** exercised against a real registered app; it follows from Google's documented definitions and the fact that a wrong secret for a real, public client_id already comes back `invalid_client` (Google authenticates the client before it looks at the code), so a wrong secret can't itself surface as `invalid_grant`. If Google ever answers a correct pair with something other than `invalid_grant`, the response falls into the inconclusive branch rather than being misread either way. A `Req.Test`-stubbed test suite (`test/phoenix_kit/users/oauth_config_google_live_check_test.exs`) pins the outgoing request (no `redirect_uri`, correct `client_id`/`client_secret`/`grant_type`) and exercises all outcomes end-to-end through stubbed HTTP, without a real network call: `invalid_client`, `invalid_grant`, `invalid_request`, a transport error, and a raise inside the stubbed request (confirming the exception's own message never reaches the flash).

**GitHub and Facebook** keep the previous format-only check for now (no live network round trip against their token endpoints — that would need its own from-scratch verification of a similar three-outcome property, which this change doesn't attempt) — but the format check itself is now actually enforced instead of an empty-string comparison.

**A short or blank-looking secret is also rejected on save**, not just by the button: `PhoenixKit.Users.OAuthConfig.validate_secret_format/2` rejects a value that's only whitespace, or implausibly short (under 16 characters — real secrets from all three providers are at least 24), before the settings LiveView persists it. This is deliberately wired into the Authorization LiveView, not into `PhoenixKit.Settings.Setting`'s generic changeset, since a Google-secret-shaped format check has no business in a schema shared by every setting in the app. A secret already saved before this fix does not retroactively block saving an unrelated field on the same form. A secret or client ID with leading/trailing whitespace (a copy-paste artifact) is trimmed before it is persisted or tested.

**Test-result messages are now translated.** The success message was never wrapped in `gettext` — a translated button produced an English response, easy to mistake for "this works" at a glance. All three providers' messages, on all three outcomes, are now gettext-wrapped, along with the adjacent "Reload Config" flash and the "Test Credentials" button name as it appears in the on-page setup instructions.

## How tested

`PGHOST`/`PGDATABASE` pointed at a scratch database (`MIX_TEST_PARTITION=_i135`, `PGPOOL=8 mix test --max-cases 4` to stay under the shared Postgres's connection limit):

- `mix format --check-formatted` — clean
- `mix credo --strict` — 966 files, no issues
- `mix dialyzer` — clean
- `mix test` — **42 doctests, 5038 tests, 1 failure** on this head, neither failure related to this change: `PhoenixKitWeb.Components.MediaBrowserTest` "info sidebar collapse" (media browser UI) also fails on upstream/main at the same commit this branch is based on, so it's pre-existing; a second, intermittent failure was also seen on an earlier commit of this branch in `PhoenixKit.Install.CommonUnreachableTest` "an unqueryable database reports `{:unreachable, _}`" — a test-ordering flake (a DB connection owned by an already-finished test), passes 3/3 run on its own, and this PR touches neither `install/` nor `migrations`.
- This PR's own test files pass cleanly on their own: `test/phoenix_kit/users/oauth_config_test.exs`, `test/phoenix_kit/users/oauth_config_google_live_check_test.exs` (the `Req.Test`-stubbed end-to-end coverage above), `test/integration/phoenix_kit_web/live/settings/authorization_oauth_secret_validation_test.exs`.


timujinne (commented):
Reviewed at 15086261. Two major issues and a set of smaller ones were found; all are fixed in the follow-up commits, listed with the commit that fixed each. One behaviour is still unconfirmed against a real Google OAuth app and is listed as an open item at the end.

**Major**

1. The Google live check could never reach a verdict (fixed in ba73c2f5). The check sent `redirect_uri=https://phoenixkit.invalid/oauth/callback`. Against `oauth2.googleapis.com/token` that answered HTTP 400 `invalid_request` for every credential set tried, including a client ID that does not exist, so every click came back inconclusive and a bad secret was never reported as invalid. The cause is Google's redirect-URI rule that host TLDs must belong to the public suffix list, which `.invalid` is not. With no `redirect_uri` a wrong secret correctly gets 401 `invalid_client`. The parameter is now omitted; the explanation in code, CHANGELOG and tests was narrowed to exactly that cause (46db74dd, 59561d65), and a stubbed test pins the outgoing request without `redirect_uri` (46db74dd).
2. The CHANGELOG entry sat in the `## 2.23.0` section, already tagged (fixed in ba73c2f5). It now lives under a new `## Unreleased` section; no version bump.

**Minor**

- "Rejected" and "inconclusive" both returned `{:error, msg}` and showed the same red flash. `test_connection/2` now returns `{:inconclusive, msg}`, shown as a warning flash (ba73c2f5). `Integrations.Probe`'s typespec was widened to match, no behaviour change for other callers.
- A timeout or crash still reached Probe's generic `{:error, ...}` fallback; Req's receive timeout equalled the Probe deadline. Req now has its own 3 s connect / 4 s receive bounds under Probe's 8 s, and raises and exits inside the check are caught and reported as inconclusive (59561d65, 15086261).
- The first crash handler put the exception message into the flash and log, on the one code path holding `client_secret`. It now returns a fixed message and logs only the exception module; a test raises with a fake secret fragment and asserts it never reaches the message (15086261).
- Secrets with surrounding whitespace passed the format check but were saved padded. Secrets (ba73c2f5) and client IDs (59561d65) are now trimmed before saving and testing, each with an integration test.
- The redirect-URI explanation overstated the failure as "any non-registered host"; only the `.invalid` TLD is rejected. Reworded everywhere (59561d65).
- "`invalid_grant` means the credentials are right" was presented as verified. It follows from Google's documented error definitions and RFC 6749 §5.2 but was not exercised with a real app; the comment, CHANGELOG and description now say so (46db74dd, 59561d65).
- The integration test never clicked Test Credentials and one assertion passed trivially. A button test and a real before/after assertion were added (ba73c2f5); `oauth_config_google_live_check_test.exs` covers all outcomes through a stubbed HTTP layer (46db74dd, 59561d65, 15086261).
- The description's test numbers did not match the actual run; fixed with measured counts.

**Nits (all fixed)**

- A transport failure was logged twice (ba73c2f5).
- No busy state on the Test Credentials buttons during the up-to-8 s check; `phx-disable-with` added on all three (ba73c2f5).
- The test-only `opts` argument could override `form:`/`retry:`/timeouts; only `:plug` is honoured now and it is documented as a test hook (40ebd7fe).
- A duplicated sentence in the `google_live_check` comment (46db74dd).

**Checked and fine**

- No credential appears in flash messages or logs; all user- and log-facing strings are fixed gettext messages.
- One request per click with `retry: false` and hard time bounds.
- The 16-character minimum cannot reject a real secret: Google's are 24 characters (35 with `GOCSPX-`), GitHub OAuth App secrets 40 hex, Facebook App Secrets 32 hex.
- Saving with an empty secret field is not blocked; an untouched field keeps the stored value, as before.
- All three providers' Test Credentials buttons are intact in the restructured template. No `.po` files touched, no version bump.

**Tests**

- 59561d65, full suite: 42 doctests, 5037 tests, 2 failures, neither related: `MediaBrowserTest` "info sidebar collapse" also fails on main 19876227, and `CommonUnreachableTest` failed once through test ordering and passes 3/3 alone.
- 15086261: the three OAuth test files plus the Probe and Req.Test-using tests give 64 tests, 0 failures on two consecutive runs; `--warnings-as-errors` compile clean; full suite 42 doctests, 5038 tests, 1 failure (the same MediaBrowserTest).

**Open item**

- The "correct credentials → `invalid_grant`" path has not been exercised with a real registered Google OAuth app. Google authenticates the client before looking at the code, so a wrong secret cannot come back as `invalid_grant` and read as valid; if Google ever answers a correct pair with anything else, the button reports inconclusive, never valid. Worth one manual click with real credentials before relying on the green result.

**Optional follow-up (not blocking)**

- The `catch kind, reason` clause logs `inspect(reason)`. With Req's default HTTP/1 pool an exit reason names only the pool, so nothing leaks today; logging just `kind` would keep that true under HTTP/2.
- Clearing a stored OAuth secret from this page is still not possible because a blank field keeps the stored value. Predates this PR.

PR #809, opened 2026-09-13
Fix Activity.list/1 and recent/1 losing order on tied timestamps

timujinne:
## Problem

`Activity.list/1` and `Activity.recent/1` both order entries by `inserted_at`
alone, which carries no uniqueness constraint, so two entries can tie on it —
rows written before V185 (`inserted_at` went from second to microsecond
precision) are all whole-second, which is the common case on any existing
install, and even microsecond-precision writes can land on the identical
instant. When they do, Postgres is free to return them in whichever order its
query plan happens to land on — not necessarily the order they were actually
created in, and that order isn't even guaranteed to be stable across two runs
of the same query.

## Fix

Add the entry's `uuid` primary key as a second, descending sort key on both
queries. It's unique and `NOT NULL` by schema, so `(inserted_at, uuid)` is a
full order with no ties left to arbitrate — deterministic, and chronological
to the millisecond for app-generated entries (UUIDv7's leading 48 bits are a
millisecond timestamp). No migration needed.

## Testing

- `test/phoenix_kit/activity/order_sql_proof_test.exs` — asserts the compiled
  `ORDER BY` SQL of both functions includes the `uuid DESC` tiebreak. Runs
  without a database.
- `test/integration/activity_order_determinism_test.exs` — inserts three
  entries with an identical `inserted_at` (forced via `insert_all/3`,
  bypassing the autogenerated timestamp), out of order (middle, latest,
  earliest), and asserts `recent/1` and `list/1` both return
  `[latest, middle, earliest]`. Three rows, not two: with only two, either
  insertion order or its reverse always happens to match the expected
  result, so the test would pass without the fix.

Ran against a local Postgres:

```
42 doctests, 5011 tests, 1 failure
```

The 1 failure (`PhoenixKitWeb.Components.MediaBrowserTest` sidebar-collapse
test) reproduces on unmodified `main` and is unrelated to this change.
`mix format --check-formatted` and `mix credo --strict` both pass clean.


timujinne (commented):
Reviewed at eb053976.

**The fix.** `list/1` and `recent/1` now order by `inserted_at DESC, uuid DESC`. `uuid` is the unique, NOT NULL primary key, so the order is total. This is the same tiebreak `Settings.History` already uses (history.ex:138/173/186). `list/1` paginates with offset/limit, not a cursor, so nothing else needs the second key. No new index is needed either: on 300k rows, EXPLAIN shows an incremental sort on top of a backward scan of `phoenix_kit_activities_inserted_at_index`; only groups of equal timestamps get sorted, and the first page stays at 4 buffers.

**What changed during review (all in the final commit):**

1. **The regression test did not reliably catch the bug (major).** The first version inserted two tied rows in chronological order. Without the tiebreak, Postgres returns tied rows in insertion order (seq scan + sort) or in reverse (backward index scan), whichever plan it picks. Run against unfixed main, that version's `recent/1` test passed and only `list/1` failed; with two rows one of those orders always equals the expected result. The test now inserts three tied rows as middle, latest, earliest, so neither insertion order nor its reverse matches the expected `[latest, middle, earliest]`. Verified: both tests fail on unfixed main and pass with the fix, 3 seeds each.
2. **Premise corrected (minor).** The description, moduledocs and commit message said `inserted_at` has second resolution. It has had microsecond precision since V185. Ties still happen on rows written before V185 (most rows on existing installs) and on writes with the identical microsecond. The wording now says so.
3. **Chronology claim narrowed (minor).** UUIDv7 is a 48-bit millisecond timestamp plus random bits, generated app-side. The tiebreak is deterministic, and chronological only to the millisecond; within one millisecond the order is fixed but arbitrary. The description now says exactly that.
4. **SQL guard made order-sensitive (nit).** `order_sql_proof_test.exs` checked that both keys were present, which would also pass `desc: uuid, desc: inserted_at`. It now matches `inserted_at DESC, … uuid DESC` as one pattern.

**Testing.** Both new test files pass. `mix format --check-formatted` and `mix credo --strict` are clean. The one full-suite failure, MediaBrowserTest "info sidebar collapse", reproduces on unmodified main and is unrelated to this change.

PR #810, opened 2026-09-13
Fix settings cache miss-fill and restricted-key nil caching

timujinne:
## Summary

Three related defects in the settings cache's miss-fill and restricted-key
handling, all in `lib/phoenix_kit/settings/settings.ex`:

1. **A cache miss silently returned `nil` instead of the database value,
   once a cache process is actually running and holds some keys but not
   others.** `get_settings_cached/2` and `get_json_settings_cached/2` both
   pass `%{}` as the `defaults` argument to `Cache.get_multiple/3`. That
   function does not omit keys it doesn't hold — it always writes every
   requested key into its result, substituting `Map.get(defaults, key)`
   (i.e. `nil`, from the `%{}`) on a miss. That made a miss indistinguishable
   from "cached, and the value happens to be nil", so the `Map.has_key?/2`
   check both functions used to detect a miss always saw every key as
   present. `fill_missing_settings/1` / `fill_missing_json_settings/1` — the
   only places that actually query the database — were therefore never
   called when the cache was answering for real. With the settings cache TTL
   now live, every expiry silently degraded reads to `nil` instead of
   re-querying.

   The existing `settings_cache_missfill_test.exs` didn't catch this because
   it never runs a live cache process against the database: with no cache
   process, `Cache.get_multiple/3` itself catches `:noproc` internally and
   returns whatever `defaults` it was given verbatim, which — before this
   fix — was the hardcoded `%{}`, making the result empty for every key
   regardless of the sentinel logic. That makes every requested key "missing"
   either way, so the miss-fill ran and answered correctly on both the old
   and the new code, masking the bug. The bug only shows up once a cache
   process is genuinely running and holds some keys while missing others.
   Added tests that start a real `PhoenixKit.Cache` GenServer against the
   database, covering both an explicit invalidation and a real TTL expiry.

   Fixed by tagging every requested key with a private sentinel as the
   cache-level default instead of `%{}`, and detecting a miss by matching
   that sentinel rather than by key presence — the same pattern the
   single-key `get_setting_cached/2` already uses.

2. **`false`/`nil` JSON values read back as the caller's default.** Both
   miss-fill writers used `||` to fall back to the "not found" sentinel
   whenever the looked-up value was falsy. `||` can't tell "key absent"
   apart from "key present with a falsy value", so a key the database
   genuinely held with a `nil` (or `false`, wherever the JSON value type
   allows it) value got written to, and read back from, the cache as "not
   found" and silently replaced by the caller's own default. Switched both
   spots to `Map.fetch`/`case`.

3. **An undecryptable restricted key got cached in a way that blocked
   recovery, on the batch-fill path.** `fill_missing_settings/1` (the
   read-through path `get_settings_cached/2` calls on a cache miss) cached a
   restricted key's decrypt failure as a bare `nil` straight from
   `decrypt_if_restricted/2`. Once cached, the failure read back exactly
   like a setting that is genuinely `nil` for the rest of the entry's life
   instead of retrying the decrypt once it expired or was invalidated. This
   is a different code path from the one #ee48b351 ("Do not cache a
   restricted setting the boot warmer cannot decrypt") already fixed — that
   one covers the single-key path and the boot warmer's own map, but not
   `fill_missing_settings/1`, which a key hits on every batch miss
   regardless of how it got there.

   ee48b351's own rule is: answer the failure but don't cache it, so the
   next read retries. Routed through that same "leave it out of the cache
   write" rule instead of the "not found" sentinel — caching the sentinel
   would have been a *different, wrong* rule (`get_setting_cached/2` treats
   it as "row does not exist" and stops retrying for as long as the entry
   stays cached), and since the single-key and batch paths share one cache
   entry per key, would have poisoned the single-key reader too, for a real,
   currently-configured secret. Verified with a test that makes a
   genuinely-stored secret temporarily undecryptable (swapping the
   resolvable key, not corrupting the ciphertext) and then restores it, so
   the assertion covers the recovery, not just the failure.

## Known pre-existing limitation (not introduced or fixed here)

The string and JSON read paths share one cache entry per key but cache
different shapes for it. A key read through both APIs after an
invalidation/expiry races whichever one fills the cache first. No core
caller mixes both APIs for the same key today; documented in the code as a
follow-up rather than fixed in this PR.

## Testing

`mix format --check-formatted`, `mix credo --strict`, and
`mix compile --warnings-as-errors` all pass clean.

`mix test` against an isolated database: **42 doctests, 5018 tests, 1
failure** — the one failure (`PhoenixKitWeb.Components.MediaBrowserTest`, a
duplicate-selector assertion in the media browser sidebar toggle test, and
separately observed as a transient Postgres deadlock in an unrelated role-
reorder integration test on other runs) is pre-existing flakiness unrelated
to this change. All existing settings/cache tests, including the ee48b351
regression tests, still pass.

Every new/changed assertion was verified to fail against the code it's
guarding: the live-cache miss-fill tests fail on unmodified upstream/main,
and the restricted-key recovery tests fail against the cache-write-a-
sentinel approach this PR does NOT take.

## Test plan

- [x] `mix credo --strict`
- [x] `mix test` (5018 tests, 1 pre-existing unrelated failure)
- [x] Mutation-checked: new tests fail on unmodified upstream/main and on
      the earlier (sentinel-based) version of the restricted-key fix


timujinne (commented):
Reviewed at 09d1ef8b against main 19876227 with a live settings cache process and a real test database.

**Proof the tests bite.** The PR's five test files copied onto unmodified main: 16 tests, 7 failures. The three live-cache miss-fill tests (string invalidation, real TTL expiry, JSON), the two "caller default on a genuine miss" unit tests and the two restricted-key recovery tests all fail there, each returning nil where the stored value or the caller's default is expected. The rewritten restricted-key test also fails against the first-round version that cached the sentinel and against the intermediate commit that cached a bare nil, so it isolates the second commit. Putting back either `||` in the JSON path fails exactly its matching test.

**Round 1 (fc8a5b1f), changes requested**
1. MAJOR: `fill_missing_settings/1` cached a restricted key's decrypt failure as the not-found sentinel. Reproduced: store a secret, make it temporarily undecryptable, batch-read it, restore the key. Both `get_settings_cached/2` and `get_setting_cached/2` kept returning the default for the life of the cache entry, while the single-key path on its own retried and got the real secret. This contradicted the rule from ee48b351 (answer a decrypt failure, never cache it) and hit the OAuth credential helpers.
2. MAJOR: no test ran the miss-fill with a live cache process and a real database. The existing integration test only exercised the no-cache-process path, which passes on main unchanged.
3. MINOR: the description said the old test passed because the `rescue` fell through to a direct query. In fact `Cache.get_multiple/3` catches `:noproc` itself and returns the defaults, so the miss-fill did run; it just never covered a running cache that misses some keys.
4. MINOR: the string and JSON readers share one cache entry per key but store different shapes, so a JSON batch read of a string setting caches nil for later string reads, and the reverse. The single-key paths already behaved this way.
5. MINOR: stale moduledoc in `settings_json_nil_false_test.exs`.
6. NITs: unreachable `:error` clauses in both readers; an internal tracking id in a test moduledoc; two unit tests that also pass on main; a redundant `Map.new` in `warm_cache_data/0`.

**Round 2 (09d1ef8b)**
1. Fixed. Decrypt failures are now left out of the cache write in both `fill_missing_settings/1` and `warm_cache_data/0`. The rewritten test makes a stored secret temporarily undecryptable and checks that both readers return the real value once the key is back.
2. Fixed. New live-cache tests cover explicit invalidation, a real TTL expiry and the JSON path; all three fail on main.
3. Fixed in the description and the commit message.
4. Documented as a known pre-existing limitation next to the sentinel and in the description. Acceptable here, since no core caller mixes both APIs for one key.
5. Fixed.
6. Resolved: the unreachable clauses stay with a "defensive only" comment, the tracking id is dropped, the two non-discriminating tests are deleted, the redundant `Map.new` is gone.

**Checked, no issue.** The sentinel never reaches callers or the cache and cannot collide with a stored value (database values are never atoms). A missing or timed-out cache process still degrades to a database query. The `secret_key_base` setup in tests is test-only: this repo has no parent endpoint, and `settings_test.exs` does the same.

**Remaining nits, non-blocking.** The `settings_json_nil_false_test.exs` moduledoc says that on main "every read here takes the `:noproc`-absorbed path"; this file starts a live cache, so on main the reads come back nil because the miss-fill never runs. The description of fix 2 mentions `false`, but `value_json` is a map field, so only `nil` can occur at the top level.

**Tests.** Full suite at 09d1ef8b: 42 doctests, 5018 tests, 1 failure, the MediaBrowserTest duplicate-selector case that reproduces on main and is unrelated. Format and credo are clean.

PR #811, opened 2026-09-14
Add label size to the persisted Etcher line params

alexdont:
Etcher 0.13 adds a **label size** control beside thickness and opacity in the annotation style panel, so a set of labels can be given the same size instead of each ending up wherever its box happened to land.

A size set on a *shape* rides that shape's `style` and already persists with the annotation. The **default** new labels start at travels the line-params road instead — out on `etcher:line-params-changed`, into the user's `custom_fields`, and back in through the `:line_params` attr on the next mount — and `sanitize_line_params/1` dropped it on the way through, keeping only `width`, `opacity` and `dash`. So the default was set, looked right for the session, and was gone on reload.

## The change

`sanitize_line_params/1` now keeps `font_size`, clamped to the control's own 6..200 and **rounded** — the control only offers whole pixels, and a stored 17.5 would come back as a size the user cannot type back in.

## Two things that are deliberate

**It is NOT in `@default_etcher_line_params`.** Having no value is what "size each label by its own box" means, and that is what every label did before the control existed. A default here would silently pin every board to one size.

**That absence is also what lets a cleared size clear.** Etcher omits the key entirely when the user unpins a size (rather than sending null), so the existing `Map.merge(@default_etcher_line_params, clean)` does the right thing: an omitted key reads as "back to sizing by the box" rather than "leave the old one alone". Merging into the previously stored map instead would make a cleared size impossible to clear — worth knowing before anyone refactors that line.

## Checks

`mix precommit` clean (compile --warnings-as-errors, deps.unlock --check-unused, quality.ci, test.js).

Pairs with etcher's side of the round trip, which was dropping the same key on both the emit and the seed; that is released separately.


alexdont:
Closing for now — the Etcher work these depend on is still in progress and nothing is ready to release yet. The branch stays on the fork; I'll reopen this (or fold it into the release PR) once 0.13.3 is actually cut.

PR #812, opened 2026-09-14
Add the Etcher arrow tool to the media viewer's toolbar

alexdont:
Etcher 0.13 adds a **single-arrow tool**: a line with one head, drawn in the selected stroke style, with no label unless the user double-clicks for one.

A dimension is a *measurement* — two heads, and a label it drops you into writing because it is unfinished without one. Pointing at something is the other thing people reach a line-with-an-end for, and it wanted its own tool rather than a dimension with one head deleted.

## The change

The media viewer names its tools explicitly, so it has to opt in. `:arrow` sits next to `:dimension` and `:line`, the other two-ended tools.

Both embeds get it — the image viewer in the template and the board in the component. They offer the same set, and a tool present in one and missing from the other is the kind of difference nobody remembers was deliberate.

## Safe to land ahead of the Etcher release

The toolbar builds a button only for tools it knows (`TOOL_DEFS`) and skips anything it doesn't, so against the currently pinned etcher 0.13.2 this is **inert, not broken** — no missing icon, no dead button. The tool starts appearing when the CDN pin and lock move to the release that ships it.

That pin bump is deliberately *not* in this PR: `vendored_cdn_pins_test.exs` requires the jsDelivr tag and `mix.lock` to agree, and 0.13.3 isn't published yet. It belongs in the release PR that moves both together.

## Checks

`mix precommit` clean (compile --warnings-as-errors, deps.unlock --check-unused, quality.ci, test.js).


alexdont:
Closing for now — the Etcher work this depends on is still in progress and nothing is ready to release yet. The branch stays on the fork; I'll reopen this (or fold it into the release PR) once 0.13.3 is actually cut.

PR #815, opened 2026-09-15
Media reorganizer: engine, Source contract, module callback and mix task

timujinne:
## Why

Every module now places new attachment folders where the host's `attachments_parent_folder` / `attachments_folder_name` hooks say, but folders created before those hooks stay at the media root. Moving them was host-side code so far (Andi has its own reorganizer, tied to its containers and orders). A host with only the catalogue had no way to tidy up. This PR puts the mechanism in core; each module contributes a plan for its own records.

## What

- `PhoenixKit.Modules.Storage.Reorganizer` — the engine: collects plans from enabled modules, drops no-ops, applies every action in its own transaction and never halts. Per action: file/link counts verified before and after, trashed folder restored first (only the subtree rows whose `trashed_at` equals the moved folder's own — a row trashed on its own AFTER the folder around it stays trashed; a row trashed BEFORE the folder was already re-stamped with the folder's own `trashed_at` by `Storage.trash_folder/1`'s subtree `update_all` and is restored along with it — that's `Storage`'s own trash/restore semantics, out of scope here, tracked against the `Storage.restore_folder/2` follow-up), `Storage.update_folder` for move/rename, name collision either resolved with the first free `"name (N)"` (`on_conflict: :suffix`, pre-checked with a SELECT because a failed UPDATE poisons the transaction) or reported (`:report`), optional `after_move` back-fill inside the same transaction. `:trash` only soft-deletes an empty folder. Nothing is hard-deleted; the engine never creates folders. A source returning actions with an unknown key logs one warning per source per key-set (not per action).
- `PhoenixKit.Modules.Storage.Reorganizer.Source` — the contract a module implements (`plan/2` → list of plain maps, so a module compiles against a core without this engine), `Reorganizer.Action` validates and normalises them.
- `PhoenixKit.Module.media_reorganizer/0` optional callback (default `nil`) and `ModuleRegistry.all_media_reorganizers/0` over enabled modules.
- `mix phoenix_kit.media.reorganize [--apply] [--source key …] [--pending-days N]` — dry run by default, prints a per-source/kind table and details; exit 1 after `--apply` if anything failed or conflicted. Option parsing (`strict:` switches, `--pending-days`) is validated BEFORE `app.start` — a malformed option never falls through to a real run. `--source` keys are validated right AFTER `app.start` (the module registry isn't available before it): any unknown or disabled key halts with exit 1 listing every bad key at once — there is no silent "ran fine, planned nothing" fallback. Releases without mix call `Reorganizer.run(actor, apply?: true)`.

Outcomes: `moved | renamed | moved_renamed | backfilled | restored | conflict | failed | trashed | reported`.

## Verified

- 68 engine/mix-task tests (`test/modules/storage/reorganizer_test.exs` + `test/mix/tasks/phoenix_kit.media.reorganize_test.exs`), 0 failures. Full scope run (`test/modules/storage/`, `test/mix/tasks/`, `test/phoenix_kit/module_registry_test.exs`): 379 tests, 0 failures. `mix format --check-formatted`, `mix credo --strict`, `mix compile --force --warnings-as-errors` and `mix dialyzer` all clean.
- End to end on the Andi dev stand with the catalogue Source (companion PR in phoenix_kit_catalogue): on a pre-reorganisation snapshot the plan matched Andi's own reorganizer folder for folder (100/100), `--apply` moved 99, back-filled 1 pointer, trashed 1 stale pending folder, file count unchanged, second run plans nothing; on the already-reorganised stand the plan is empty.

No migration, no version bump.


timujinne (commented):
Independent review (six rounds, 2026-09-15/16), verdict PASS at db894bdf.

Checked: every action runs in its own transaction with file/link counts verified before and after the write; a trashed folder restored during a move restores its subtree in the same transaction (rows trashed after the folder stay trashed; rows trashed before it were already re-stamped by Storage's trash_folder and restore with it — documented, tracked as a Storage follow-up); name collisions are pre-checked (a failed UPDATE would poison the transaction) and resolved with a suffix or reported; nothing is hard-deleted and the engine never creates folders; a failing Source, an invalid action or an exception inside after_move never loses the report; mix task parses options strictly before app.start and validates --source keys right after it (unknown key → exit 1); Source moduledoc carries the full contract the seven module Sources implement. 68 engine/mix-task tests, 379 in the touched scope, format, credo --strict, compile --warnings-as-errors and dialyzer clean; CHANGELOG.md, mix.exs and mix.lock identical to upstream/main; branch rebased on upstream/main.

Verified end to end on the Andi dev stand with the seven module Sources against a pre-reorganisation database snapshot: 110 folder moves, file count unchanged, second run plans only reports.

PR #816, opened 2026-09-15
Fix ProcessFileJob discarding every PDF on hosts with poppler

timujinne:
## Problem

On a host with poppler installed, every PDF upload was discarded by Oban after three attempts and never got a preview.

`ProcessFileJob.process_pdf/1` merged pdfinfo's **string-keyed** fields (`"page_count"`, `"author"`, `"creator"`, `"creation_date"`, `"title"`) into the **atom-keyed** `%{status: "active"}` update. `Ecto.Changeset.cast/3` rejects a mixed-key map:

```
** (Ecto.CastError) expected params to be a map with atoms or string keys, got a map with mixed keys:
   %{:status => "active", "author" => "Hydroforce Engineering OÜ", "creation_date" => "...", "page_count" => 6, ...}
```

The raise happens on the metadata step, before any variant is rendered. Without poppler `PdfProcessor.extract_metadata/1` returns `%{}` and the merge happened to be clean — which is why the crash only surfaced once the tool was installed, and why PDF previews have effectively never worked on a fully provisioned host.

None of those fields are columns on `phoenix_kit_files`; the row has a `metadata :map` for exactly this.

## Fix

- `PdfProcessor.file_attrs/2` builds `%{metadata: Map.merge(existing || %{}, pdf_fields)}`.
- `process_pdf/1` passes that to `update_and_log_metadata/2` instead of the raw pdfinfo map.
- Unit test for `file_attrs/2`, including a cast through `File.changeset/2` next to the atom-keyed `:status`.

## Verified

On a host database with three PDFs stuck at `original` only: re-enqueued `ProcessFileJob` for each — all three completed on attempt 1, each now has `thumbnail`/`small`/`medium`/`large` JPEG instances, `metadata` holds the pdfinfo fields, and a file stuck in `status = "processing"` since April moved to `active`. The thumbnail serves as `image/jpeg` through the file route.

`mix precommit` green (compile, format, credo --strict, dialyzer, JS tests); the new test runs against the isolated `phoenix_kit_test` database.


timujinne (commented):
## Review (two-stage: spec compliance, then code quality)

### Summary

The fix is correct and complete. `PdfProcessor.file_attrs/2` nests pdfinfo's string-keyed fields under the file's `:metadata` map, so `process_pdf/1` no longer hands `Storage.update_file/2` a mixed atom/string-keyed map. Every other metadata-producing path in `process_file_job.ex` was checked: image (`extract_image_metadata`, atom keys), video (`parse_ffprobe_output`, atom keys), non-PDF documents (`extract_document_metadata/2` always returns `{:ok, %{}}`). PDF was the only string-keyed producer. `media_detail.html.heex:250-266` already reads `file.metadata["page_count"|"author"|"title"]`, so the job's output now matches what the template expected all along.

The CHANGELOG claim that `extract_metadata/1` returns `%{}` without poppler is accurate (`pdf_processor.ex:87-98`, both the non-zero-exit branch and the `:enoent` rescue).

### Stage 1 — spec compliance: PASS

### Stage 2 — code quality

**IMPROVEMENT - MEDIUM — CHANGELOG entry filed under an already-released version.** `CHANGELOG.md`: the entry was added under `## 2.23.2 - 2026-09-15`, but `v2.23.2` is tagged and on hex; the file has an `## Unreleased` section for exactly this. → **Fixed in 85716148b9**: moved to `## Unreleased` / `### Fixed`.

**NITPICK — tests cover the helper, not the call site.** `test/modules/storage/services/pdf_processor_test.exs` exercises `file_attrs/2` (plus a cast through `File.changeset/2`), but nothing runs `ProcessFileJob.process_pdf/1`; a revert of the one-line change at `process_file_job.ex:178` would not go red. A job-level test would need to stub `extract_metadata/1` or ship a fixture PDF and require poppler in the test environment; left as is for this PR.

Checks: the 4 new tests pass (`MIX_ENV=test PGDATABASE=phoenix_kit_test mix test test/modules/storage/services/pdf_processor_test.exs`), `mix credo --strict` clean on the changed files, `mix precommit` green via the commit hook. Live verification on a host DB: three PDFs that had only an `original` instance got thumbnail/small/medium/large JPEG variants on attempt 1, metadata populated, thumbnail served as `image/jpeg`.

### Verdict: PASS

PR #817, opened 2026-09-15
Keep the original's extension on storage temp copies and report a missing binary as not installed

timujinne:
## Problem

Two storage-processing defects, both seen on a production host (Andi):

1. **`.ico` uploads never get variants.** `Storage.retrieve_file/1` copies the stored original to `System.tmp_dir!()/phoenix_kit_<random>` — no extension (`generate_temp_path/0`; `Manager.retrieve_file/2` does the same when no `:destination_path` is given). ImageMagick identifies some formats by extension alone, ICO among them, so `ImageProcessor.extract_dimensions/1` got `identify: no decode delegate for this image format` for every variant, `ProcessFileJob` ended with `All variant generations failed` and the job was discarded after 3 attempts. Installing ImageMagick's ICO support doesn't help: on a host whose ImageMagick 7.1.1 reads ICO, the very temp copy `retrieve_file/1` produced for a real branding `.ico` failed `identify` as-is and read as `ICO 256x232` once renamed to `.ico`.
2. **A missing binary is never reported as missing.** `System.cmd/3` on a command that isn't on `PATH` raises `%ErlangError{original: :enoent, reason: nil}`. `PdfProcessor.first_page_to_jpeg/3`, `PdfProcessor.extract_metadata/1` and `System.Dependencies.check_command/2` tested `e.reason == :enoent`, so the not-installed branch never ran: a host without poppler logged `pdftoppm error: nil` for every PDF, and `check_imagemagick/0` / `check_ffmpeg/0` returned `{:error, "Error checking …: nil"}` instead of `{:error, :not_installed}`.

Related: #816 fixes the next failure a PDF hits once poppler *is* installed (mixed-key metadata cast in `ProcessFileJob`). No overlap in code — both touch `pdf_processor.ex`, in different functions — and the CHANGELOG entries sit in different sections.

## Fix

- `Storage.retrieve_file/1` and `Manager.retrieve_with_failover/3`'s default destination append `Path.extname/1` of the stored path — what `Manager.replicate_to_buckets/3` already does. Stored instance names carry the extension (`<hash>_original.ico`). `AnnotationThumbnail` reads its source through `retrieve_file/1` too.
- The three `rescue` clauses check `e.original` (and print it in the fallback message).
- CHANGELOG: `Unreleased` → `Fixed`, no version bump.

## Tests

- `test/modules/storage/missing_binary_test.exs` points `PATH` at an empty directory: `first_page_to_jpeg/3` → `{:error, :poppler_not_installed}`, `extract_metadata/1` logs `pdfinfo not installed`, `check_poppler/0`, `check_imagemagick/0`, `check_ffmpeg/0` → `{:error, :not_installed}`.
- `test/integration/storage/retrieve_file_extension_test.exs` stores an `.ico` in a local bucket and retrieves it: the temp path ends in `.ico` with identical bytes, and (when ImageMagick is present) `identify` reads a real ICO from it. The test resets `Manager`'s `:persistent_term` bucket cache in setup/on_exit — without that a test reads through the previous test's removed bucket.
- Against `lib/` from `main` (fix reverted, tests kept) all 5 fail with the production symptoms: `{:error, "pdftoppm error: nil"}`, `pdfinfo error: nil`, `{:error, "Error checking identify: nil"}`, `Path.extname(path) == ""`. With the fix: 5 tests, 0 failures, repeated with seeds 1–5.
- `test/integration/storage`, `test/modules/storage`, `test/phoenix_kit/modules/storage` against an isolated `phoenix_kit_tempext_test` database: 137 tests, 0 failures.
- `mix precommit` green end to end: `compile --warnings-as-errors`, `deps.unlock --check-unused`, `test.compile`, `quality.ci` (format check, `credo --strict`, dialyzer passed), JS tests.


timujinne (commented):
**Verdict: PASS**

## Spec compliance

1. **`.ico` temp-copy extension** — Met. `Storage.retrieve_file/1` (lib/modules/storage/storage.ex:2820) and `Manager.retrieve_with_failover/3`'s default destination (lib/modules/storage/services/manager.ex:225-227) now append `Path.extname/1` of the stored `file_name`/`file_path`, mirroring the pre-existing pattern in `Manager.replicate_to_buckets/3` (manager.ex:129). Verified live: with the fix reverted (lib-only patch applied in reverse, tests kept, then reapplied — worktree left clean, `git status` empty throughout), `Storage.retrieve_file/1` on a stored `.ico` returns a temp path with no extension and `identify` fails with "no decode delegate"; with the fix, the temp path ends in `.ico` and `identify` reports `ICO`.
2. **`e.original` vs `e.reason`** — Met. `PdfProcessor.first_page_to_jpeg/3` (pdf_processor.ex:67-71), `PdfProcessor.extract_metadata/1` (pdf_processor.ex:98-102) and `System.Dependencies.check_command/2` (dependencies.ex:170-176) all now check `e.original == :enoent`. Confirmed empirically on this runtime (Elixir 1.19.5 / OTP 28) that `%ErlangError{}`'s `reason` field is always `nil` and `original` always carries the real term (`:enoent` for a missing binary, the raw term otherwise) — so the fix is correct not just for the `:enoent` branch but also for the `else` branch's `inspect(e.original)` fallback message on non-enoent errors.
3. **CHANGELOG** — Met. `## Unreleased` → new `### Fixed` subsection (CHANGELOG.md:20-36), no version bump (`mix.exs` `@version` untouched, matches diff stat).

Tests present as specified: `test/modules/storage/missing_binary_test.exs`, `test/integration/storage/retrieve_file_extension_test.exs`.

## Findings

None — no bugs found. Two minor nitpicks:

1. **NITPICK** — `test/modules/storage/missing_binary_test.exs:46` — The third test's `Dependencies.check_poppler() == {:error, :not_installed}` assertion doesn't actually exercise this PR's fix: `check_poppler/0` (dependencies.ex:52-63) has its own inline `System.cmd("pdftoppm", ...)` call with a blind `rescue _error -> {:error, :not_installed}` that never inspected `e.reason`/`e.original` to begin with — it was already correct before this PR (and stays correct if reverted, I confirmed only `check_imagemagick`/`check_ffmpeg` actually flip from failing to passing when the lib patch is reverted; `check_poppler`'s assertion holds either way). Only `check_imagemagick()` and `check_ffmpeg()` in that same test genuinely regression-test the `check_command/2` fix. Not wrong, just slightly overstates coverage — the PR description's test summary line has the same framing.
2. **NITPICK** — CHANGELOG.md:29 groups `pdftoppm`, `pdfinfo`, `identify` and `ffmpeg` under one sentence ("A missing `pdftoppm`, `pdfinfo`, `identify` or `ffmpeg` is reported as not installed"), which reads as if all four go through the same code path. In fact `pdftoppm`/`pdfinfo` are fixed in `PdfProcessor`'s own rescue clauses while `identify`/`ffmpeg` are fixed via the shared `System.Dependencies.check_command/2` — two different call sites hitting the same `e.original`-vs-`e.reason` root cause. Accurate in substance, just worth knowing the sentence conflates two sites for readers who go looking for a single fix location.

## Follow-ups (out of scope)

- `ImageProcessor` (image_processor.ex), `AnnotationThumbnail.run_convert/2` (annotation_thumbnail.ex) and `VariantGenerator`'s ffmpeg/ffprobe calls (variant_generator.ex:443, process_file_job.ex:247) all shell out via `System.cmd` without ever checking `e.original`/`e.reason` for a missing-binary case specifically — most either have a broad `rescue e -> {:error, inspect(e)}` (which does surface `original: :enoent` inside the raw `inspect`, so it's not misleading the way the fixed call sites were) or, for `ffmpeg`/`ffprobe` in variant_generator.ex/process_file_job.ex, no local rescue at all (an unhandled `ErlangError` would propagate to the caller). Not part of this PR's stated scope and not a regression it introduces — flagging only as a possible next cleanup pass.
- Checked PR #816 (mixed-key metadata cast) for a real conflict: it inserts a new `file_attrs/2` function into `pdf_processor.ex` right after `extract_metadata/1`'s closing `end` (context anchored at the pre-existing `{:ok, %{}} end` boundary), while this PR only edits the two `rescue` clauses above that point. No overlapping hunks; the two branches touch disjoint regions of the same file and should merge cleanly in either order.

## Verification

- Worktree `/root/andi-work-pk-temp-ext`, branch `fix/storage-temp-file-extension`, HEAD `9aee851a` — confirmed equal to `gh pr view 817 --json headRefOid`.
- `git diff upstream/main...HEAD --stat` — 7 files, +181/-9, matches `gh pr view` additions/deletions.
- `cd /root/andi-work-pk-temp-ext && PGDATABASE=phoenix_kit_tempext_test MIX_ENV=test mix test test/modules/storage/missing_binary_test.exs test/integration/storage/retrieve_file_extension_test.exs` → **5 tests, 0 failures** (DB reachable — no "Cannot reach test database" warning; ImageMagick present, `identify -version` → 7.1.1-43, so the ICO-identify test ran for real, not silently skipped by its `imagemagick?()` guard).
- Reverted the fix without touching tests (`git diff upstream/main...HEAD -- lib > /tmp/pk817-lib.patch; git apply -R /tmp/pk817-lib.patch`), re-ran the same 5 tests → **5 tests, 5 failures**, each failing exactly as the PR description predicts (`{:error, "pdftoppm error: nil"}`, `pdfinfo error: nil` in the log, `{:error, "Error checking identify: nil"}`, `Path.extname(path) == ""`, `identify: no decode delegate for this image format`). Reapplied the patch (`git apply /tmp/pk817-lib.patch`) and confirmed `git status --short` is empty — worktree returned to the exact PR-head state throughout.
- `PGDATABASE=phoenix_kit_tempext_test MIX_ENV=test mix test test/modules/storage/ test/integration/storage/` → **127 tests, 0 failures** (no regressions in the surrounding storage suite).
- Checked whether `missing_binary_test.exs`'s global `System.put_env("PATH", empty_dir)` (an `async: false` test) can leak into a concurrently-running `async: true` test (e.g. `image_processor_sanitize_test.exs`, which shells out to `identify`/`convert`) and cause flaky, unrelated failures. Empirically probed this exact toolchain (Elixir 1.19.5 / OTP 28) with a standalone two-module ExUnit script: async and sync (`async: false`) test phases run **sequentially, not concurrently** (total wall time = async time + sync time, not `max(async, sync)`), so this is not a real flakiness risk on this runtime. `:persistent_term` bucket-cache reset in `retrieve_file_extension_test.exs` is erased in both `setup` and `on_exit`, which is sufficient given the same sequential-sync guarantee.
- `mix precommit` (compile --warnings-as-errors --all-warnings, deps.unlock --check-unused, test.compile, quality.ci [format --check-formatted, credo --strict, dialyzer], test.js) → **exit 0**, fully green. Grepped the log for the PR's touched files (`storage.ex`, `manager.ex`, `pdf_processor.ex`, `dependencies.ex`) — zero hits among the compiler/credo/dialyzer warnings, confirming all warnings printed are pre-existing and unrelated to this diff.
- Grepped every caller of `Storage.retrieve_file/1` and `Manager.retrieve_file/2` in `lib/` (annotation_thumbnail.ex, process_file_job.ex ×3, variant_generator.ex, file_controller.ex ×3). The three `file_controller.ex` sites always pass an explicit `destination_path:`, so they bypass the new default-extension branch entirely and are unaffected. The others either are the intended beneficiaries of the fix or only use the returned path opaquely (`File.rm/1`, passed through to `System.cmd` as an input argument) — no caller does `Path.rootname`/basename-splitting on the returned temp path that an added extension would break.

PR #818, opened 2026-09-16
Add <.decimal_input> and Number.parse_decimal/2 for free decimal form fields

timujinne:
## Why

Every module that takes a quantity, price or measurement re-implements "replace the comma with a dot" by hand — catalogue 6×, warehouse 5×, entities, manufacturing, and host apps (Andi: 16 places) — each slightly differently. Most of those fields render `<input type="number">`, whose decimal separator follows the browser locale (on a `lang="et"` page a typed `2.5` submits as `""`; on `lang="en"` a typed `2,5` does) and whose `step` silently blocks a `phx-submit` for any other precision. The immediate trigger was BeamLabEU/phoenix_kit_catalogue#121, where the item picker rounded `2,5` to 3.

## What

**`<.decimal_input>`** (`PhoenixKitWeb.Components.Core.DecimalInput`, imported in `core_components` next to `<.input>`): a plain `<input type="text" inputmode="decimal" autocomplete="off">` with the `<.input>` contract — `field` or `name`/`value`, `label`, gettext-translated `errors`, required marker, `class` on the control / `wrapper_class` on the `phx-feedback-for` wrapper — plus a `unit` suffix rendered inside the field (daisyUI 5 `<label class="input">`). A binary value is echoed back unchanged so `2,5` does not turn into `2.5` under the cursor on re-render; numbers are shown normalized (`2.500` → `2.5`, `1E+3` → `1000`).

**`PhoenixKit.Utils.Number.parse_decimal/2`** (+ `parse_decimal!/2`, `format_decimal/1`) — the server side of the same field:
- dot or comma is the decimal point; spaces (no-break/thin too) are grouping;
- both kinds present → the last one is the decimal point, the other must be valid 3-digit grouping (`1.234,56`, `1,234.56`); one kind repeated → grouping (`1,234,567`); `2..5`, `1,23,4` are `:invalid`;
- optional sign; exponents, `NaN`, `Infinity`, hex, stray letters → `{:error, :invalid}`; blank/nil → `{:error, :empty}`;
- `:min`/`:max` reject (`:below_min`/`:above_max`), never clamp; a magnitude ≥ 10¹² is `:above_max` even without `:max` (keeps pasted `1e1000000`-class strings away from arithmetic);
- integers/floats/Decimals pass through as a normalized `Decimal`.

No CHANGELOG/version edits (maintainer's).

## Tests

- `test/phoenix_kit/utils/number_parse_decimal_test.exs` (16) and `test/phoenix_kit_web/components/core/decimal_input_test.exs` (8), plus doctests.
- Full suite: 5226 tests, 4 failures, all pre-existing/unrelated: `media_browser_test.exs:567` fails identically on `main`; the three role/permission tests pass when run on their own (order-dependent).
- `mix quality.ci` (format, credo --strict, dialyzer): clean.

## Follow-ups (separate PRs, after release)

Switch catalogue (item form, picker stepper), warehouse, entities, manufacturing and Andi to `<.decimal_input>` + `parse_decimal/2`, deleting the local comma parsers.


timujinne (commented):
Review (two rounds, spec compliance then code quality; verified on the branch, not on the description).

Round 1 — PASS with one blocking finding:
- MAJOR: no length guard before parsing — an 8 MB text field cost seconds of CPU (10M digits ≈ 3.9 s, linear). Fixed in f7069d59: inputs over 64 bytes are rejected before any regex or `Decimal.parse` runs (8 MB string now returns `{:error, :invalid}` in ~1 µs).
- MINOR: `\s` swallowed tabs/newlines as grouping while the doc promised spaces. Fixed: only space characters (ASCII, NBSP, thin, narrow NBSP) are grouping; `"\t5\n"` is `:invalid`.
- NITPICK: unused `use Gettext` in DecimalInput. Removed.

Round 2 (f7069d59) — PASS, all three verified live; no new issues.

Also checked: no `decimal_input/1` name collision across core and every phoenix_kit_* module; the FormField `value` derivation uses the explicit nil-check rather than `assign_new` (the checkbox regression pattern); adversarial inputs (exponents, NaN, Infinity, hex, unicode/fullwidth digits, mixed signs) all reject; `mix compile --warnings-as-errors`, `credo --strict`, dialyzer clean; 26 targeted tests green.

timujinne (commented):
Round 3 (c56236f1: integers keep exponent 0) — PASS.

Verified `plain_normalize/1` independently against sign (`-10`, `-1000`, `-0`), zero (`0`, `0.0`), the 10^12 ceiling boundary, `Decimal` inputs with a positive exponent (`Decimal.new("1E+3")`, `Decimal.new("-5E+2")`) and float input (`1.0e3`) — all produce the expected plain-form `Decimal`, structurally `==` to `Decimal.new/1` of the same text.

`Integer.pow/2` has been available since Elixir 1.12; this project requires `~> 1.18`.

NaN/Infinity are safe through `plain_normalize` because both carry `exp: 0` in their struct, so the `exp > 0` guard never fires — confirmed on the raw struct fields and by calling `format_decimal/1` on both without error.

`mix credo --strict`, `mix dialyzer` and both test files (27 tests) pass clean.

Non-blocking note: for `Decimal`-struct input (not the string path, capped at 64 chars) a caller-supplied `1E+400` builds the multiplied-out bignum before the ceiling check rejects it — measured at ~0.8 ms, internal callers only.

PR #819, opened 2026-09-16
parse_decimal/2: integers keep exponent 0, never 1E+1

timujinne:
Follow-up to #818 (found by a host test minutes after the merge).

`Decimal.normalize/1` strips trailing zeros on both sides, so `parse_decimal("10")` returned `1E+1`: `Decimal.equal?` to `Decimal.new("10")` but not `==`, and printed in exponent form by `to_string/1`, Jason and Postgres text casts. A host test comparing a parsed discount with `==` against `Decimal.new("10")` caught it.

Fix: a private `plain_normalize/1` — `Decimal.normalize/1`, then if the exponent went positive, multiply it back into the coefficient (`exp: 0`). Trailing fraction zeros still go (`"2.500"` → `2.5`, `"10.0"` → `10`); integer digits stay. `format_decimal/1` follows the same rule (`Decimal.new("1E+1")` → `"10"`). NaN/Infinity carry `exp: 0`, so the guard never fires for them. Docs updated.

Tests: 29 in `number_parse_decimal_test.exs` + `decimal_input_test.exs`, credo --strict clean. No CHANGELOG/version edits.


timujinne (commented):
Verified against the live app (mix run) as well as reading the diff — every claim in the description checks out:

- `parse_decimal("10")` now `==` `Decimal.new("10")` (not just `Decimal.equal?`); `to_string` prints `"10"`, not `"1E+1"`.
- Sign handling unaffected (`-10`, `-1000`, `-0` → unsigned `0`) — `plain_normalize` only rewrites `coef`/`exp`, the sign stays untouched; order is plain_normalize → bound → unsigned_zero.
- Trailing fraction zeros still strip (`"2.500"` → `2.5`), matching prior behaviour.
- `Decimal` inputs with a positive exponent (`1E+3`, `-5E+2`) and floats (`1.0e3`) all normalize correctly.
- NaN/Infinity carry `exp: 0` in their structs, so the new `exp > 0` guard never touches them — `format_decimal` still prints "NaN"/"Infinity" without error.
- The 10^12 magnitude bound and the `999999999999` boundary still evaluate correctly after normalization; a synthetic `coef: 1, exp: 300` is rejected via `:above_max` rather than crashing.
- `Integer.pow/2` is fine under the project's `~> 1.18` Elixir requirement.
- `mix format`, `mix credo --strict`, `mix dialyzer` and the targeted test files (29 tests) all pass clean.

No issues found. PASS.

PR #822, opened 2026-09-17
Let a Telegram integration link a group, and say what it linked

timujinne:
## What this fixes

A bot added to a Telegram group cannot be linked at all today.

Capture filters updates to `chat["type"] == "private"` (`my_integration_form.ex`), so a group update is discarded even when it arrives — and the Notifications card offers no other way in: it renders the linked ids and an all-or-nothing **Unlink**, with no field to type an id the operator already knows.

The delivery side was never the problem. `Channels.Telegram` sends to whatever is in `chat_ids`, and `sendMessage` accepts a group id as happily as a private one — so a group id placed in that list by hand delivers fine. The gap is purely the entry path.

## What changed

- **Capture accepts `group` / `supergroup`.** The single-mode lock becomes per-kind: it still pins ONE private chat against a stranger who messaged the bot moments later, but it no longer blocks a group — a group is only ever linked by someone who can post there and deliberately ran the command.
- **Link a chat by id.** Capture only reaches updates still inside Telegram's ~24h queue; an id you already know shouldn't depend on that window.
- **Unlink one chat.** With a DM and a group linked, the all-or-nothing button can't express "drop the group, keep mine". The old button stays for the full reset.
- **Linked chats are named, not numbered.** Kind (Direct message / Group / Channel) plus the captured title, so the list can be audited. Kind falls back to the id's own shape for rows linked before this — Telegram signs group ids negative.
- **Test stops reporting success it doesn't have.** A capture that found nothing was folded into "Connection works" — exactly the state where no notification will ever arrive. It now says so, and names the group command.
- **The card and the provider instructions explain linking.** Privacy mode means a bot never sees ordinary group chatter, so "message the bot, then press Test" is not an instruction that works in a group; the instructions stopped at pasting the token, which delivers nothing on its own.

The rules move into `PhoenixKit.Integrations.Telegram.ChatLink`, where they can be proven without driving a form event.

## Not in scope

`multi` mode's trust model. Anyone who starts the bot is linked on the next Test and receives everything sent to that connection — with no approval step, no allowlist, and (until this PR) no way to see or remove one subscriber. That needs a design discussion, not a quiet change inside this one; the card now at least warns what the mode means, and per-chat unlink makes a mistake recoverable.

## Tests

- `ChatLinkTest` — 14 unit tests over capture, the per-kind lock, multi union, and id normalisation.
- `MyIntegrationFormChatLinkingTest` — drives the real LiveView events: link by id, junk refused, no duplicates, per-chat unlink, and that the card states the group command.
- `ProvidersTest` — the Telegram instructions cover linking, and cover a group.

Full suite on this branch: 42 doctests, 5302 tests, 0 failures.


timujinne (commented):
## Review

Two rounds, independent of the author. First round was a **FAIL** on a real bug; it has been fixed and re-verified. Final verdict on `de7f27d7`: **PASS**.

### Blocker found and fixed

`ChatLink.lockable_private/2` decided "is a private chat already linked?" by testing the stored id for a leading `-`. But `normalize_chat_id/1` also accepts `@channelusername`, which has no minus and is not a private chat — so a hand-linked channel counted as the owner's own chat, the single-mode lock closed, and a private chat captured later by `/start` was silently discarded forever. The PR's own new instructions steer people into exactly that sequence: link the id you already know, then press Test.

Repro on the original commit:

```elixir
ChatLink.merge("single", ["@myannounce"], [%{"id" => "428897538", "type" => "private"}])
#=> {["@myannounce"], []}     # the owner's chat never links
```

Fixed by classifying ids by kind (`ChatLink.kind/1` → `:group | :channel | :private`) and asking the lock about the kind. Same repro on `de7f27d7` now returns `{["@myannounce", "428897538"], ["428897538"]}`.

### Also fixed

- `chat_meta` was written for every chat seen in a poll — including ones the lock had just refused — and was never pruned, so it would have grown into a record of everyone who ever messaged the bot. Capture and every write now derive it from the linked list and prune to it; the all-chats unlink goes through the same path.
- `:unreachable` produced no distinct message and fell through to "Connection works"; it now says the chats could not be read.
- Two identical bot-handle regex helpers became one.
- The new chat-id input had no DOM id, leaving its `<label for>` empty.
- An over-long moduledoc line was wrapped, and "at least one captured chat" corrected to "linked chat" — accurate now that linking by id is a real path.

### Raised and deliberately not changed

`link_chat` does not consult `mode`, so manual linking is not capped in single mode. Kept: the lock exists to stop an unauthenticated stranger who messaged the bot, not to cap a deliberate act by the connection's owner — and capping it would defeat the PR's purpose, since the owner needs a private chat *and* a group at once. Both stale doc sites (`ChatLink.merge/3`, the notifications channel moduledoc) were corrected to say a mode governs auto-capture only.

Still out of scope, and called out in the PR body rather than quietly changed: `multi` mode's trust model.

### Gates

`mix format --check-formatted` clean, `mix credo --strict` clean on the changed files, full fork suite 63 doctests + 5605 tests, 0 failures, with the integration tag actually running (no `Excluding tags` banner).

Known coverage gap, accepted: no end-to-end test drives the Telegram capture path through the Test button, because the core has no seam for stubbing `Req` there and adding production code purely for a test seam was not worth it. The decision logic it would cover lives in `ChatLink.capture/4` and is unit-tested; the LiveView clause is three lines of call.

PR #828, opened 2026-09-18
Storage: host-registered file reference sources for orphan detection (config :file_reference_sources)

timujinne:
`Storage.orphaned_files_query/0` hard-codes every table that may reference a file, and `mix phoenix_kit.cleanup_orphaned_files` / `DeleteOrphanedFileJob` delete what it returns — so a host whose own tables reference files is one cleanup run away from data loss. Catalogue items already get JSONB-aware checks built into core, but a host app has no way to add its own: Andi's sub-order featured image (`andi_orders.data->>'featured_image_uuid'`), and by extension every file sitting in an order or project folder, is invisible to the check today and looks orphaned in the global listing. This adds a config hook so a host can register its own reference sources instead of needing a core change per table.

**Contract**

```elixir
config :phoenix_kit, :file_reference_sources, [
  {MyApp.Media, :file_reference_sources}
]

def file_reference_sources do
  [
    dynamic(
      [f],
      fragment(
        "NOT EXISTS (SELECT 1 FROM my_app_orders o WHERE o.data->>'featured_image_uuid' = ?::text)",
        f.uuid
      )
    )
  ]
end
```

- Each entry is `{module, function}` or `{module, function, args}`, called for a list of `Ecto.Query.dynamic/2` expressions over the file binding `f`; every expression is appended to the orphan query with `NOT EXISTS`, same shape as core's own catalogue/shop checks.
- Shorthand: a plain `{table, column}` tuple for a native column reference, `{table, :jsonb_key, key}` for a `data->>'key'` pointer.
- A source whose table does not exist, or whose function raises, is skipped with a `Logger.warning/1` instead of failing the orphan query.
- Documented alongside the existing `:protected_file_uuids` hook in `PhoenixKit.Modules.Storage`'s moduledoc, under "Protecting host-owned files".

**Tested**

`test/integration/storage/file_reference_sources_test.exs`: a file referenced only through an `{module, function}` source is not orphaned, and is orphaned again without the config; a source function that raises is skipped with a logged warning instead of crashing the check; both shorthand forms (`{table, column}`, `{table, :jsonb_key, key}`) protect a referenced file against a real (throwaway) host table; a shorthand pointing at a nonexistent table is skipped with a warning instead of crashing.

Full suite: `63 doctests, 5687 tests` — 7-8 failures across two separate full runs, none of them touching this change or its test file, and none reproducing when run in isolation (registration/profile/integrations-settings/role-order/V191-migration/orphaned-FK-doctor tests — pre-existing order-dependent flakiness in the suite). `mix format`, `mix credo --strict` and `mix compile --warnings-as-errors` are clean on the touched file.


timujinne (commented):
## Stage 1: Spec compliance (Block B, section B3)

Verified against `docs/superpowers/specs/2026-09-17-suborder-featured-image-design.md` §B3 at head `b43a56d7`.

- New config `config :phoenix_kit, :file_reference_sources` accepting `{module, function}` / `{module, function, args}` entries returning `Ecto.Query.dynamic/2` expressions over the `f` binding, appended to `orphaned_files_query/0` with `NOT EXISTS` — present, `lib/modules/storage/storage.ex:2588-2640`. Matches the spec's contract example verbatim.
- Optional `{table, column}` and `{table, :jsonb_key, key}` shorthands — present, `storage.ex:2592-2610`, `2656-2681`. Spec explicitly allows these ("implement the shorthand only if it stays small") — the implementer built them; see Stage 2 note on scope below, not a spec violation.
- Guard: a source whose table does not exist is skipped with `Logger.warning`, never a crash — present and verified live: mutating the `host_table_exists?/1` branch away causes the "nonexistent table" test to fail with a raw `Postgrex.Error (undefined_table)`, confirming the guard is load-bearing (`storage.ex:2592-2600`, `2646-2654`).
- Guard: a source function that raises is skipped with a warning, never a crash — present and verified live: mutating away the `rescue` in `apply_reference_source_mfa/4` causes the "bad source" test to fail with `** (RuntimeError) boom`, confirming that guard is load-bearing too (`storage.ex:2629-2640`).
- `mix phoenix_kit.doctor` update — correctly skipped; confirmed by grep that the doctor task does not enumerate orphan-detection/reference sources today, so the spec's "if it lists reference sources" condition does not apply.
- `PhoenixKit.Modules.Storage` moduledoc gets a "Protecting host-owned files" section covering both `:protected_file_uuids` and `:file_reference_sources` — present, `storage.ex:49-89`, correct heading level, accurate.
- Tests: `test/integration/storage/file_reference_sources_test.exs` — the three required cases (referenced-via-host-source not orphaned; orphaned without config; bad source skipped with warning) are present and pass (6/6 including the shorthand extras). `async: false`, consistent with sibling orphan-detection test files (`storage_catalogue_orphan_test.exs`, etc.) that also mutate this same global `Application` config.
- No CHANGELOG.md / mix.exs / @version changes — confirmed via `git diff upstream/main -- CHANGELOG.md mix.exs` (empty).
- Diff is scoped to exactly the two reported files (`git diff upstream/main --stat`).

**Spec Verdict:** PASS

---

## Stage 2: Code quality

### MINOR: Per-shorthand-source table-existence check isn't batched, unlike the built-in check it sits next to
**File**: `lib/modules/storage/storage.ex:2646-2654` (`host_table_exists?/1`), called from `apply_reference_source/2` at `2592-2610`
**Problem**: `existing_optional_tables/0` (pre-existing, `storage.ex` a few lines above) does ONE `information_schema.tables` query to check all of core's own optional tables at once. The new `host_table_exists?/1` instead issues a SEPARATE `information_schema.tables` round-trip per shorthand entry, on every call to `orphaned_files_query/0` — which itself runs once per `file_orphaned?/1` call, i.e. once per file in `DeleteOrphanedFileJob.perform/1`.
**Suggestion**: Batch all configured shorthand table names into a single `WHERE table_name = ANY($1)` query, following the existing `existing_optional_tables/0` pattern, or fold the check into that same query.
**Rationale**: information_schema lookups are cheap individually, but this multiplies with the number of configured shorthand sources on every orphan check in a job queue that processes files one at a time. The implementer flagged this themselves in the report as worth a reviewer glance — agreed, worth tightening, not a blocker.

### MINOR: Column/JSONB-key typos in the shorthand forms crash instead of warning, asymmetric with the table-missing guard
**File**: `lib/modules/storage/storage.ex:2592-2610`, `2656-2681`
**Problem**: The table-existence guard is checked before building the shorthand dynamic, but the column (or JSONB key) name is never validated. A typo in `{"my_table", "colum_typo"}` composes fine (identifiers are just quoted, not checked against the schema) and only fails when the query actually executes, as a raw unrescued `Postgrex.Error` surfacing wherever `file_orphaned?/1` is called (e.g. inside `DeleteOrphanedFileJob.perform/1`) — not the graceful `Logger.warning`-and-skip the spec promises for "a source whose table does not exist." The MFA form's `rescue` doesn't cover this either, since the bad column is baked into a dynamic that isn't executed until later.
**Suggestion**: Either check `information_schema.columns` alongside the table check, or note in the moduledoc that only the table-existence case is guarded — a misconfigured column will still crash the orphan-cleanup job.
**Rationale**: matches the letter of the spec (which only asks to guard "a source whose table does not exist"), so not a spec violation, but a real robustness gap a host will hit via a config typo, with no test coverage for it either way.

### NITPICK: Logger.warning prefix breaks from the module's established "Storage: ..." convention
**File**: `lib/modules/storage/storage.ex:2622-2624`, `2634-2637`, `2643`
**Problem**: Every other `Logger.warning`/`.error`/`.info` call in this module (e.g. `storage.ex:3165`, `3176`, `3346`) is prefixed `"Storage: ..."`. The three new ones use `"phoenix_kit: file_reference_sources ..."` instead.
**Suggestion**: `"Storage: file_reference_sources ..."` for consistency with the rest of the file's log lines.
**Rationale**: cosmetic, but it's the established local convention and these are exactly the kind of ops-facing log lines someone will grep for later.

### NITPICK: catch-all invalid-entry clause has no test
**File**: `lib/modules/storage/storage.ex:2621-2627` (`apply_reference_source(query, other)`)
**Problem**: The defensive fallback for a malformed `:file_reference_sources` entry (wrong shape entirely) logs and ignores, but no test exercises it.
**Suggestion**: A cheap one-liner test (`{:not_a_valid_entry}` in the config, assert it's skipped with a warning) would close the loop, matching the coverage given to every other branch in this file.
**Rationale**: minor completeness gap; every other guard clause here has a dedicated test, this one doesn't.

### Scope note (not a defect)
`storage.ex:2592-2610`, `2656-2681`, plus 4 of the 6 tests: the `{table, column}` / `{table, :jsonb_key, key}` shorthands were explicitly optional per the spec's own wording ("implement the shorthand only if it stays small"). They roughly double the diff (from ~90 lines for the required MFA-only contract to 178). They're well-tested and don't leak complexity into the required path, so I'm not treating this as EXTRA/YAGNI, but it's a judgment call worth naming for the record.

**Quality Summary:** 0 critical, 0 major, 2 minor, 2 nitpick

**Quality Verdict:** Ship

---

## Verification performed

- `PGHOST=localhost PGUSER=postgres PGPASSWORD=postgres MIX_ENV=test mix test test/integration/storage/file_reference_sources_test.exs` → 6 tests, 0 failures.
- Full suite, `PGHOST=localhost PGUSER=postgres PGPASSWORD=postgres MIX_ENV=test mix test`, run twice: first run 7 failures (matching the implementer's own report of pre-existing, order-dependent flakiness under heavy concurrent load — this box was running 3-4 other worktrees' full suites/dialyzer at the same time); **second clean run: 63 doctests, 5687 tests, 0 failures**, confirming none of that flakiness is caused by this change.
- Mutation-tested two of the new guards against an untracked copy (symlinked source tree, only `storage.ex` duplicated) — confirmed both the "bad source raises" and "table missing" tests actually fail when their respective guard is removed (see Stage 1).
- `mix format --check-formatted` (whole project): clean.
- `mix credo --strict` (whole project): "found no issues".
- `mix compile --warnings-as-errors --force`: clean.
- `git diff upstream/main -- CHANGELOG.md mix.exs`: empty.
- `identifier(^table)` / `identifier(^column)` fragment usage confirmed against `ecto_sql` 3.14's Postgres adapter (`quote_name/1`-based identifier quoting) — no SQL-injection concern; the JSONB key and file uuid are ordinary bound parameters, not string-interpolated.

---

## Overall Verdict: PASS

PR #827, opened 2026-09-18
Core PreviewCard: the catalogue product card render (carousel + fields + files) as a reusable core modal component

timujinne:
## What / why

`PhoenixKitCatalogue.Web.Components.ProductCard.product_card/1` is a pure,
DB-free render component — a swipeable image/file carousel with jump-strip
navigation, a fields `<dl>`, and a compact file list, wrapped in a `<.modal>`
— that a second consumer now needs: Andi's sub-order "featured image" card
(catalogue items and Andi sub-orders both open the same kind of preview,
built from data each host already resolves its own way). This PR moves the
render (and its private helpers) into core as
`PhoenixKitWeb.Components.Core.PreviewCard`, generalised in name only. The
catalogue keeps its DB-backed `resolve_images/1`, `resolve_files/1`,
`build_fields/2` helpers untouched and will switch `product_card/1` to
delegate to this component in a follow-up PR once this one is released
(there is no behavior change here for the catalogue — it isn't touched by
this diff at all).

## Contract

- `preview_card/1` — attrs `id` (required), `show` (required), `target`
  (`:any`, required, `nil` allowed — the host may be a LiveView itself),
  `title` (default `nil`, falls back to gettext "Preview"), `images`
  (`[%{uuid, name}]`, default `[]`), `files` (`[%{uuid, name, size, pdf?}]`,
  default `[]`), `fields` (`[{label, value}]`, default `[]`), `on_close`
  (default `"preview_card_close"`), `max_width` (default `"3xl"`); slot
  `extra_actions`. Renders the modal shell around `preview_card_body/1`.
- `preview_card_body/1` — the same attrs minus the modal ones (`id`, `show`,
  `on_close`, `max_width`); the "notpopup" form for embedding the preview
  inline without a modal.
- Gettext via `PhoenixKitWeb.Gettext` (msgids "Preview", "Close",
  "Previous", "Next", "Show image %{number}", "Files", "Open"), extracted
  into `priv/gettext`.
- No config, no migration, no new dependency.

## Testing

- `test/phoenix_kit_web/components/core/preview_card_test.exs` (new, 15
  tests, ported from the catalogue's `product_card_test.exs` render cases):
  carousel slide count and titles, first-image-eager/rest-lazy loading,
  client-side-only slide switching (jump strip + arrows, no server event),
  single-image renders no strip/arrows, files continue the same swipe track
  with the PDF iframe hidden below `sm`, fields render and the empty-fields
[...]

PR #829, opened 2026-09-18
MediaBrowser: optional host-owned featured image (Set as featured in the kebab/viewer, star badge, {:set_featured, uuid} to the host)

timujinne:
## What / why

Two consumers need a host-owned "featured image" pointer on a folder-backed resource: catalogue items (already have their own `featured_image_uuid` convention) and Andi sub-orders (Block A, shipped, stores `order.data["featured_image_uuid"]`). Both currently need a bespoke picker outside `MediaBrowser` to choose that image, because the browser had no way to say "this file is the featured one" or let a user set it in place. This PR gives `MediaBrowser` an opt-in, host-owned featured-image UI — the browser only ever *asks* for a choice and relays it; it never writes anything itself, keeping the persistence and validation (which storage scope, which host table) entirely on the host side.

## Contract

- New optional attr `featured`, default `nil` (feature off). When set, it is `%{uuid: uuid | nil, label: String.t() | nil}` — the host's current pointer. `label` is the star badge's tooltip, defaulting to gettext "Featured image" when nil.
- When `featured` is set, every image tile/row (grid, list, and stack views) gets a "Set as featured" / "Unset featured" kebab item, and the current pointer's tile carries a `data-role="featured-badge"` star overlay next to the existing video/PDF badges. The modal viewer's sidebar shows the same toggle for image files.
- Events `set_featured` (`%{"file-uuid" => uuid}`) and `unset_featured` fire `send(self(), {MediaBrowser, id, {:set_featured, uuid | nil}})` to the host process and optimistically move the component's own `featured.uuid` so the badge/kebab flip immediately, ahead of the host's write landing. Hosts that delegate every `{MediaBrowser, _, _}` message through `handle_parent_info/2` must match `{:set_featured, _}` themselves first — `handle_parent_info/2` does not handle it (documented in the moduledoc).
- `MediaCanvasViewer` (shared with `MediaViewer`/`MediaDetail`) gets a mirrored, off-by-default `:featured` assign so those other consumers are unaffected; its sidebar button posts straight to the parent LiveComponent via the existing `phx-target={"#" <> parent_id}` routing MediaBrowser already uses for close/prev/next, so it needs no event handler of its own.

## Testing

- New `test/phoenix_kit_web/components/media_browser_featured_test.exs`: render assertions (no badge/kebab when `featured: nil`; the matching tile's kebab offers Unset and gets the badge, others offer Set; a non-image file gets neither) via `render_component/2` + `LazyHTML.query/2`; `set_featured`/`unset_featured` event behavior (including the no-op when the host never opted in) via direct `handle_event/3` calls against a plain socket, the style `media_browser_viewer_url_test.exs` uses; and an end-to-end viewer-sidebar test through a tiny `live_isolated/3` host LiveView, confirming the click reaches the host and the badge flips.
- Full existing MediaBrowser/MediaCanvasViewer/MediaViewer/ImageEditor suites re-run to confirm no regression.
- `mix format`, `mix credo --strict` on the touched files, `mix compile --warnings-as-errors`, and the full `mix test` suite all pass — see PR checks / the session report for exact counts.


timujinne (commented):
Review round 1 — Block B2 (`MediaBrowser` featured image), spec: `docs/superpowers/specs/2026-09-17-suborder-featured-image-design.md` §B2. Worktree HEAD `70c22b3407e0a956a2e1c10011d4d5c1e790b688` (matches PR head).

## Stage 1: Spec Compliance

Went through every B2 bullet against the diff:

- `featured` attr: `nil` default (feature off) via `assign_new`, contract `%{uuid: uuid | nil, label: String.t() | nil}` — matches (`media_browser.ex`, moduledoc + `update/2`).
- UI when set: "Set as featured"/"Unset featured" kebab item added to all three kebabs — grid (`media_browser.html.heex:1630`), list (`:2302`), and stack (shared `file_card/1`, wired at `:1776`/`:1835`) — confirmed each is image-gated and trash-gated (`!@filter_trash`) per spec.
- Star badge: `data-role="featured-badge"` overlay, correctly gated on `file_type == "image"` (guards against a stale pointer to a non-image file painting the badge, verified by mutation test below) and *not* trash-gated — matches the spec text literally (the badge sentence has no `filter_trash` clause, only the kebab-item sentence does).
- Viewer sidebar toggle: `media_canvas_viewer.html.heex:402-419`, posts to `:parent_id` via the same `phx-target={"##{@parent_id}"}` routing close/prev/next already use — confirmed `:parent_id` is pre-existing (not touched by this diff), so this is reuse, not new plumbing.
- Events `set_featured`/`unset_featured`: component never persists, `send(self(), {__MODULE__, id, {:set_featured, uuid|nil}})` + optimistic `assign(:featured, %{featured | uuid: uuid})` — matches exactly, and is real (see mutation check).
- `handle_parent_info/2`: untouched, its catch-all clause silently no-ops `{:set_featured, _}` — moduledoc documents that hosts delegating through it must match `{:set_featured, _}` first. Verified accurate by reading the actual catch-all clause.
- Optional `PhoenixKit.Media.Featured` helper: correctly skipped (marked "Optional" in spec).
- Tests (a)-(d) all present in `test/phoenix_kit_web/components/media_browser_featured_test.exs`, plus a bonus e2e viewer-sidebar test via `live_isolated/3`.
- No CHANGELOG/`mix.exs` changes (`git diff upstream/main -- CHANGELOG.md mix.exs` is empty).

No MISSING / WRONG / EXTRA / EDGE_CASE / AMBIGUOUS findings against the B2 letter.

**Spec Verdict:** PASS

---

## Stage 2: Code Quality

### Verification performed
- `mix test test/phoenix_kit_web/components/media_browser_featured_test.exs` → 8/8 pass.
- Re-ran the touched-adjacent suites (media_browser, media_browser_viewer_url, media_browser_upload_drawer, media_browser_scope, media_canvas_viewer + label_sync + media_meta + board, image_editor, media_viewer) → 117/117 pass (0 failures; some pre-existing sandbox-ownership warnings unrelated to this diff, matching the implementer's noted flakiness).
- Full suite: `mix test` → 63 doctests, 5689 tests, 0 failures.
- `mix format --check-formatted` → clean. `mix credo --strict` → "found no issues" (whole repo). `mix compile --warnings-as-errors` → clean.
- **Mutation-checked 2 of the new tests for real** (copied the source file to `/tmp`, mutated it, `Code.compile_file/1`'d the mutant over the already-loaded module in a `mix test`/`mix run` process — the tracked worktree was never touched, confirmed clean via `git status`/`git diff` before and after):
  1. Dropped the `nil`-guard in `notify_featured/2` (so it always relays even when the host never opted in) → the mutant *does* send `{MediaBrowser, "mb", {:set_featured, uuid}}` when `featured: nil`, which the real test's `refute_received {MediaBrowser, _, _}` would catch. Guard is load-bearing.
  2. Dropped `@file.file_type == "image"` from the badge's `:if` → a featured pointer to a PDF now leaks the star badge onto the PDF tile, which the real test's `assert Enum.empty?(...badge...)` would catch. Guard is load-bearing.
  
  Both tests are real, not vacuous.

### Findings

#### [MINOR / should-fix]: Featured badge's `title` tooltip never displays — dead code
**File**: `lib/phoenix_kit_web/components/media_browser.ex:3047-3049`
```elixir
class="absolute top-2 left-2 bg-warning text-warning-content p-1 rounded-full pointer-events-none shadow"
title={(@featured && @featured[:label]) || gettext("Featured image")}
```
**Problem**: The badge `<div>` carries `pointer-events-none` (needed so the badge doesn't block clicks on the tile underneath — correct, matches the pre-existing PDF/video badge pattern). But `pointer-events: none` removes the element from hit-testing entirely, so a mouse hovering that spot never triggers the browser's native `title` tooltip on *this* element. The `label`/tooltip is the one place in the whole change that implements the spec's ambiguous `featured.label` field (per the PR's own "deviations" note) — and it's unreachable in every browser.
**Suggestion**: Wrap just the icon in a small inner element with `pointer-events-auto` sized to the badge (so it alone receives hover while the rest of the tile stays click-through), or drop `title` and expose the label a different way (e.g. `aria-label` won't fix the visual tooltip either — same hit-testing problem) if a real tooltip component exists elsewhere in this repo's convention.
**Rationale**: Ships a non-functional attribute in a public component's contract; harmless today (nothing else reads it) but surprising for the next consumer who sets a custom `label` expecting it to show anywhere.

#### [NITPICK]: `set_featured`/`unset_featured` forward the client-supplied uuid unvalidated
**File**: `lib/phoenix_kit_web/components/media_browser.ex` (`notify_featured/2`, ~3016-3029; `handle_event("set_featured", ...)` ~2303)
**Problem**: Unlike `open_image_editor/2`, `rotate_file`, etc., which resolve the incoming `file-uuid` via `fetch_viewable_file/2` before acting, `notify_featured/2` relays whatever string arrives in `phx-value-file-uuid` straight to the host with no existence/scope/`file_type` check.
**Suggestion**: Not required — this is the documented, intentional "core initiates, host validates" contract (moduledoc says so explicitly), and Block A's host-side `Orders.set_featured_image/2` already re-validates fully. Flagging only as a defense-in-depth nice-to-have for hosts that might trust the relay too much.
**Rationale**: Low risk given the explicit contract, but worth a maintainer's awareness.

#### [NITPICK]: New msgids landed `fuzzy` (machine-guessed) in all 8 non-en locales
**File**: `priv/gettext/{de,es,et,fr,it,pl,ru}/LC_MESSAGES/default.po`
**Problem**: "Featured image" was pre-filled from the existing similar msgid "Featured" (e.g. ru `"Избранное"`), `fuzzy`-flagged, unreviewed.
**Suggestion**: Hand-translate before this ships user-facing (flagged for whoever picks up B5/Andi wiring — B1/B2 scope doesn't mandate translation per the spec).
**Rationale**: Correctly disclosed by the implementer; not blocking this PR, just tracked here for visibility.

#### [INFO, not a defect]: extraction rode along with one pre-existing gap
`mix gettext.extract --merge` is a whole-catalogue recompute, so alongside the 3 new msgids it also newly extracted `"Title & description"` (added in commit `75a48239`, apparently never extracted since) into every locale. Correctly left in rather than hand-reverted (reverting would leave the catalogue in a worse, inconsistent state). Not something to fix here.

**Quality Summary:** 0 critical, 0 major, 1 minor, 2 nitpick (+1 info)
**Quality Verdict:** Needs Work (one small, well-scoped fix)

---

## Overall Verdict: FAIL

Spec compliance is clean — nothing missing, wrong, or extra against B2. The only thing blocking a PASS is the should-fix badge-tooltip bug above (dead `title` attribute on a `pointer-events-none` element). Everything else is advisory (nitpicks/info) and does not need to change.

**Top issue to fix, in priority order:**
1. `media_browser.ex:3047-3049` — make the featured badge's tooltip actually reachable, or drop it.

timujinne (commented):
Review round 2 — Block B2 (`MediaBrowser` featured image), spec: `docs/superpowers/specs/2026-09-17-suborder-featured-image-design.md` §B2. Worktree HEAD `4451a214c915c4329df9d1dd05c4667c9a91ccca` (matches PR head, was `70c22b34` in round 1).

## Previous round finding — resolved

**[should-fix] `media_browser.ex:3047-3049` — featured badge's `title` tooltip unreachable (dead code) because the outer badge div carries `pointer-events-none`.**

Fix in this round (`media_browser.ex:3046-3057`): the outer `<div data-role="featured-badge">` keeps `pointer-events-none` (still doesn't block tile clicks) and no longer carries `title`. A new inner `<span class="block w-3.5 h-3.5 pointer-events-auto" title={...}>` wraps the `<.icon>`, sized to the icon, so it alone receives hover. This is exactly the "wrap just the icon in a small pointer-events-auto inner element" option from round 1 — confirmed the right call over dropping `title` outright, since it preserves `featured.label` actually reaching the browser.

Verified independently, not just read:
- Read the diff line-by-line (`git diff 70c22b34..4451a214`) — matches the description above exactly, no other changes snuck in.
- Confirmed `<.icon>` (`core/icon.ex`) renders a bare `<span class={[@name,@class]}/>` with no pointer-events of its own, so the wrapping span's `pointer-events-auto` is what actually restores hit-testing over the star glyph — not a no-op.
- Mutation-reasoned the new test (`media_browser_featured_test.exs`, "the badge's title tooltip sits on a pointer-events-auto inner element..."): reverting to the old markup (title back on the outer div, no inner span) makes `assert LazyHTML.attribute(badge, "title") == []` fail (old code puts a non-empty title there) and makes the `[data-role="featured-badge"] [title]` query return nothing (old code has no such descendant) — the test is real, not vacuous, and correctly pins the fix.
- `grep -rn "featured-badge"` — no other markup, CSS or JS hook references the badge div's `title` attribute, so removing it from the outer div is not a regression anywhere else.

Resolved, no follow-up needed.

## Stage 1: Spec Compliance (full re-check against B2, not just the round-2 diff)

Went through every B2 bullet again against current HEAD, not just the delta:

- `featured` attr: `nil` default via `assign_new`, `%{uuid: uuid | nil, label: String.t() | nil}` contract — unchanged from round 1, still correct.
- Kebab item on all three surfaces — grid (`media_browser.html.heex:1631`), list (`:2304`), stack (shared `file_card/1`, wired at `media_browser.ex:3181`, called from `:1770`/`:1830`) — all three carry the identical `:if={@featured && file_type=="image" && !@filter_trash}` guard, verified by direct read of each site (not just grep-count).
- Star badge — `data-role="featured-badge"`, gated on `file_type == "image"`, not trash-gated, now with the fixed tooltip — confirmed.
- Viewer sidebar toggle — `media_canvas_viewer.html.heex:399-419`, posts via the pre-existing `phx-target={"##{@parent_id}"}` routing, mirrors close/prev/next — confirmed no new event-handler plumbing was needed, matches the spec's own claim.
- Events `set_featured`/`unset_featured`, optimistic local update, no persistence in the component — confirmed in `notify_featured/2`.
- `handle_parent_info/2` — untouched; moduledoc documents the must-match-first requirement for hosts that delegate everything through it — confirmed accurate against the actual catch-all clause.
- `hero-star-solid` — confirmed already a registered icon elsewhere in this repo (`multilang_form.ex:685`), so no need for the spec's `hero-x-mark` fallback.
- No CHANGELOG/`mix.exs` changes: `git diff upstream/main...HEAD -- CHANGELOG.md mix.exs` is empty.
- `priv/gettext/*` diff is pure `mix gettext.extract` line-number churn plus the 3 new msgids (`Featured image`, `Set as featured`, `Unset featured`) landing with empty `msgstr` in every locale — consistent with how every other untranslated string in this catalogue already sits (spot-checked several pre-existing `msgstr ""` entries); not a regression, translation is explicitly out of scope for B1/B2 per the spec.

No MISSING / WRONG / EXTRA / EDGE_CASE findings against the B2 letter.

One AMBIGUOUS/test-coverage observation (not a spec violation, see Stage 2 nitpick below): the render tests only exercise the default "grid" `view_mode`; the (identically-guarded) list-view and stack-view kebab code paths are exercised by no test. Verified by reading every code path directly — they mirror the tested grid guard exactly — but flagging since a future edit to just one of the three copies would go unnoticed by CI.

**Spec Verdict:** PASS

---

## Stage 2: Code Quality

### Verification performed (all run myself, in the worktree, this round)
- `mix format --check-formatted` on the full repo — clean.
- `MIX_ENV=test mix credo --strict` — "14097 mods/funs, found no issues."
- `PGHOST=localhost PGUSER=postgres PGPASSWORD=postgres MIX_ENV=test mix test test/phoenix_kit_web/components/media_browser_featured_test.exs` — 9 tests, 0 failures.
- Full suite: `PGHOST=localhost PGUSER=postgres PGPASSWORD=postgres MIX_ENV=test mix test` — **63 doctests, 5690 tests, 0 failures** (ran to completion this round, ~223s; the implementer's report correctly flagged this was still running at their report time — it is now confirmed green).
- `git diff upstream/main...HEAD -- CHANGELOG.md mix.exs` — empty, confirmed.

### Findings

#### [NITPICK]: List-view and stack-view featured kebab/badge code paths are untested
**File**: `test/phoenix_kit_web/components/media_browser_featured_test.exs` (whole file); code at `lib/phoenix_kit_web/components/media_browser.html.heex:2304` (list) and `lib/phoenix_kit_web/components/media_browser.ex:3181` via `file_card/1` (stack)
**Problem**: Every render assertion in this test file calls `render_component(MediaBrowser, ...)` with no `phoenix_kit_current_user`/session, so `load_user_view_mode/1` defaults to `"grid"` (`media_browser.ex:3452`) and only the grid view's inline badge/kebab markup is ever rendered and asserted on. The list-view (`:view_mode == "list"`) and stack-view (`file_card/1`, used inside `:view_mode == "stacks"`) copies of the same `:if={@featured && ...}` guard and `set_featured`/`unset_featured` buttons are never exercised by any test.
**Suggestion**: Not required for this PR (I independently verified, by direct read, that all three copies carry the identical guard and phx-click wiring today), but worth a follow-up test that renders with `view_mode: "list"` and `view_mode: "stacks"` (or drives `set_view_mode` before asserting) so a future edit to only one of the three near-duplicate blocks doesn't silently drift.
**Rationale**: The three kebab locations are maintained as independent copy-pasted markup (not a single shared component for grid+list), so nothing currently protects list/stack from regressing independently of the well-tested grid path.

#### Carried forward from round 1, still valid, still non-blocking
- `notify_featured/2` relays the client-supplied `file-uuid` to the host unvalidated — intentional per the documented "core initiates, host validates" contract; no change needed.
- New msgids landed with empty/`fuzzy` translations in the 8 non-en locales — expected, translation is out of scope for B1/B2 per the spec, tracked for whoever does B5.

**Quality Summary:** 0 critical, 0 major, 0 minor, 1 nitpick (new) + 2 nitpicks carried forward, non-blocking
**Quality Verdict:** Ship

---

## Overall Verdict: PASS

The round-1 should-fix (unreachable badge tooltip) is fixed correctly and the fix is covered by a real, mutation-checked test. No new should-fix or blocker findings. Full suite green (63 doctests, 5690 tests, 0 failures), `mix format`/`mix credo --strict` clean, no CHANGELOG/`mix.exs` drift. The one new nitpick (list/stack view-mode test coverage) is advisory only and does not need to change before merge.

timujinne:
Follow-up commit: the translation files now carry only the three new msgids (Featured image / Set as featured / Unset featured) instead of a full `gettext.extract --merge` re-extraction, so this PR no longer rewrites 347 reference lines per locale and does not collide with #827's hand-added entries. Estonian and Russian are filled in, the fuzzy flags are gone; the other locales fall back to English. Featured tests: 9/9.

PR #830, opened 2026-09-18
Read the provider's text body when an email template comes from the database

timujinne:
## The bug

`Content.resolve/5` has two branches that answer in different shapes.

The filesystem branch returns what `PhoenixKit.Templates` renders — `subject`, `text`, `html` — and every consumer reads `content.text` / `content.html` accordingly (`UserNotifier.deliver_templated/5`, and both send paths in `Mailer`).

The database branch mapped `html` from `rendered.html_body`, but took `text` from `rendered.text`. No provider has ever had a `:text` key. Core's own `DefaultProvider` answers `%{subject: "", html_body: "", text_body: ""}`, and `phoenix_kit_emails` validates exactly those three keys on every render.

So **every send that found an active template raised `KeyError`**:

```
** (KeyError) key :text not found in:
    %{subject: "Confirm your account", html_body: "...", text_body: "..."}
    (phoenix_kit) lib/phoenix_kit/email/content.ex:102: PhoenixKit.Email.Content.resolve/5
    (phoenix_kit) lib/phoenix_kit/users/auth/user_notifier.ex:44: UserNotifier.deliver_templated/5
```

On an install with the stock templates seeded that is registration confirmation, password reset and email change — plus anything `Mailer.send_from_template` sends, the billing invoice and receipt included. Probed against a running install:

```
register                → raises      reset_password → raises      update_email → raises
magic_link_registration → ok (no database template)   new_login_alert → ok
```

Present since the branch was introduced in 2.15.1 (`Add filesystem template resolution to the auth emails`): the new filesystem path was written against `:text`/`:html`, and the existing database path was left reading the same names off a differently-shaped map.

## How it surfaced

A guest checkout in a shop built on this. The order committed, then the guest confirmation email raised, which killed the checkout LiveView; the browser reconnected, the view remounted, found the cart already converted and redirected to a new, empty cart. The shopper's entire feedback for a placed order was "your cart is empty". The order's activity record and the operator's new-order notification never ran either. (The host module's own part of that — not letting the email failure abort the rest — is a separate PR.)

## The fix

`text: rendered.text_body`. One line, at the point where the two shapes are supposed to be normalised into the one every consumer already expects.

## Tests

The database branch had no tests at all, which is how a total shape mismatch stayed invisible. `ContentDbTemplateTest` stubs the provider (the only way to reach the branch) and pins subject, text, html and `db_template`, plus one test that keeps the fallback branch honest so a future change cannot quietly swap which shape wins.

Full suite on this branch: 63 doctests, 5686 tests, 0 failures.


timujinne (commented):
## Review

Independent review, separate from the author. Verdict: **PASS**, no blockers.

### Verified, not taken on trust

- The provider contract really is `%{subject:, html_body:, text_body:}` on both sides: core's own `DefaultProvider` (`lib/phoenix_kit/email/default_provider.ex:26`) and the real `phoenix_kit_emails` implementation, where a second in-package consumer (`template_editor.ex:377-386`) already reads `html_body`/`text_body`. So the mismatch was purely in the caller, and normalising at the `Content.resolve` boundary — rather than changing a provider — is the correct placement: `Content` owns its own `resolved()` shape, and changing the provider would break that other consumer.
- All three consumers (`UserNotifier.deliver_templated/5`, `Mailer.send_from_template/4`, `Mailer.send_magic_link_email/2`) read `.text`/`.html`, matching `Content.resolve`'s own typedoc — now honoured by both branches.
- `content.ex:98` is the only `render_template/3` call site in core; no other branch was missed.
- The filesystem branch needs no analogous change: `PhoenixKit.Templates.render/4` returns `%{subject:, text:, html:}` natively.
- The "previously untested" claim holds: every test in the existing `content_test.exs` exercises the nil-template branch, and `mailer_test.exs`'s provider tests build the `Swoosh.Email` by hand, bypassing `Content.resolve` entirely.
- The new tests fail on the old code by construction — the stub returns a map with no `:text` key at all, so `rendered.text` raises — rather than asserting the field they set.

### Gates, run independently

`mix format --check-formatted` clean; `mix credo --strict` clean; `mix compile --force --warnings-as-errors` clean; targeted tests 12/12; email + mailer + notifier 61/61; full suite **63 doctests, 5686 tests, 0 failures**, with the output grepped for "Excluding"/"excluded" to confirm a genuinely complete run rather than a fail-soft one.

### Two non-blocking findings, and why they were not acted on

- **No CHANGELOG entry.** Checked the convention against a merged PR from the same contributor (#822): seven files, all `lib/` and `test/`, no CHANGELOG entry — the maintainer writes those at release time. Adding one here would collide with that flow rather than follow it.
- **The test's `StubProvider` does not declare `@behaviour PhoenixKit.Email.Provider`.** The behaviour has 14 callbacks and exactly one optional, so declaring it on a three-function stub emits twelve "not implemented" warnings — and this project compiles with `--warnings-as-errors`. `DefaultProvider` already serves as the anchor for that contract.

timujinne (commented):
Independent review, not written by the person who submitted this PR.

What I checked:
- Diffed `upstream/main..HEAD` in a fresh worktree of `origin/fix/db-template-text-body` (1 commit on top of `upstream/main` 2.29.1, only `lib/phoenix_kit/email/content.ex` and one new test file changed, 90 lines).
- Read `Provider` behaviour and `DefaultProvider` (`lib/phoenix_kit/email/provider.ex`, `lib/phoenix_kit/email/default_provider.ex`): the render callback contract really is `%{subject:, html_body:, text_body:}`, no provider has ever had a `:text` key.
- Read the filesystem branch, `PhoenixKit.Templates.render/4` (from the `phoenix_kit_templates` hex dependency, after `mix deps.get`): it already returns `%{subject:, text:, html:}` directly, so only the database branch of `Content.resolve/5` needed normalising.
- Grepped every caller of `Content.resolve/5` (`lib/phoenix_kit/mailer.ex:193,672`, `lib/phoenix_kit/users/auth/user_notifier.ex:44`) — all three read only `subject`, `text`, `html`, `db_template`, matching the fixed shape.
- Confirmed no version bump or CHANGELOG entry is in the diff (fine — that's the maintainer's call at merge time).
- `mix format --check-formatted` and `mix credo --strict` on both touched files: clean, no issues.
- Compiled clean (`mix compile`, `mix deps.get` beforehand; no changes were left in `mix.lock`).
- Ran the new test file against a live Postgres test DB:
  - Against this branch: `5 tests, 0 failures`.
  - After reverting only `content.ex` to `upstream/main` (keeping the new test): `5 tests, 4 failures`, with the exact `KeyError: key :text not found in: %{subject:, html_body:, text_body:}` from the PR description. This proves the bug is real on `upstream/main` and the test actually pins it.
  - Ran the broader `test/phoenix_kit/email/`, `mailer_test.exs`, `user_notifier_test.exs` on the fixed branch: `61 tests, 0 failures`.
- Checked the new test's provider-config swap (`setup`/`on_exit` around `:phoenix_kit, :email_provider`) against the existing idiom in `test/phoenix_kit/mailer_test.exs` (same save/restore pattern, same `async: false` module). Consistent with the codebase's own convention, so no config-leak risk into other async test files.

Spec verdict: PASS. The claimed bug is real and reproducible, the fix is the minimal one-line correction at the point that's supposed to normalise both branches, and every consumer of the resolved map only reads the normalised keys. No scope creep (no version bump, no unrelated changes).

Findings: none. No MISSING/WRONG/EXTRA/EDGE_CASE/AMBIGUOUS issues in spec compliance, and no code-quality issues (nitpick or above) in the diff.

Verdict: PASS

PR #834, opened 2026-09-19
Add FeaturedImage component: one image pointer with a touch-friendly change/remove menu

timujinne:
## What

`PhoenixKitWeb.Components.FeaturedImage` — a LiveComponent for "one image pointer on an entity": thumbnail (or an empty placeholder) with an always-visible Change / Remove menu beside it. It owns the whole `MediaSelectorModal` protocol and leaves the host a single required message:

```elixir
{PhoenixKitWeb.Components.FeaturedImage, id, {:set_featured, uuid | nil}}
```

Same payload shape as `MediaBrowser`'s `featured` attr (#829), so a host funnels both into one write function. The component never writes to the database and does not know where the pointer lives (`data`, `metadata`, `custom_fields`, a setting, a column) or when the host persists it (immediately or on the form's Save).

## Why

The control is hand-rolled in ~10 places (catalogue ×3, locations ×2, manufacturing, staff, crm ×2, hosts) plus 4 relatives in core (settings, authorization, user form, entities), and the copies have diverged: only some guard against trashed files, locations/manufacturing don't lock the picker to images, and several use hover-only buttons that cannot be reached on a touch screen.

## Contract highlights

- **Controlled**: renders only the `uuid` the host gives, no optimistic update.
- **`picker_scope`**: `{:folder, uuid}` | `:lazy` | `nil`. With `:lazy` (every module creates the entity's folder on click) the component sends `:scope_requested`; the host creates the folder and answers `send_update(..., open_picker: true | false | error: msg)`. `nil` never opens the picker (an unscoped selector would expose the whole library).
- **Picker rendered through `<.portal>`**: the selector has its own `<form>` + `live_file_input`, and most hosts place this control inside their own `<.form>`. Always `mode: :single`, images only, type locked.
- **Dangling pointer** (trashed / missing / not an image / failed): drawn as "Image missing" with Remove still available; a read-only viewer sees plain empty. `processing` is re-checked (3 s × 10).
- **`on_preview`**: `"event"` | `{"event", values}` | `%JS{}` — the host usually needs the entity's uuid, not the file's.
- `readonly` re-checked in every event and `update/2` clause; the moduledoc states the component does not authorize.
- Opt-in `report_close` → `:picker_closed` (sent after `:set_featured`), for hosts that refresh a file list after an upload from the picker.
- `size` (`:sm`/`:md`/`:lg`), `shape` (`:square`/`:circle`), `confirm_remove`, slots `:empty` (avatar initials) and `:empty_hint`.

Also: `trigger_size` (`"xs"` default | `"sm"` | `"md"`) on `table_row_menu` — the trigger was a hard `btn-xs` (24px), and in the built CSS `btn-xs` wins over a `btn-sm` passed via `trigger_class`. Default output is unchanged apart from a trailing space in the class attribute.

## Tests

99 tests in the two new files (ConnCase, real Postgres; green on 10 random seeds). ExUnit cannot see inside a `<.portal>` `<template>`, so the modal → `notify` → `update/2` seam is covered by a test with the real `MediaSelectorModal` beside the component. Mutation-checked: 52 mutants, 47 killed, the 4 survivors are equivalent / unobservable from ExUnit (listed in the review).

## Browser check

Verified by the owner in a host app (Andi, sub-order Overview, the component outside a form): the menu opens by click/tap and is not clipped, Change / Remove / View work through the portal-rendered picker, the trigger sits centred against the thumbnail. Tailwind emission of the new literals (`size-12/16/24`, `btn-md`, `border-warning/60`, `text-[0.65rem]`) confirmed in the host's built CSS.

Not verified in a browser, only from the LiveView 1.2.12 client sources and tests: the picker inside a host `<form phx-change>` (upload, Esc, no leaked form events), the `:lazy` handshake on a real page, RowMenu + `phx-target` inside a nested `live_render`. The moduledoc says so.

Rebased onto main (2.31.1); `mix precommit` clean, component + media tests green on Postgres.

Not included, by design: `:library` scope, crop, list thumbnails. Module migrations follow as separate PRs.


timujinne (commented):
## Review (two rounds) — no open findings

Reviewed against the design spec, then for quality; every finding from round 1 is fixed in this tree.

**Round 1 → fixed**
- A dangling pointer ignored the `:empty` slot (avatars would have lost their initials) → the slot is kept, Remove stays in the menu.
- A read-only viewer saw the "Image missing" warning and a stale error line → renders as plain empty; `error` is never shown under `readonly`.
- `pending_scope` was not cleared when `picker_scope` left `:lazy`; group had no accessible name; DB-error `rescue` branch and non-memoisation of `:unavailable` were untested; `Ecto.UUID.cast` indistinguishable from the `rescue`.
- One vacuous test ("buttons never fire host form change" — LiveViewTest cannot produce that event) replaced.
- Non-scalar `on_preview` values crashed the render → dropped; `uuid: ""` is treated as `nil`; nested-LiveView section added to the moduledoc.

**Gates re-run on the final tree**
- 99 tests in the two new files, 0 failures on 10 random seeds; they really hit Postgres (with an unreachable `PGHOST` the run reports `0 tests, 94 excluded`).
- Wider media/core set green; `mix format --check-formatted`, `mix credo --strict` clean.
- Mutation check (worktree only): 52 mutants, 47 killed. Survivors: one equivalent (`not @readonly` in the portal's `:if` — a read-only component cannot have `show_picker`); the modal's `notify` tuple inside the portal `<template>` (invisible to ExUnit; the seam is tested with the real `MediaSelectorModal` beside the component); two missing asserts on sr-only text (nit).
- `CHANGELOG.md` / `mix.exs` untouched.

**Open — browser only, keeps the PR in draft:** the checklist in the description (picker inside a host `<form>`, `:lazy` on a real page, RowMenu portal in a nested `live_render`, Tailwind emission, visuals).

timujinne (commented):
## Re-review after the rebase onto 2.31.1 — no findings

Rebase only; the design was not re-reviewed.

- `git range-diff` against the pre-rebase tip: commits 1 and 2 are patch-identical, commit 3 differs only in the gettext catalogs. `featured_image.ex`, `featured_image.html.heex`, `table_row_menu.ex`, both test files and the guide are byte-equal before and after.
- Catalogs (`default.pot` + de/en/es/et/fr/it/pl/ru): all parse; +38/−0 lines per file against `main`; 2827 → 2833 entries each; 0 upstream entries lost or altered, 0 duplicate msgids, headers unchanged; the 8 entries `main` gained since the old base are preserved. The 6 new msgids are in all 9 files with the same translations as before the rebase; every `#:` reference points at a line with the gettext call.
- `mix compile` exits 0; the compiled `PhoenixKitWeb.Gettext` backend returns the expected string for 8 locales × 6 msgids.
- Postgres, nothing excluded: `featured_image_test` 97, `table_row_menu_test` 5, `media_gallery_test` 43, `media_browser_featured_test` 10 → 155 tests, 0 failures; the gettext tests add 21, 0 failures. `mix precommit` exits 0.
- `CHANGELOG.md` / `mix.exs` untouched; `git diff --check` clean.

Not re-checked: per-commit bisectability (verified before the rebase). Browser-only items remain as listed in the description.

timujinne (commented):
## Rebased onto 2.32.1 — rebase check, no findings

`main` appended new entries at the end of the nine gettext files again, which conflicted with this PR's block at the same spot. Rebased; to stop this recurring with every release, the six new entries now sit mid-file, right after the existing `"Unset featured"` entry, instead of at EOF.

- `git range-diff`: commits 1 and 2 are patch-identical; commit 3 differs only in where the gettext block sits.
- Each catalog = `main`'s file + one contiguous 38-line block: removing the block reproduces `main` byte for byte in all 9 files; 2841 entries each, 0 duplicate msgids; translations unchanged.
- Postgres, nothing excluded: `featured_image_test`, `table_row_menu_test`, `media_gallery_test`, `media_browser_featured_test` → 155 tests, 0 failures. `mix precommit` exits 0.
- `CHANGELOG.md` / `mix.exs` untouched.

